PR #1, opened 2020-11-13
populate with dps kotlin template

tomsmyers:


.circleci/config.yml
```diff
@@ -0,0 +1,86 @@
...
+#          type: approval
+#          requires:
+#            - deploy_preprod
+#      - hmpps/deploy_env:
+#          name: deploy_prod
+#          env: "prod"
+#          slack_notification: true
+#          context: hmpps-interventions-service-prod
+#          requires:
+#            - request-prod-approval
```
sldblog, line 74: What do you think about removing these comments until we decide what to do with the pipeline?

.circleci/config.yml
```diff
@@ -0,0 +1,86 @@
+version: 2.1
+
+orbs:
+  owasp: entur/owasp@0.0.10
+  hmpps: ministryofjustice/hmpps@1.1.2
```
sldblog, line 5: _suggestion_
Looks like these have newer versions now:
```suggestion
  owasp: entur/owasp@0.0.13
  hmpps: ministryofjustice/hmpps@1.1.3
```

build.gradle.kts
```diff
@@ -0,0 +1,12 @@
+plugins {
+  id("uk.gov.justice.hmpps.gradle-spring-boot") version "1.0.6"
```
sldblog, line 2: _suggestion_
A newer version seems to be available:
```suggestion
  id("uk.gov.justice.hmpps.gradle-spring-boot") version "1.1.1"
```

sldblog (commented):
Great stuff!

A couple of version bump opportunities and questions which mostly stem from my unfamiliarity with the template.

The only kinda-serious thing in my mind is clarifying the license.

LICENSE
```diff
@@ -0,0 +1,674 @@
+                    GNU GENERAL PUBLIC LICENSE
```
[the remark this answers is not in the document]
tomsmyers: yep, you're right. will update. 

https://gds-way.cloudapps.digital/manuals/licensing.html#use-mit

.circleci/config.yml
tomsmyers: makes sense. otherwise could lead to confusion now.

src/main/resources/banner.txt
```diff
@@ -0,0 +1,10 @@
+  _       _                            _   _                 
+ (_)     | |                          | | (_)                
+  _ _ __ | |_ ___ _ ____   _____ _ __ | |_ _  ___  _ __  ___ 
+ | | '_ \| __/ _ \ '__\ \ / / _ \ '_ \| __| |/ _ \| '_ \/ __|
+ | | | | | ||  __/ | (_) V /  __/ | | | |_| | (_) | | | \__ \
+ |_|_| |_|\__\___|_|___ \_/_\___|_| |_|\__|_|\___/|_| |_|___/
+ / __|/ _ \ '__\ \ / / |/ __/ _ \                            
+ \__ \  __/ |   \ V /| | (_|  __/                            
+ |___/\___|_|    \_/ |_|\___\___|                            
+
```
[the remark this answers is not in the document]
tomsmyers: this pattern is baked into the template. let's address this in a seperate small PR to avoid danger of breaking the template build?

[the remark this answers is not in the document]
tomsmyers: the banner only shows at INFO level i believe. will not affect production logs under normal circumstances.

sldblog (approved):

sldblog:
🎉

PR #3, opened 2020-11-16
Add code owners and pull request template

sldblog:

Adds code owners and a pull request template like this one.

So we can auto-request reviews and explain our intent with every change for the future.


tomsmyers (approved):
👍🏼

PR #4, opened 2020-11-16
Add dependabot configuration

sldblog:

Adds [dependabot](https://docs.github.com/en/free-pro-team@latest/github/administering-a-repository/configuration-options-for-dependency-updates) config for Gradle and Docker.

We want to make it easier to be aware of out of date dependencies.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,15 @@
+version: 2
+updates:
+  - package-ecosystem: "gradle"
+    directory: "/"
+    schedule:
+      interval: "daily"
+      time: "02:00"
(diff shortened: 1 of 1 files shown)


.github/dependabot.yml
```diff
@@ -0,0 +1,15 @@
+version: 2
+updates:
+  - package-ecosystem: "gradle"
```
sldblog, line 3: Hmm, it seems dependabot cannot process `build.gradle.kts` files at the moment
I'll mark as draft

sldblog:
I _think_ #12 covers this visibility concern around dependencies.

PR #6, opened 2020-11-17
Remove filtering IPs on the ingress

sldblog:

Removes IP access list controls from **all** environments.

Current guidance:

> IP addresses access control lists (and/or techniques such as HTTP basic authentication) should be used to restrict access to non-production systems you do not wish general users to access.

Source: https://ministryofjustice.github.io/security-guidance/authentication/#ip-address-for-non-production-systems

We will

- have production open to the internet,
- have pre-production have the same security measures,
- have no sensitive data on development,
- will use authentication and authorisation on all environments.

Given the above, we don't have a strong use case to restrict access.

### DIFF
--- a/helm_deploy/hmpps-interventions-service/templates/ingress.yaml
+++ b/helm_deploy/hmpps-interventions-service/templates/ingress.yaml
@@ -9,7 +9,6 @@ metadata:
     {{- include "app.labels" . | nindent 4 }}
   annotations:
     kubernetes.io/ingress.class: "nginx"
-    {{ if .Values.ingress.enable_whitelist }}nginx.ingress.kubernetes.io/whitelist-source-range: {{ include "app.joinListWithComma" .Values.whitelist | quote }}{{ end }}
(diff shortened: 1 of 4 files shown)


mattops:
In principal it is ok to make this public - but I dont know enough about your application or intended users.  The article you linked to makes some excellent points about IP address access lists - however they still can play a part in a defence-in-depth strategy. With this being a new application it might be an idea to review and ensure other measures are in place.  One thing you could consider is a WAF.  You can enable this using mod_security and the OWASP core rule set - a feature available in the nginx-ingress controllers.

sldblog:
Thanks, @mattops! I heard a year ago that having a WAF in front of every service was a recommended thing, so perhaps we should enable that anyway 🤔 I'll pull on some threads

tomsmyers (approved):
based on feedback from cyber security team i'm happy to approve - as long as we prioritize getting WAF enabled for prod/preprod deploys asap

PR #8, opened 2020-11-17
Remove all current secrets

sldblog:

Remove all secrets as we don't need any at the moment.

This, in turn, makes deployment work 🎉 

```
$ helm upgrade hmpps-interventions-service hmpps-interventions-service \
  --install --wait --reset-values --timeout 5m --history-max 10 \
  --values values-dev.yaml --set image.tag="2020-11-17.26.8541d50"
...

$ kubectl get deployment --namespace hmpps-interventions-dev
NAME                          READY   UP-TO-DATE   AVAILABLE   AGE
hmpps-interventions-service   2/2     2            2           17m
```

We do not currently have any secrets. The `APPINSIGHTS_INSTRUMENTATIONKEY` is not immediately required for us.

### DIFF
--- a/helm_deploy/hmpps-interventions-service/templates/_envs.tpl
+++ b/helm_deploy/hmpps-interventions-service/templates/_envs.tpl
@@ -13,10 +13,4 @@ env:
   - name: SPRING_PROFILES_ACTIVE
     value: "logstash"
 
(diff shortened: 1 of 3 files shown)


tomsmyers (approved):
ok for now 👍

PR #11, opened 2020-11-18
build docker image on PR branches, but do not publish

tomsmyers:

runs circleci build_docker job on all branches, but only publishes to quay.io on main

validate changes to dockerfile in PRs before merging to main

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -36,7 +36,7 @@ workflows:
           name: helm_lint
       - hmpps/build_docker:
           name: build_and_publish_docker
-          snyk-scan: true
+          snyk-scan: false # fixme!
           filters:
             branches:
               only:
@@ -44,7 +44,7 @@ workflows:
       - hmpps/build_docker:
           name: build_docker
           publish: false
-          snyk-scan: true
+          snyk-scan: false # fixme!
           filters:
             branches:
               ignore:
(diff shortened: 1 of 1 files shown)


sldblog:
Looks like it needs a token in CircleCI secrets

tomsmyers:
working now :) 

looks like we have to address some known high severity CVEs. will look tomorrow.

sldblog:
🎉 

Seems like all of them come from the base image itself https://snyk.io/test/docker/openjdk%3A11-jre-slim?severity=high&policy=open&tab=issues

tomsmyers:
@sldblog it looks like using the alpine tags of openjdk is not a good idea based on the docs:

```
What this means is that Alpine based images are only released for early access release versions of OpenJDK. Once a particular release becomes a "General-Availability" release, the Alpine version is dropped from the "Supported Tags"; they are still available to pull, but will no longer be updated.
```

sldblog:
👍 Indeed. It looks like https://hub.docker.com/r/adoptopenjdk/openjdk11 is unofficial as well.

tomsmyers:
@sldblog i've disabled the vulnerability scanning on this PR for the time being. just to unblock this until we sort out the base image issues.

sldblog (approved):

PR #13, opened 2020-11-23
Add resource server config to secure api endpoints

tomsmyers:

Adds resource server configuration for creating secure endpoints. 

Adds a sample endpoint for testing in dev environment. response looks like 

```
{
    "client": "interventions",
    "roles": [
        "ROLE_INTERVENTIONS",
        "ROLE_AUTH_DELIUS_LDAP",
        "ROLE_COMMUNITY"
    ],
    "tokenExpiry": "2020-11-26T12:11:00Z"
}
```

scaffolding to allow development of secure endpoints

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,17 +1,18 @@
 version: "3"
 services:
-  hmpps-interventions-service:
-    build:
-      context: .
-    network_mode: "host"
-    container_name: hmpps-interventions-service
+  hmpps-auth:
+    image: quay.io/hmpps/hmpps-auth:latest
+    networks:
+      - hmpps
+    container_name: hmpps-auth
     ports:
-      - "8081:8080"
+      - "8090:8080"
     healthcheck:
-      test: ["CMD", "curl", "-f", "http://localhost:8080/health/ping"]
+      test: ["CMD", "curl", "-f", "http://localhost:8080/auth/health"]
     environment:
       - SERVER_PORT=8080
       - SPRING_PROFILES_ACTIVE=dev
+      - SPRING_JPA_PROPERTIES_HIBERNATE_SHOW_SQL=false
 
 networks:
   hmpps:
--- a/README.md
+++ b/README.md
@@ -4,6 +4,11 @@ This is the HMPPS Interventions Service
 
 ## Quickstart
 
+### Requirements
+
+- Docker 
+- Java
+
 Build and test:
 ```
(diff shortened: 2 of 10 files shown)


docker-compose.yml
```diff
@@ -1,14 +1,27 @@
 version: "3"
 services:
-  hmpps-interventions-service:
-    build:
-      context: .
-    network_mode: "host"
-    container_name: hmpps-interventions-service
+  hmpps-auth:
+    image: quay.io/hmpps/hmpps-auth:latest
```
sldblog, line 4: Very generic question, sorry. Is it safe to assume `latest` is the most recent build on the `main` branch?

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/CommunityApiService.kt
```diff
@@ -0,0 +1,24 @@
...
+class CommunityApiService(private val communityApiWebClient: WebClient) {
+  fun <T> emptyWhenNotFound(exception: WebClientResponseException): Mono<T> = emptyWhen(exception, HttpStatus.NOT_FOUND)
+  fun <T> emptyWhen(exception: WebClientResponseException, statusCode: HttpStatus): Mono<T> =
+    if (exception.rawStatusCode == statusCode.value()) Mono.empty() else Mono.error(exception)
+
+  fun getDetailsByUsername(username: String): User? {
+    return communityApiWebClient.get()
+      .uri("/secure/users/$username/details")
+      .retrieve()
+      .bodyToMono(User::class.java)
```
sldblog, line 20: We'll most likely have CRN lookups from the probation offender search (possibly in the UI?), so we may not use this endpoint. Is it ok to remove the community API service until we have its first use case from a story?

sldblog (commented):
I'm wondering if we should wait for the community-api bits on the first user story that requires it? The (read) API _may_ not be used in this app, which could lead to unused code.

tomsmyers:
@sldblog i added the community api endpoint purely to demo that auth is working e2e. no intention of keeping around if not needed.

docker-compose.yml
tomsmyers: that's right

sldblog (approved):

sldblog:
Sorry for taking so long!

PR #14, opened 2020-11-30
ignore base image vulns with no upgrade path for 2 months

tomsmyers:

ignore the 10 high severity CVEs which are currently found in the docker base image. the ignore config expires in 2 months so we can reasses the image then.

unblock the deploy. ensure we look at the image again before release by setting expiry dates.

### DIFF
--- a/.snyk
+++ b/.snyk
@@ -1,0 +1,47 @@
+# Snyk (https://snyk.io) policy file, patches or ignores known vulnerabilities.
+version: v1.14.1
+
+# ignores vulnerabilities until expiry date; change duration by modifying expiry date
+ignore:
+  SNYK-DEBIAN10-SYSTEMD-345386:
+    - '*':
+        reason: base image vuln with no upgrade path; will reassess in 2 months
+        expiry: 2021-02-01T00:00:00.000Z
+  SNYK-DEBIAN10-SYSTEMD-345391:
+    - '*':
+        reason: base image vuln with no upgrade path; will reassess in 2 months
+        expiry: 2021-02-01T00:00:00.000Z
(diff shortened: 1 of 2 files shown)


.snyk
```diff
@@ -0,0 +1,47 @@
+# Snyk (https://snyk.io) policy file, patches or ignores known vulnerabilities.
+version: v1.14.1
+
+# ignores vulnerabilities until expiry date; change duration by modifying expiry date
+ignore:
+  SNYK-DEBIAN10-SYSTEMD-345386:
```
sldblog, line 6: I'm not sure the situation will improve significantly if we keep Debian base images. The `openjdk:11-slim` image has the security apt sources commented out:

```
$ docker run -ti openjdk:11 cat /etc/apt/sources.list
# deb http://snapshot.debian.org/archive/debian/20201117T000000Z buster main
deb http://deb.debian.org/debian buster main
# deb http://snapshot.debian.org/archive/debian-security/20201117T000000Z buster/updates main
deb http://security.debian.org/debian-security buster/updates main
# deb http://snapshot.debian.org/archive/debian/20201117T000000Z buster-updates main
deb http://deb.debian.org/debian buster-updates main
```

❓ I think the recommended way to deal with security problems is to update the base image. Shall we do that next?

sldblog (approved):
LGTM as an interim step. One question around next steps:

sldblog:
🤔 Hmm, it would be good to involve the cybersecurity team in these PRs. I'll reach out

.snyk
tomsmyers: there are no upgrade paths for most of these issues without going to the development release of debian. 

we can look into using different base images, but i couldn't find anything obvious that would be suitable via quick google

PR #15, opened 2020-11-30
add postgres db for local dev and test environment

tomsmyers:

adds docker-compose service for postgres for local dev and test

first steps to getting the hibernate stuff up and running.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -15,5 +15,6 @@ dependencies {
   // database
   implementation("org.springframework.boot:spring-boot-starter-data-jpa")
   implementation("org.springframework.boot:spring-boot-starter-validation")
+  implementation("org.hibernate:hibernate-core:5.4.24.Final")
   runtimeOnly("org.postgresql:postgresql")
 }
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -8,6 +8,12 @@ configurations {
 }
 
 dependencies {
+  // security
   implementation("org.springframework.boot:spring-boot-starter-webflux")
   implementation("org.springframework.boot:spring-boot-starter-oauth2-resource-server")
+
+  // database
+  implementation("org.springframework.boot:spring-boot-starter-data-jpa")
+  implementation("org.springframework.boot:spring-boot-starter-validation")
+  runtimeOnly("org.postgresql:postgresql")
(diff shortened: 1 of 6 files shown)


karenmillermoj (approved):

docker-compose.yml
```diff
@@ -14,5 +14,15 @@ services:
       - SPRING_PROFILES_ACTIVE=dev
       - SPRING_JPA_PROPERTIES_HIBERNATE_SHOW_SQL=false
 
+  postgres:
+    image: postgres:13
```
sldblog, line 18: _suggestion_
`postgres:13-alpine` is about 50% smaller

```
$ docker images | grep postgres
postgres   13                      f51c55ac75ed        12 days ago         314MB
postgres   13-alpine               770f846cacce        2 weeks ago         159MB
```

build.gradle.kts
```diff
@@ -8,6 +8,13 @@ configurations {
...
 dependencies {
+  // security
   implementation("org.springframework.boot:spring-boot-starter-webflux")
   implementation("org.springframework.boot:spring-boot-starter-oauth2-resource-server")
+
+  // database
+  implementation("org.springframework.boot:spring-boot-starter-data-jpa")
+  implementation("org.springframework.boot:spring-boot-starter-validation")
+  implementation("org.hibernate:hibernate-core:5.4.24.Final")
+  runtimeOnly("org.postgresql:postgresql")
```
sldblog, line 19: Do we use an in-memory database for testing on CI?

docker-compose.yml
tomsmyers: 👍 thanks for this - i will update

build.gradle.kts
tomsmyers: i would prefer to not to use the in memory db at all if we can help it. 

for integration tests on circle we can run the services via docker-compose.

sldblog: Totally agree -- I didn't see the PostgreSQL container spun up on the CircleCI config which is why I wondered

tomsmyers: ah - yeh we don't have anything to test yet so left that out of this PR. i actually explicitly added an int test that checks the DB is down!

PR #16, opened 2020-11-30
add helm env vars for hmpps auth baseurl

tomsmyers:

adds required env vars for hmpps auth config

fix the deploy

----

At startup, we were getting a `CrashLoopBackOff` on the pods, with this log:

```
2020-11-30 15:54:28.512 ERROR 1 --- [           main] o.s.boot.SpringApplication               : Application run failed |

org.springframework.beans.factory.BeanCreationException: Error creating bean with name 'springSecurityFilterChain' defined in class path resource [org/springframework/security/config/annotation/web/configuration/WebSecurityConfiguration.class]: Bean instantiation via factory method failed; nested exception is org.springframework.beans.BeanInstantiationException: Failed to instantiate [javax.servlet.Filter]: Factory method 'springSecurityFilterChain' threw exception; nested exception is org.springframework.beans.factory.BeanCreationException: Error creating bean with name 'jwtDecoderByJwkKeySetUri' defined in class path resource [org/springframework/boot/autoconfigure/security/oauth2/resource/servlet/OAuth2ResourceServerJwtConfiguration$JwtDecoderConfiguration.class]: Bean instantiation via factory method failed; nested exception is org.springframework.beans.BeanInstantiationException: Failed to instantiate [org.springframework.security.oauth2.jwt.JwtDecoder]: Factory method 'jwtDecoderByJwkKeySetUri' threw exception; nested exception is java.lang.IllegalArgumentException: Invalid JWK Set URL "/.well-known/jwks.json" : no protocol: /.well-known/jwks.json
<snip>
Caused by: java.net.MalformedURLException: no protocol: /.well-known/jwks.json
```

### DIFF
--- a/helm_deploy/hmpps-interventions-service/templates/_envs.tpl
+++ b/helm_deploy/hmpps-interventions-service/templates/_envs.tpl
@@ -13,7 +13,7 @@ env:
   - name: SPRING_PROFILES_ACTIVE
     value: "logstash"
 
-  - name: HMPPS_AUTH_BASEURL
-    value: "{{ .Values.env.HMPPS_AUTH_BASEURL }}"
+  - name: HMPPSAUTH_BASEURL
+    value: "{{ .Values.env.HMPPSAUTH_BASEURL }}"
 
 {{- end -}}
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -15,7 +15,7 @@ services:
       - SPRING_JPA_PROPERTIES_HIBERNATE_SHOW_SQL=false
 
   postgres:
-    image: postgres:13
+    image: postgres:13-alpine
     container_name: postgres
     ports:
       - "5432:5432"
--- a/helm_deploy/values-dev.yaml
+++ b/helm_deploy/values-dev.yaml
@@ -16,3 +16,4 @@ ingress:
 
 env:
   JAVA_OPTS: "-Xmx512m"
+  HMPPSAUTH_BASEURL: "https://sign-in-dev.hmpps.service.justice.gov.uk/auth"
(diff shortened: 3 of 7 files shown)


helm_deploy/hmpps-interventions-service/templates/_envs.tpl
```diff
@@ -13,7 +13,7 @@ env:
   - name: SPRING_PROFILES_ACTIVE
     value: "logstash"
 
-  - name: HMPPS_AUTH_BASEURL
-    value: "{{ .Values.env.HMPPS_AUTH_BASEURL }}"
+  - name: HMPPSAUTH_BASEURL
```
sldblog, line 16: Is this because

https://github.com/ministryofjustice/hmpps-interventions-service/blob/e3ccd3129c53005f99ebaf4c9bb40be9fb6f6802/src/main/resources/application.yml#L17-L19

resolves to an environment variable name without the dash?

sldblog (commented):
Clarifying question, I'm not sure I understand how the change connects to the configuration (probably my ignorance, sorry)

helm_deploy/hmpps-interventions-service/templates/_envs.tpl
tomsmyers: yes. as per https://docs.spring.io/spring-boot/docs/2.1.18.RELEASE/reference/html/boot-features-external-config.html#boot-features-external-config-relaxed-binding-from-environment-variables

sldblog: If yes, would it make sense to match the argument as well to `${hmppsauth.baseurl}` so text searches can find them at the same time?

tomsmyers: good idea.

sldblog: thanks for clarifying!

sldblog (approved):

PR #18, opened 2020-12-02
add minimal skeleton code for referral creation

tomsmyers:

adds an endpoint for referral creation and a test for the repository interface.

adding the h2 test dependency caused a vulberability scan failure, which alerted me to the fact that snyk configuration is not getting loaded for certain build jobs. so that is fixed in this PR too. 

get the referral endpoint work off the ground by putting the JPA scaffolding in place

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -43,6 +43,12 @@ jobs:
       monitor:
         type: boolean
         default: false
+      snyk-scan:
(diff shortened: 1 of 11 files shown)


karenmillermoj (commented):
Referral structure looks good

karenmillermoj (approved):

PR #20, opened 2020-12-04
Referral dto

karenmillermoj:

Add tests for creation of DraftReferral dto from the Referral Entity object

Ensure draft referral created correctly

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,6 +4,11 @@ This is the HMPPS Interventions Service
 
 ## Quickstart
 
+### Requirements
+
+- Docker 
+- Java
+
 Build and test:
 ```
 ./gradlew build
@@ -11,7 +16,9 @@ Build and test:
 
 Run:
 ```
-./gradlew bootRun
+docker-compose pull
+docker-compose up -d
+SPRING_PROFILES_ACTIVE=local ./gradlew bootRun
 ```
 
 ## Architecture
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -9,4 +9,5 @@ configurations {
 
 dependencies {
   implementation("org.springframework.boot:spring-boot-starter-webflux")
+  implementation("org.springframework.boot:spring-boot-starter-oauth2-resource-server")
 }
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,17 +1,18 @@
 version: "3"
 services:
-  hmpps-interventions-service:
-    build:
-      context: .
-    network_mode: "host"
-    container_name: hmpps-interventions-service
+  hmpps-auth:
+    image: quay.io/hmpps/hmpps-auth:latest
+    networks:
+      - hmpps
+    container_name: hmpps-auth
     ports:
-      - "8081:8080"
+      - "8090:8080"
     healthcheck:
-      test: ["CMD", "curl", "-f", "http://localhost:8080/health/ping"]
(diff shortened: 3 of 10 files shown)


tomsmyers:
couple of "hygene" things
- please ensure PR titles are descriptive
- these three commits should be squashed down into a single commit, with a descriptive commit message. see [here](https://github.com/ministryofjustice/hmpps-architecture-as-code/blob/main/decisions/interventions/0003-use-git-commits-to-explain-our-work.md) for a (much better than the one i gave you) explanataion of why.

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/dto/DraftReferralTest.kt
```diff
@@ -0,0 +1,31 @@
...
+
+class DraftReferralTest : IntegrationTestBase() {
+
+  @Test
+  fun `create draft referral dto from empty referral`() {
+    val referral = Referral()
+    var draftReferral = DraftReferral(referral)
+
+    assertThat(draftReferral.id).isNull()
+    assertThat(draftReferral.created).isNull()
```
tomsmyers, line 18: i'm glad you wrote this test because it does raise an interesting point about if this draft referral is valid. 

should we enforce that draft referrals always have an id and created date by throwing a runtime exception if you attempt to create one without? e.g.

```
data class DraftReferral(
  val id: String,
  val created: String,
) {
  constructor(referral: Referral) : this(
    referral.id!!.toString(),
    DateTimeFormatter.ISO_INSTANT.format(referral.created!!.toInstant()),
  )
}
```

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/dto/DraftReferralTest.kt
```diff
@@ -0,0 +1,31 @@
...
+  }
+
+  @Test
+  fun `create draft referral dto from referral`() {
+    val referral = Referral()
+    referral.created = Date()
+    referral.id = UUID.randomUUID()
+    var draftReferral = DraftReferral(referral)
+
+    assertThat(draftReferral.id).isEqualTo(referral.id.toString())
```
tomsmyers, line 28: i really strive to avoid replicating the logic under test and instead test against concrete values e.g. 

```
    val date = SimpleDateFormat("yyyy-MM-dd HH:mm:ss").parse("2020-12-04 10:42:43")
    val id = UUID.fromString("70d3a47c-d539-4f76-8fc9-1c50e34aea29")
    var draftReferral = DraftReferral(Referral(created=date, id=id))

    assertThat(draftReferral.id).isEqualTo("70d3a47c-d539-4f76-8fc9-1c50e34aea29")
    assertThat(draftReferral.created).isEqualTo("2020-12-04T10:42:43Z")
```

see https://testing.googleblog.com/2014/07/testing-on-toilet-dont-put-logic-in.html for more on this

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/dto/DraftReferralTest.kt
karenmillermoj: I think you're right.  The draft referral needs both.

PR #21, opened 2020-12-04
IC-637: Notify Slack after deployments

sldblog:

Notify Slack after deployments.

They will look similar to

image

To raise awareness when deployments happen (and fail (although that does not seem to work yet)).


tomsmyers (approved):

sldblog:
Thank you!

PR #23, opened 2020-12-07
Draft Referral dto and Tests

karenmillermoj:

Adds a Draft Referral dto and its associated tests

Ensures a Draft Referral is created correctly from a Referral Entity object

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/controller/ReferralController.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/controller/ReferralController.kt
@@ -7,6 +7,7 @@ import org.springframework.web.bind.annotation.PathVariable
 import org.springframework.web.bind.annotation.PostMapping
 import org.springframework.web.bind.annotation.RestController
 import org.springframework.web.servlet.support.ServletUriComponentsBuilder
+import uk.gov.justice.digital.hmpps.hmppsinterventionsservice.dto.DraftReferral
 import uk.gov.justice.digital.hmpps.hmppsinterventionsservice.jpa.entity.Referral
 import uk.gov.justice.digital.hmpps.hmppsinterventionsservice.jpa.repository.ReferralRepository
 import java.util.UUID
@@ -14,7 +15,7 @@ import java.util.UUID
 @RestController
 class ReferralController(private val repository: ReferralRepository) {
   @PostMapping("/draft-referral")
-  fun createDraftReferral(): ResponseEntity<Referral> {
+  fun createDraftReferral(): ResponseEntity<DraftReferral> {
     val referral = Referral()
     repository.save(referral)
 
@@ -26,7 +27,7 @@ class ReferralController(private val repository: ReferralRepository) {
 
     return ResponseEntity
       .created(location)
-      .body(referral)
+      .body(DraftReferral(referral))
   }
 
   @GetMapping("/draft-referral/{id}")
@@ -38,7 +39,7 @@ class ReferralController(private val repository: ReferralRepository) {
     }
 
     return repository.findByIdOrNull(uuid)
-      ?.let { ResponseEntity.ok(it) }
+      ?.let { ResponseEntity.ok(DraftReferral(it)) }
       ?: ResponseEntity.notFound().build()
   }
 }
--- a/src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/dto/DraftReferralTest.kt
+++ b/src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/dto/DraftReferralTest.kt
@@ -1,0 +1,72 @@
+package uk.gov.justice.digital.hmpps.hmppsinterventionsservice.dto
+
+import com.fasterxml.jackson.module.kotlin.jacksonObjectMapper
+import org.assertj.core.api.Assertions.assertThat
+import org.junit.jupiter.api.Test
+import org.junit.jupiter.api.assertThrows
+import uk.gov.justice.digital.hmpps.hmppsinterventionsservice.integration.IntegrationTestBase
+import uk.gov.justice.digital.hmpps.hmppsinterventionsservice.jpa.entity.Referral
+import java.time.LocalDate
+import java.time.OffsetDateTime
+import java.util.UUID
+
+class DraftReferralTest : IntegrationTestBase() {
+
+  @Test
+  fun `empty input cannot create draft referral`() {
+    assertThrows<RuntimeException> {
+      DraftReferral(Referral())
+    }
+  }
+
+  @Test
+  fun `empty referral id cannot create draft referral`() {
+    val createdDate = OffsetDateTime.parse("2020-12-04T10:42:43+00:00")
+
+    assertThrows<RuntimeException> {
+      DraftReferral(Referral(created = createdDate))
+    }
+  }
+
+  @Test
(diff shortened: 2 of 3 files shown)


src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/dto/DraftReferralTest.kt
```diff
@@ -0,0 +1,64 @@
...
+import uk.gov.justice.digital.hmpps.hmppsinterventionsservice.integration.IntegrationTestBase
+import uk.gov.justice.digital.hmpps.hmppsinterventionsservice.jpa.entity.Referral
+import java.text.SimpleDateFormat
+import java.util.UUID
+
+class DraftReferralTest : IntegrationTestBase() {
+
+  @Test
+  fun `create draft referral dto from empty referral`() {
+    assertThrows<RuntimeException> {
```
sldblog, line 16: What do you think about adding the expected outcome into the test, so that the intent is clear from the test outline without looking at the test code?

E.g. `empty input cannot create draft referral`?

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/controller/ReferralController.kt
```diff
@@ -7,14 +7,15 @@ import org.springframework.web.bind.annotation.PathVariable
...
 import uk.gov.justice.digital.hmpps.hmppsinterventionsservice.jpa.entity.Referral
 import uk.gov.justice.digital.hmpps.hmppsinterventionsservice.jpa.repository.ReferralRepository
 import java.util.UUID
 
 @RestController
 class ReferralController(private val repository: ReferralRepository) {
   @PostMapping("/draft-referral")
-  fun createDraftReferral(): ResponseEntity<Referral> {
+  fun createDraftReferral(): ResponseEntity<DraftReferral> {
     val referral = Referral()
```
sldblog, line 19: `Referral` might be ambiguous soon when we end up having this, the draft and the sent referrals; would it be worth to call this a `ReferralInput`?

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/dto/DraftReferralTest.kt
```diff
@@ -0,0 +1,64 @@
...
+      DraftReferral(Referral())
+    }
+  }
+
+  @Test
+  fun `create draft referral dto with empty referral id`() {
+    val createdDate = SimpleDateFormat("yyyy-MM-dd HH:mm:ss").parse("2020-12-04 10:42:43")
+
+    assertThrows<RuntimeException> {
+      DraftReferral(Referral(created = createdDate))
```
sldblog, line 26: Would it be worth creating an "example valid" referral and setting its `id` to `null`? It may make the intent in the test description clearer (and not depend on remembering to add the other fields when they're added).

sldblog (commented):
A couple of questions for my curiosity :)

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/controller/ReferralController.kt
tomsmyers: i think internally the referral entity will remain the singular entity for storing all these type of referral, and the DTO will represent how the referral is presented to the API consumer. what do you think ?

sldblog: Right, I was thinking of something else, `Referral` is the persistence entity. No "Input" postfix needed 😄 

Model-wise, I think it depends.

1. If each entity is to be without a `state` column, then we may have a `sent_referrals` table that references a `draft_referral_id` and some additional fields like `sender`, `sent_at`, etc.

    That kind of structure would help us avoid having unrelated fields in persistence entities (and the unnecessary branching that comes with state/extra fields/validations).

1. On the other hand, `state` entities are simpler but require some extra semantics checks after deserialisation (is this entity a draft or sent referral?)

sldblog: I was gravitating towards (1) because the state is explicit in the code and it's more concise at the database layer. That said, I don't have more than just a gut feeling.

tomsmyers: let's discuss properly with the team. we can rename the table down the line with ease so we don't have to decide now.

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/dto/DraftReferralTest.kt
karenmillermoj: At the moment, only 2 fields are mandatory for a valid Referral.  I think this seems fine.   If this changes, I can update the test to reflect this (as you suggested).

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/dto/DraftReferralTest.kt
karenmillermoj: ok.  I'll update to make it clearer

tomsmyers (approved):

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/dto/DraftReferralTest.kt
```diff
@@ -0,0 +1,64 @@
...
+    val createdDate = SimpleDateFormat("yyyy-MM-dd HH:mm:ss").parse("2020-12-04 10:42:43")
+    val id = UUID.fromString("70d3a47c-d539-4f76-8fc9-1c50e34aea29")
+    val draftReferral = DraftReferral(Referral(id = id, created = createdDate))
+
+    assertThat(draftReferral.id).isEqualTo("70d3a47c-d539-4f76-8fc9-1c50e34aea29")
+    assertThat(draftReferral.created).isEqualTo("2020-12-04T10:42:43Z")
+  }
+
+  @Test
+  fun `determine json for draft referral dto`() {
```
sldblog, line 50: Out of curiosity, are we testing the serialisation to JSON to assert that the correct field formats are used?

sldblog:
_inlining errors on CI for visibility_
```
> Task :compileTestKotlin FAILED
e: /.../dto/DraftReferralTest.kt: (26, 40): Type mismatch: inferred type is Date! but OffsetDateTime? was expected
e: /.../dto/DraftReferralTest.kt: (43, 67): Type mismatch: inferred type is Date! but OffsetDateTime? was expected
e: /.../dto/DraftReferralTest.kt: (59, 86): Type mismatch: inferred type is Date! but OffsetDateTime? was expected
```

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/dto/DraftReferralTest.kt
sldblog: If yes, what do you think about testing this on unit level by testing the `dto.DraftReferral` construction? That'd avoid having to maintain a JSON payload here for all field changes and avoid a reasonably costly test case.

PR #25, opened 2020-12-07
IC-603: add initial flyway migration script for empty referral

tomsmyers:

adds an initial migration script for the existing referral entity which is super simple, and adds flyway config to run migrations on startup

prepare for running against a real database in our dev environment

## Related

- https://github.com/ministryofjustice/cloud-platform-environments/pull/3908


src/main/resources/application.yml
```diff
@@ -18,14 +18,13 @@ spring:
...
   datasource:
-    driver-class-name: org.postgresql.Driver
     url: jdbc:postgresql://${postgres.uri}/${postgres.db}
     username: ${postgres.username}
     password: ${postgres.password}
   jpa:
-    database-platform: org.hibernate.dialect.PostgreSQL94Dialect
     hibernate:
-      ddl-auto: update
+      show_sql: true
```
sldblog, line 26: Can this be configurable? (We want to avoid this on kubernetes environments, AFAIK)

src/main/resources/db/migration/V1_0__create_initial_referral.sql
```diff
@@ -0,0 +1 @@
+create table referral (id uuid not null, created timestamp, primary key (id));
```
sldblog, line 1: What do you think about auto-assigning a default primary key with `uuid_generate_v4()`, rather than relying on the application code to do that?

sldblog, line 1: Since we are going to store draft referrals in here, does it make sense to call it `draft_referrals`?

sldblog (commented):
Random questions, sorry -- I wonder if there's a better way to discuss these

sldblog:
Out of curiosity, do we know how will we kick off the migration locally, on docker-compose, and during deployment?

src/main/resources/application.yml
tomsmyers: this was left in by mistake. will remove. good spot.

src/main/resources/db/migration/V1_0__create_initial_referral.sql
tomsmyers: i left a comment on another PR with some more discussion on this point. we can dicuss in that thread.

https://github.com/ministryofjustice/hmpps-interventions-service/pull/23#discussion_r537589834

tomsmyers: i think i'd prefer to let hibernate do as much of this as possible, e.g. also auto-created timestamps etc, rather than by hand in these files.

sldblog: 👍 commented there, thanks!

tomsmyers:
> Out of curiosity, do we know how will we kick off the migration locally, on docker-compose, and during deployment?

the migrations run on application start in all three scenarios

sldblog:
> the migrations run on application start in all three scenarios

Is that safe in case more than one pod boots up at the same time?

tomsmyers:
> We might want to pull the migration script out in a separate init container: https://kubernetes.io/docs/concepts/workloads/pods/init-containers/#init-containers-in-use

yeh i'd be up for something like this. i'll open a ticket to look into it later.

src/main/resources/db/local/V99_0__draft_referrals.sql
```diff
@@ -0,0 +1,4 @@
+insert into referral (id, created, completion_deadline)
+values ('F7B4A237-94CC-448F-A2C0-CDFFF17B58B9', '2020-12-07 18:02:01.599803+00', '2021-02-14'),
+       ('71E6EAF6-5899-4A6B-9651-1AA14B3843A3', '2020-12-07 20:45:21.986389+00', '2021-03-01');
```
sldblog, line 3: _note, not a blocker_
This looks like a place where a database-generated default for the UUID and timestamp might be useful -- we could seed the relevant fields without having to pre-generate values for the plumbing.

sldblog (approved):

src/main/resources/db/local/V99_0__draft_referrals.sql
tomsmyers: agree. but having the abililty to add records with known uuids, or created dates in the past is a benefit.

sldblog: true

PR #26, opened 2020-12-07
add helm env vars for dev postgres instance

tomsmyers:

add helm env vars for dev postgres instance

get the dev deploy hooked up to the new rds instance

## Related

https://github.com/ministryofjustice/cloud-platform-environments/pull/3908

### DIFF
--- a/helm_deploy/hmpps-interventions-service/templates/_envs.tpl
+++ b/helm_deploy/hmpps-interventions-service/templates/_envs.tpl
@@ -16,4 +16,28 @@ env:
   - name: HMPPSAUTH_BASEURL
     value: "{{ .Values.env.HMPPSAUTH_BASEURL }}"
 
+  - name: POSTGRES_URI
+    valueFrom:
+      secretKeyRef:
+        name: postgres
+        key: rds_instance_endpoint
+
+  - name: POSTGRES_DB
+    valueFrom:
+      secretKeyRef:
+        name: postgres
+        key: database_name
+
+  - name: POSTGRES_USERNAME
+    valueFrom:
+      secretKeyRef:
+        name: postgres
+        key: database_username
+
+  - name: POSTGRES_PASSWORD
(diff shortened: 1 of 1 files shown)


helm_deploy/hmpps-interventions-service/templates/_envs.tpl
```diff
@@ -16,4 +16,28 @@ env:
   - name: HMPPSAUTH_BASEURL
     value: "{{ .Values.env.HMPPSAUTH_BASEURL }}"
 
+  - name: POSTGRES_URI
+    valueFrom:
+      secretKeyRef:
+        name: postgres
+        key: rds_instance_endpoint
```
sldblog, line 23: Could `POSTGRES_URI` point to the [`url`](https://github.com/ministryofjustice/cloud-platform-environments/blob/1d3a8eed05977fd71bcd09272b7eb507c6cf5817/namespaces/live-1.cloud-platform.service.justice.gov.uk/hmpps-interventions-dev/resources/rds.tf#L33
) variable? That way we could save the db/username/password variables.

sldblog: Ah okay, I can see it's used in constructing a URL
```
 21     driver-class-name: org.postgresql.Driver
 22     url: jdbc:postgresql://${postgres.uri}/${postgres.db}
 23     username: ${postgres.username}
 24     password: ${postgres.password}
```

sldblog (approved):

helm_deploy/hmpps-interventions-service/templates/_envs.tpl
sldblog: `postgres.uri` has connotations of [`URI = scheme:[//authority]path[?query][#fragment]`](https://tools.ietf.org/html/rfc3986#section-1.1.1), which confused me

PR #27, opened 2020-12-08
add all required data fields to referral object and create db migrations

tomsmyers:

adds the required data fields for draft referrals based on this document https://docs.google.com/spreadsheets/d/1gVKJpBklpx0XN-MG6igylE5KMkhwZJFf60NjTUP8zl4/edit#gid=1184237857

add the flyway scripts to update the production db with the required columns.

alters the type of `completionDeadline` from `timestamp with time zone` to `date`.

adds some sample data to the local db for running locally.

continue to build out the referrals api.


sldblog (changes requested):
I don't think we should add any of these fields until we actually use them through the client. Until then, it's undefined data without a use case -- so it will likely change, so I don't see value in adding them now.

tomsmyers:
> I don't think we should add any of these fields until we actually use them through the client. Until then, it's undefined data without a use case -- so it will likely change, so I don't see value in adding them now.

fair enough - personally, i think the use case is clear enough since these fields are contractually required. i also think having all the fields there _before_ the front end needs to use them is beneficial since there is no 'catch up' required once they are added there.

PR #29, opened 2020-12-09
IC-642: Add provider-side pact verification

sldblog:

Adds a `PactTest` to the test suite that verifies the publisher against pacts created by the consumers.

To ensure both sides can agree on contracts defined [here](https://pact-broker-prod.apps.live-1.cloud-platform.service.justice.gov.uk/pacts/provider/Interventions%20Service/consumer/Interventions%20UI/latest).

## What's outstanding?

- [ ] I'm unhappy with 93a0d3d3c2272eca8110612443eca383ef3488c1, there has to be a better way to pre-authenticate Pact requests or disable a filter in the request filter chain

    I've tried MockMvc tests but that does not seem to play nice, but might be my unfamiliarity with Boot.

- [x] Publishing the results on CI

_Optional_

- [x] Make the Pact test runnable with a single `./gradlew` task, where we can (later) configure the consumer tag/version to use.
    
    Context: we want consumer-made pacts to also trigger the provider verification against a specific consumer pact version and that shouldn't run the entire test suite.


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/controller/ReferralController.kt
```diff
@@ -41,4 +43,21 @@ class ReferralController(private val repository: ReferralRepository) {
...
+  fun patchDraftReferralByID(@PathVariable id: String, @RequestBody partialUpdate: Referral): ResponseEntity<Any> {
+    val uuid = try {
+      UUID.fromString(id)
+    } catch (e: IllegalArgumentException) {
+      return ResponseEntity.badRequest().body("malformed id")
+    }
+
+    return repository.findByIdOrNull(uuid)
+      ?.let {
+        it.completionDeadline = partialUpdate.completionDeadline
```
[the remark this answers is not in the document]
sldblog: 👍 done in d4ee226

This made me think...

As a REST API user, I'd expect `PATCH /draft-referral/:uuid` to reset the field, if the `completionDeadline` is explicitly specified, i.e. when the payload is `{"completionDeadline": null}`.

Is the presence of a key in the request possible to detect with a `@RequestBody` mapping? It would be great to throw back a validation error if we don't accept `null` values, rather than skipping them.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/controller/ReferralController.kt
```diff
@@ -41,4 +43,21 @@ class ReferralController(private val repository: ReferralRepository) {
...
+  fun patchDraftReferralByID(@PathVariable id: String, @RequestBody partialUpdate: Referral): ResponseEntity<Any> {
+    val uuid = try {
+      UUID.fromString(id)
+    } catch (e: IllegalArgumentException) {
+      return ResponseEntity.badRequest().body("malformed id")
+    }
+
+    return repository.findByIdOrNull(uuid)
+      ?.let { ref ->
+        partialUpdate.completionDeadline?.let { ref.completionDeadline = it }
```
sldblog, line 57: Is this idiomatic Kotlin/Spring Boot?

.circleci/config.yml
```diff
@@ -11,12 +11,10 @@ _snyk_options: &snyk_options
 
 jobs:
   validate:
-    executor: hmpps/java
```
tomsmyers, line 14: this line pulls in the JVM configuration and other bits and pieces defined for the executor. would prefer to leave in.

.circleci/config.yml
```diff
@@ -37,6 +35,33 @@ jobs:
       - store_artifacts:
           path: build/reports/tests
 
+  validate_and_publish_pact:
+    environment:
+      PACT_BROKER_HOST: "pact-broker-prod.apps.live-1.cloud-platform.service.justice.gov.uk"
+      PACT_BROKER_USERNAME: "interventions"
+    docker:
+      - image: cimg/openjdk:11.0
```
tomsmyers, line 43: same thing here, using the hmpps/java executor allows us to share the config.

.circleci/config.yml
```diff
@@ -64,10 +89,8 @@ workflows:
   version: 2
   build_test_and_deploy:
     jobs:
-      - validate:
-          filters:
-            tags:
-              ignore: /.*/
+      - validate
+      - validate_and_publish_pact
```
tomsmyers, line 93: i think the name of this job is a bit confusing with the repetition of 'validate', it makes it sound like it does the same as 'validate' + some more stuff.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/config/ResourceServerConfiguration.kt
```diff
@@ -11,13 +12,24 @@ import org.springframework.security.oauth2.server.resource.authentication.JwtGra
 
 @Configuration
 @EnableWebSecurity
-class ResourceServerConfiguration : WebSecurityConfigurerAdapter() {
-  // todo: add custom token validator which calls token verification service
+@Profile("test")
+class BasicResourceServerConfiguration : WebSecurityConfigurerAdapter() {
```
tomsmyers, line 16: i understand why this is here, but it makes me nervous of deploying the app with this config enabled by mistake! is there any other way to alter this config in the test code?

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/controller/ReferralController.kt
```diff
@@ -42,4 +45,21 @@ class ReferralController(private val repository: ReferralRepository) {
...
+    val uuid = try {
+      UUID.fromString(id)
+    } catch (e: IllegalArgumentException) {
+      return ResponseEntity.badRequest().body("malformed id")
+    }
+
+    return repository.findByIdOrNull(uuid)
+      ?.let {
+        ReferralService.updateReferral(it, partialUpdate)
+        repository.save(it)
```
tomsmyers, line 60: can we delegate the repository interactions to the service layer?

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/jpa/service/ReferralService.kt
```diff
@@ -0,0 +1,13 @@
+package uk.gov.justice.digital.hmpps.hmppsinterventionsservice.jpa.service
+
+import uk.gov.justice.digital.hmpps.hmppsinterventionsservice.dto.DraftReferral
+import uk.gov.justice.digital.hmpps.hmppsinterventionsservice.jpa.entity.Referral
+
+object ReferralService {
+  fun updateReferral(original: Referral?, update: DraftReferral): Referral? {
```
tomsmyers, line 7: `original` can be `Referral` without the nullable qualifier. we are checking for null at the call site (line 58 in ReferralController.kt)

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/jpa/service/ReferralService.kt
```diff
@@ -0,0 +1,13 @@
...
+    original?.let {
```
tomsmyers, line 8: and thus this can be removed

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/jpa/service/ReferralServiceTest.kt
```diff
@@ -0,0 +1,36 @@
+package uk.gov.justice.digital.hmpps.hmppsinterventionsservice.jpa.service
+
+import org.assertj.core.api.Assertions
+import org.junit.jupiter.api.Test
+import uk.gov.justice.digital.hmpps.hmppsinterventionsservice.dto.DraftReferral
+import uk.gov.justice.digital.hmpps.hmppsinterventionsservice.jpa.entity.Referral
+import java.time.LocalDate
+
+class ReferralServiceTest {
```
tomsmyers, line 9: please can we add tests which specify that certain fields (namely 'id' and 'created') cannot be overwritten by updates?

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/config/ResourceServerConfiguration.kt
sldblog: I couldn't find any. I'm sure there are ways, but I'm not familiar with Spring Boot 😞

.circleci/config.yml
sldblog: 👍 Ah, I had a schema validator tell me it's one or the other, you're right.

TIL the expanded version gets an extra `environment` after disposing of the [original image](https://github.com/ministryofjustice/hmpps-circleci-orb/blob/aa63b0b076d564ba1b4e933fdda7f75065dfa75f/src/executors/java.yml) defined in the executor.
```
$ circleci config process .circleci/config.yml
<snip>
  validate:
    docker:
    - image: cimg/openjdk:11.0
      environment:
        POSTGRES_DB: circle_test
    - image: circleci/postgres:13-ram
      environment:
        POSTGRES_PASSWORD: password
<snip>
    environment:
    - _JAVA_OPTIONS: -Xmx256m -XX:ParallelGCThreads=2 -XX:ConcGCThreads=2 -XX:ParallelGCThreads=2 -Djava.util.concurrent.ForkJoinPool.common.parallelism=2 -Dorg.gradle.daemon=false -Dorg.gradle.jvmargs=-XX:+UseContainerSupport
```

It's back now.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/controller/ReferralController.kt
sldblog: yes -- could it be done separately from this PR?

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/jpa/service/ReferralService.kt
sldblog: done in 82211c3015da4b02b8c2a6a035563e856c102969

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/jpa/service/ReferralServiceTest.kt
sldblog: done in 82211c3015da4b02b8c2a6a035563e856c102969
```
@Test
fun `update cannot overwrite identifier fields`() { ... }
```

.circleci/config.yml
sldblog: 21c0b1525959201bf3fc4c5e96ff801209badbc0 has `pact_check_and_publish`, what do you think?

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/config/ResourceServerConfiguration.kt
sldblog: 💭 Maybe I can flip this and move to the test source set. ⌛

sldblog: Is this a blocker? I'm happy to commit to dig into this and arrive at a sensible thing separately.

sldblog: @tomsmyers Is this a blocker? I'm happy to commit to dig into this and arrive at a sensible thing separately.

tomsmyers: no will address seperately

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/controller/ReferralController.kt
tomsmyers: yep. karen looking at this this afternoon 👍

tomsmyers (approved):
🤘🏼

src/main/resources/db/local/V99_0__draft_referrals.sql
```diff
@@ -1,4 +1,3 @@
 insert into referral (id, created, completion_deadline)
-values ('F7B4A237-94CC-448F-A2C0-CDFFF17B58B9', '2020-12-07 18:02:01.599803+00', '2021-02-14'),
-       ('71E6EAF6-5899-4A6B-9651-1AA14B3843A3', '2020-12-07 20:45:21.986389+00', '2021-03-01');
-
+values ('ac386c25-52c8-41fa-9213-fcf42e24b0b5', '2020-12-07 18:02:01.599803+00', '2021-02-14'),
+       ('dfb64747-f658-40e0-a827-87b4b0bdcfed', '2020-12-07 20:45:21.986389+00', '2021-03-01');
```
sldblog, line 3: cc @green-digger -- I may need to remove `local` from the codeowners for data engineering 🤔

PR #31, opened 2020-12-10
add 'createdByUserId' to draft referral creation 

tomsmyers:

gets the user id from from auth token for referral creation  which sets the 'createdByUserId' field in the referral entity.

also adds simple repository method to find referrals by username and a test for this.

changes the permissions from custom `ROLE_INTERVENTIONS` to `ROLE_PROBATION` to allow delius users access by default.

allows us to track who has created referrals and allows us to search for referrals by username.

### DIFF
--- a/src/main/resources/db/migration/V1_1__create_user_table.sql
+++ b/src/main/resources/db/migration/V1_1__create_user_table.sql
@@ -1,0 +1,2 @@
+alter table referral
+    add column created_by_user_id text;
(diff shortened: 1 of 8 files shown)


karenmillermoj (approved):
This looks fine to me.  I'd usually go with just createdBy, but that's just personal preference - happy to accept either.

tomsmyers:
> This looks fine to me. I'd usually go with just createdBy, but that's just personal preference - happy to accept either.

agree, that's what i started with, but since this field is just a string i wanted to specify what the column represented, as oppsed to `createdByUserId` for example

src/main/resources/db/migration/V1_1__create_user_table.sql
```diff
@@ -0,0 +1,2 @@
+alter table referral
+    add column created_by_user_id text;
```
sldblog, line 2: Given most of our operations will be scoped to the user on the probation side, can we add an index too, please?

sldblog, line 2: If we expect multiple auth sources, I think we'd need a way to signal to the consumers of our data the source of the identity. To somehow to mark that this `user_id` is "from delius", etc.

Random idea: we could use JSONB columns with `{"id": "12345", "auth_source": "delius"}` content. Or any other, more sensible options :)

tomsmyers:
this required more work to get the pacts working with auth - closing for now

PR #33, opened 2020-12-10
Downgrade PostgreSQL versions to match target env

sldblog:

Use PostgreSQL 11 across everything.

We had to provision PostgreSQL 11 on the (Kubernetes) development environment: this PR is matching all versions of Postgres to avoid surprises.

Reference: https://github.com/ministryofjustice/cloud-platform-environments/pull/3914


sldblog:
Drat, it was PostgreSQL 10... not 11

PR #35, opened 2020-12-14
Move repository calls into service layer and update tests

karenmillermoj:

Moves references to the repository files from the controller into the service layer

Provide a cleaner architecture of the code to separate the appropriate logic into the correct layers

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/controller/ReferralController.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/controller/ReferralController.kt
@@ -1,6 +1,5 @@
 package uk.gov.justice.digital.hmpps.hmppsinterventionsservice.controller
 
-import org.springframework.data.repository.findByIdOrNull
 import org.springframework.http.ResponseEntity
 import org.springframework.web.bind.annotation.GetMapping
 import org.springframework.web.bind.annotation.PatchMapping
@@ -10,18 +9,16 @@ import org.springframework.web.bind.annotation.RequestBody
 import org.springframework.web.bind.annotation.RestController
 import org.springframework.web.servlet.support.ServletUriComponentsBuilder
 import uk.gov.justice.digital.hmpps.hmppsinterventionsservice.dto.DraftReferral
-import uk.gov.justice.digital.hmpps.hmppsinterventionsservice.jpa.entity.Referral
-import uk.gov.justice.digital.hmpps.hmppsinterventionsservice.jpa.repository.ReferralRepository
 import uk.gov.justice.digital.hmpps.hmppsinterventionsservice.jpa.service.ReferralService
 import java.util.UUID
 
 @RestController
-class ReferralController(private val repository: ReferralRepository) {
+class ReferralController(private val referralService: ReferralService) {
+
   @PostMapping("/draft-referral")
   fun createDraftReferral(): ResponseEntity<DraftReferral> {
-    val referral = Referral()
-    repository.save(referral)
 
+    val referral = referralService.createDraftReferral()
     val location = ServletUriComponentsBuilder
       .fromCurrentRequest()
       .path("/{id}")
@@ -41,7 +38,7 @@ class ReferralController(private val repository: ReferralRepository) {
       return ResponseEntity.badRequest().body("malformed id")
     }
 
-    return repository.findByIdOrNull(uuid)
+    return referralService.getDraftReferral(uuid)
       ?.let { ResponseEntity.ok(DraftReferral(it)) }
       ?: ResponseEntity.notFound().build()
   }
@@ -54,11 +51,9 @@ class ReferralController(private val repository: ReferralRepository) {
       return ResponseEntity.badRequest().body("malformed id")
     }
 
-    return repository.findByIdOrNull(uuid)
+    return referralService.updateDraftReferral(uuid, partialUpdate)
       ?.let {
-        ReferralService.updateReferral(it, partialUpdate)
-        repository.save(it)
-        ResponseEntity.ok(it)
+        ResponseEntity.ok(DraftReferral(it))
       }
       ?: ResponseEntity.notFound().build()
   }
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/jpa/service/ReferralService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/jpa/service/ReferralService.kt
@@ -1,11 +1,30 @@
 package uk.gov.justice.digital.hmpps.hmppsinterventionsservice.jpa.service
 
+import org.springframework.data.repository.findByIdOrNull
+import org.springframework.stereotype.Service
 import uk.gov.justice.digital.hmpps.hmppsinterventionsservice.dto.DraftReferral
 import uk.gov.justice.digital.hmpps.hmppsinterventionsservice.jpa.entity.Referral
+import uk.gov.justice.digital.hmpps.hmppsinterventionsservice.jpa.repository.ReferralRepository
+import java.util.UUID
 
-object ReferralService {
-  fun updateReferral(original: Referral, update: DraftReferral): Referral {
-    update.completionDeadline?.let { original.completionDeadline = it }
-    return original
+@Service
+class ReferralService(val repository: ReferralRepository) {
+
+  fun createDraftReferral(): Referral {
+    return repository.save(Referral())
+  }
+
+  fun getDraftReferral(id: UUID): Referral? {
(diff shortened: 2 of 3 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/controller/ReferralController.kt
```diff
@@ -41,7 +38,7 @@ class ReferralController(private val repository: ReferralRepository) {
       return ResponseEntity.badRequest().body("malformed id")
     }
 
-    return repository.findByIdOrNull(uuid)
+    return referralService.getReferral(uuid)
```
tomsmyers, line 41: same here - `getDraftReferral`

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/controller/ReferralController.kt
```diff
@@ -54,11 +51,9 @@ class ReferralController(private val repository: ReferralRepository) {
       return ResponseEntity.badRequest().body("malformed id")
     }
 
-    return repository.findByIdOrNull(uuid)
+    return referralService.updateReferral(uuid, partialUpdate)
```
tomsmyers, line 54: and here

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/jpa/service/ReferralService.kt
```diff
@@ -1,11 +1,31 @@
...
+    return repository.save(Referral())
+  }
+
+  fun getReferral(id: UUID): Referral? {
+    return repository.findByIdOrNull(id)
+  }
+
+  fun updateReferral(id: UUID, update: DraftReferral): Referral? {
+    val ref = repository.findByIdOrNull(id)
+    if (ref != null) {
```
tomsmyers, line 23: nitpick:

in general i prefer 'guard clauses' to avoid ever expanding levels of indentation

i.e. `if (ref == null) return null`

however kotlin has a handy helper for this type of situation. this can be written as:

```
val ref = getReferral(id) ?: return null 

...
```

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/jpa/service/ReferralServiceTest.kt
```diff
@@ -1,57 +1,71 @@
...
+  val entityManager: TestEntityManager,
+  val referralRepository: ReferralRepository
+) {
+
+  private val referralService = ReferralService(referralRepository)
 
   @Test
   fun `update cannot overwrite identifier fields`() {
-    val update = DraftReferral(
-      id = UUID.fromString("ce364949-7301-497b-894d-130f34a98bff"),
```
tomsmyers, line 24: we have deleted the id field from the incoming draft referral here.

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/jpa/service/ReferralServiceTest.kt
```diff
@@ -1,57 +1,71 @@
...
-    val updated = ReferralService.updateReferral(newOriginal(), update)
-    assertThat(updated.completionDeadline).isEqualTo(LocalDate.parse("2021-06-26"))
+    val updated = referralService.updateReferral(referral.id!!, draftReferral)
+    assertThat(updated!!.completionDeadline).isEqualTo(LocalDate.parse("2021-06-26"))
   }
 
   @Test
   fun `non-null fields in the update overwrite original fields`() {
-    val update = DraftReferral(completionDeadline = LocalDate.parse("2020-12-01"))
+    val referral = Referral(completionDeadline = LocalDate.of(2021, 6, 26))
```
tomsmyers, line 50: nitpick (not a blocker): would be good to stick with one or other of `LocalDate.of` or `LocalDate.parse`

tomsmyers (commented):
couple of small things in the comments. 

also: 
- we could use a test to validate that the `update` method persists the updated referral. 
- there are no test for `create` or `get`

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/jpa/service/ReferralServiceTest.kt
karenmillermoj: You're right.  I prefer the LocalDate.of method, updated all.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/jpa/service/ReferralService.kt
karenmillermoj: I'm happy either way, updated

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/controller/ReferralController.kt
```diff
@@ -10,18 +9,16 @@ import org.springframework.web.bind.annotation.RequestBody
...
 @RestController
-class ReferralController(private val repository: ReferralRepository) {
+class ReferralController(private val referralService: ReferralService) {
+
   @PostMapping("/draft-referral")
   fun createDraftReferral(): ResponseEntity<DraftReferral> {
-    val referral = Referral()
-    repository.save(referral)
 
+    val referral = referralService.createReferral()
```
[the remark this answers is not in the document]
karenmillermoj: Updated to use DraftReferral

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/controller/ReferralController.kt
karenmillermoj: Updated to use DraftReferral

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/jpa/service/ReferralServiceTest.kt
karenmillermoj: Well spotted, was unsure whether to just set the id passed in, into the DraftReferral, but then decided to just pass separately.  Added id back into test

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/jpa/service/ReferralServiceTest.kt
```diff
@@ -1,57 +1,111 @@
...
+  @Test
+  fun `update successfully persists the updated draft referral`() {
+    val referral = Referral(completionDeadline = LocalDate.of(2021, 6, 26))
+    entityManager.persist(referral)
+    entityManager.flush()
+
+    val draftReferral = DraftReferral(completionDeadline = LocalDate.of(2020, 12, 1))
+
+    val updated = referralService.updateDraftReferral(referral.id!!, draftReferral)
+    assertThat(updated!!.id).isEqualTo(referral.id)
```
tomsmyers, line 85: in this test we aren't actually testing that the update has been saved to the database. i think we need to make a call to `getDraftReferral` to check.

tomsmyers (approved):
sweet, thanks!

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/jpa/service/ReferralServiceTest.kt
karenmillermoj: Removed check against returned update to persisted object from getDraftReferral

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/controller/ReferralController.kt
```diff
@@ -54,11 +51,9 @@ class ReferralController(private val repository: ReferralRepository) {
       return ResponseEntity.badRequest().body("malformed id")
     }
 
-    return repository.findByIdOrNull(uuid)
+    return referralService.updateDraftReferral(uuid, partialUpdate)
```
sldblog, line 54: (Apologies if this is out of context)

It seems like this `update` signature does two things:
- finds a referral by ID or shortcut to null
- updates the referral and returns the updated object

To avoid having to change both `getDraftReferral` and this function when the lookup logic changes, what do you think about having one function for each of these responsibilities and composing them?

PR #37, opened 2020-12-15
change pact config to stop tests blocking merges and deploys

tomsmyers:

change pact config to stop tests blocking merges and deploys

see context on this here https://mojdt.slack.com/archives/C01DYKJUKDX/p1607961388144600 and here https://mojdt.slack.com/archives/C01DYKJUKDX/p1607966891150500


Gweaton:
So, does this mean we now need to tag our pacts as `last-implemented` when we push new Pact tests?

sldblog:
@Gweaton I think the semantic is that the provider side moves the tag when it's ready to implement a new set of pacts. Would that work?

sldblog (approved):

PR #38, opened 2020-12-15
IC-652: Switch to a nightly built JRE 11 alpine base image

sldblog:

Switch to `adoptopenjdk/openjdk11:alpine-jre` for the final base image.

Reduce vulnerabilities by using a frequently patched base image.

A secondary benefit is the reduced size of the final image.

**Before**
image

**After**
image


tomsmyers (approved):
thanks for this. should we remove the ignored snyk vulns as part of the same PR?

tomsmyers:
doh - missed the file deletion. thanks

PR #39, opened 2020-12-15
remove unused code / allow any authenticated user to interact with the backend 

tomsmyers:

- removes the unused controller code that was originally there for demo purposes.
- removes the requirement for `ROLE_INTERVENTIONS` to interact with the backend 

a) reduce tech debt
b) allow other members of the team to easily use the service in the dev environment
c) defer any role management at all until we know exactly what is required


karenmillermoj (approved):
Looks good

PR #40, opened 2020-12-15
add db column for storing user ID, and associated endpoint for looking up referrals by ID

tomsmyers:

add db column for storing user ID, and associated endpoint for looking up referrals by ID. no changes to existing api endpoints in this PR.

working towards IC-680

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/jpa/repository/ReferralRepository.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/jpa/repository/ReferralRepository.kt
@@ -4,4 +4,6 @@ import org.springframework.data.repository.CrudRepository
 import uk.gov.justice.digital.hmpps.hmppsinterventionsservice.jpa.entity.Referral
 import java.util.UUID
 
-interface ReferralRepository : CrudRepository<Referral, UUID>
+interface ReferralRepository : CrudRepository<Referral, UUID> {
+  fun findByCreatedByUserID(userId: String): List<Referral>
+}
--- a/src/main/resources/db/migration/V1_1__add_created_by_user_id.sql
+++ b/src/main/resources/db/migration/V1_1__add_created_by_user_id.sql
@@ -1,2 +1,4 @@
 alter table referral
-    add column created_by_userid text;
+    add column created_by_userid varchar(255);
+
+create index IX_referral_created_by_userid on referral (created_by_userid);
--- a/src/main/resources/db/migration/V1_1__add_created_by_user_id.sql
+++ b/src/main/resources/db/migration/V1_1__add_created_by_user_id.sql
@@ -1,0 +1,2 @@
+alter table referral
+    add column created_by_userid text;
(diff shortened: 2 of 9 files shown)


karenmillermoj (approved):
Everything looks good

src/main/resources/db/migration/V1_1__add_created_by_user_id.sql
```diff
@@ -0,0 +1,2 @@
+alter table referral
+    add column created_by_userid text;
```
sldblog, line 2: For the data users we will have to establish where these users come from. Is it possible to capture the auth source here, either as a separate column, table or JSONB payload?

sldblog (commented):
LGTM, one question

src/main/resources/db/migration/V1_1__add_created_by_user_id.sql
sldblog: Or by encoding it into the name and switching to integer?

`created_by_delius_user_id integer`?

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/jpa/repository/ReferralRepository.kt
```diff
@@ -4,4 +4,6 @@ import org.springframework.data.repository.CrudRepository
 import uk.gov.justice.digital.hmpps.hmppsinterventionsservice.jpa.entity.Referral
 import java.util.UUID
 
-interface ReferralRepository : CrudRepository<Referral, UUID>
+interface ReferralRepository : CrudRepository<Referral, UUID> {
+  fun findByCreatedByUserID(userId: String): List<Referral>
```
sldblog, line 8: Can we create an index for this lookup, please? It'll be a major use case to filter the table by user.

sldblog (commented):
Oops, forgot an index question

src/main/resources/db/migration/V1_1__add_created_by_user_id.sql
tomsmyers: do we need to store that info, or just output in our events?

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/jpa/repository/ReferralRepository.kt
tomsmyers: yep, i'll create a ticket and address is a seperate PR

tomsmyers: scratch that, i'll just do it now.

src/main/resources/db/migration/V1_1__add_created_by_user_id.sql
sldblog: I think so... In this use case:
- Someone looking at our data outside our application for reporting purposes
- They need to establish who this identity is
- We don't have a user table -- so this is an external reference -- but there's no context where it points

sldblog (approved):
👍 LGTM

I think the user context is worth discussing but isn't a blocker

PR #41, opened 2020-12-16
Change resource server config to always require auth tokens, but relax verification for testing 

tomsmyers:

Change resource server config to always require auth tokens, but relaxes token verification for testing 

also reverts 76d02d7

allow us to test endpoints that require tokens (e.g. endpoints that need to know the user accessing them) with pact


karenmillermoj (approved):
lgtm

PR #44, opened 2020-12-23
Add sql definitions and initial descriptions of complexity levels for existing service categories

tomsmyers:

Add sql definitions and initial descriptions of complexity levels for existing service categories

satisfy more front end pacts and seed the db with the complexity level data.

--

Unfortunately this PR also contains some light refactoring of the service category entity
and service to remove unused code and improve consistency.

The new endpoint satisfies pact 'a GET request to fetch the service category'.

Thje complexity level data was taken from the Dynamic Framework Schedule 2.1 document dated 10 June 2020.


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/dto/ServiceCategoryDTO.kt
```diff
@@ -0,0 +1,21 @@
+package uk.gov.justice.digital.hmpps.hmppsinterventionsservice.dto
+
+import uk.gov.justice.digital.hmpps.hmppsinterventionsservice.jpa.entity.ComplexityLevel
+import uk.gov.justice.digital.hmpps.hmppsinterventionsservice.jpa.entity.ServiceCategory
+import java.util.UUID
+
+data class ServiceCategoryDTO(
```
tomsmyers, line 7: this looks like a waste of time, but a subtle difference in sending this over the wire vs the entity object is that there is no 'created' field, which is not required by the client.

src/main/resources/application-sqlgen.yml
```diff
@@ -0,0 +1,14 @@
+spring:
```
tomsmyers, line 1: this config is to allow us to easily see what hibernate generates for new entities instead of just writing it all ourselves from scratch.

src/test/resources/application-jpa-test.yml
```diff
@@ -1,4 +1,6 @@
 spring:
+  flyway:
```
tomsmyers, line 2: this stops the migrations being applied for this simple jpa tests with the mock database.

src/main/resources/db/migration/V1_4_0__complexity_level_definitions.sql
```diff
@@ -0,0 +1,16 @@
+create table complexity_level(
+    id uuid not null,
+    title text not null,
+    description text not null,
+    primary key (id)
+);
+
+create table service_category_complexity_levels(
+    service_category_id uuid not null,
+    complexity_levels_id uuid not null
```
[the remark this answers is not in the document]
tomsmyers: this is the JPA naming conventions - saves me annotating the models with custom column names

src/main/resources/db/migration/V1_4_1__complexity_level_data.sql
```diff
@@ -0,0 +1,103 @@
...
+insert into service_category_complexity_levels (service_category_id, complexity_levels_id)
+values ('96a63c39-4371-4f17-a6ec-265755f0cf7b', '178ec337-0ec9-422c-b9da-ea969fdcf98a'),
+       ('96a63c39-4371-4f17-a6ec-265755f0cf7b', 'd771ae5e-01d4-43b0-a941-99b2c0c52d29'),
+       ('96a63c39-4371-4f17-a6ec-265755f0cf7b', '65a682a7-9621-4dab-8040-08f3109321f8');
+
+-- Dependency and Recovery (D&R) Service Category:
+-- a) the complexity levels
+insert into complexity_level (id, title, description)
+values ('5392858e-3432-40cb-b277-4552e8ac4638', 'Low complexity', 'Service User who is highly motivated and have had sustained period of abstinence. Service User has successfully engaged in a period of treatment but requires support to sustain recovery.'),
+       ('f6a063e4-7daa-4cf6-9042-22476f8b5ed9', 'Medium complexity', 'Services Users who has had lapses/crises that may lead to relapse and/or is at a transition point, e.g. from prison into community. Service User has a moderate level of motivation but requires a high level of support.'),
```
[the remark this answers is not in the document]
tomsmyers: good point. this is copied directly. but we're going to get this information audited by our data people closer to release to fix up any issues like this.

src/main/resources/db/migration/V1_4_0__complexity_level_definitions.sql
```diff
@@ -0,0 +1,16 @@
+create table complexity_level(
+    id uuid not null,
+    title text not null,
```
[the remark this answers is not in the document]
tomsmyers: hmm yeh, i asked about this at some point previously and there wasn't certainty that this value will always be one of the three types you see here. i think leaving as text gives us more flexibility.

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/jpa/service/ServiceCategoryServiceTest.kt
```diff
@@ -7,23 +7,33 @@ import org.springframework.boot.test.autoconfigure.orm.jpa.TestEntityManager
...
+
+  @Test
+  fun `get valid service category`() {
+    val idToFind = UUID.randomUUID()
+
+    val categories = listOf(
+      ServiceCategory(name = "Accomodation", id = idToFind, created = OffsetDateTime.now(), complexityLevels = emptyList()),
+      ServiceCategory(name = "Accomodation", id = UUID.randomUUID(), created = OffsetDateTime.now(), complexityLevels = emptyList())
+    )
     categories.forEach { entityManager.persist(it) }
```
mxco86, line 34: This can be converted to a .map() on the list which removes the need for the 'categories' variable

mxco86 (approved):
Looks good to me - just a small style comment on one of the tests

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/dto/ServiceCategoryDTO.kt
mxco86: I don't think it's a bad idea to decouple from the database entity as this might expand over time

green-digger (approved):

PR #46, opened 2020-12-24
Allow updates to a referral's complexity level 

tomsmyers:

- some refactoring - please review commit by commit.
- adds fields for service category and complexity level to the referral entity and DTO.
- allows updating complexity level in the PATCH /draft-referral/{id} endpoint

continue to get the API more up to date with the front end consumer

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/controller/ReferralController.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/controller/ReferralController.kt
@@ -57,11 +57,11 @@ class ReferralController(
       return ResponseEntity.badRequest().body("malformed id")
     }
 
-    return referralService.updateDraftReferral(uuid, partialUpdate)
-      ?.let {
-        ResponseEntity.ok(DraftReferral(it))
-      }
-      ?: ResponseEntity.notFound().build()
+    val referralToUpdate = referralService.getDraftReferral(uuid)
+      ?: return ResponseEntity.notFound().build()
+
+    val updatedReferral = referralService.updateDraftReferral(referralToUpdate, partialUpdate)
+    return ResponseEntity.ok(DraftReferral(updatedReferral))
   }
 
   @GetMapping("/draft-referrals")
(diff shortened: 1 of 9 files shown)


src/main/resources/db/migration/V1_4__more_referral_fields.sql
```diff
@@ -0,0 +1,3 @@
+alter table referral
+    add column service_categoryid uuid,
+    add column complexity_levelid uuid;
```
mxco86, line 3: Should these two columns have the 'id' portion separated by an underscore (e.g. service_category_id) to be consistent with the complexity_level table?

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/controller/ReferralController.kt
```diff
@@ -63,4 +68,17 @@ class ReferralController(private val referralService: ReferralService) {
   fun getDraftReferralsCreatedByUserID(@RequestParam userID: String): List<DraftReferral> {
     return referralService.getDraftReferralsCreatedByUserID(userID)
   }
+
+  @GetMapping("/service-category/{id}")
+  fun getServiceCategoryByID(@PathVariable id: String): ResponseEntity<Any> {
+    val uuid = try {
+      UUID.fromString(id)
+    } catch (e: IllegalArgumentException) {
+      return ResponseEntity.badRequest().body("malformed id")
```
mxco86, line 77: Is it worth adding some more detail to the error message here?

mxco86 (commented):
Couple of minor queries

src/main/resources/db/migration/V1_4__more_referral_fields.sql
tomsmyers: good question - these are quirks of the JPA default naming which means we don't have to specify the column name for each entity property

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/controller/ReferralController.kt
tomsmyers: yes - but this is handled as part of another ticket.

src/main/resources/db/migration/V1_4__more_referral_fields.sql
mxco86: OK!

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/controller/ReferralController.kt
mxco86: 👍

mxco86 (approved):
LGTM

karenmillermoj (approved):
This looks good - no problems.

PR #49, opened 2021-01-05
allow updating service category id in patch referral endpoint

tomsmyers:

allows the service category id to be updated on the draft referral.

allow API consumers to set the service category ID to get the dev environment working


karenmillermoj (approved):
looks great

PR #50, opened 2021-01-05
IC-664 add backing fields for more referral form data

tomsmyers:
## What do these changes do?

adds the following fields to the referral entity and draft referral DTO:

```
  var additionalRiskInformation: String? = null,
  var furtherInformation: String? = null,
  var additionalNeedsInformation: String? = null,
  var accessibilityNeeds: String? = null,
  var needsInterpreter: Boolean? = null,
  var interpreterLanguage: String? = null,
  var hasAdditionalResponsibilities: Boolean? = null,
```

these changes are tested by the the contract tests, which are not yet running up-to-date. it's a work in progress.

update the backend to reflect recent front end changes.

### DIFF
--- a/src/main/resources/db/migration/V1_5__referral_further_information.sql
+++ b/src/main/resources/db/migration/V1_5__referral_further_information.sql
@@ -1,0 +1,9 @@
+alter table referral
+    add column further_information text,
+    add column accessibility_needs text,
+    add column additional_needs_information text,
+    add column needs_interpreter boolean,
+    add column interpreter_language text,
+    add column has_additional_responsibilities boolean,
+    add column when_unavailable text,
+    add column additional_risk_information text;
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/dto/DraftReferralDTO.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/dto/DraftReferralDTO.kt
@@ -11,6 +11,14 @@ data class DraftReferralDTO(
   val completionDeadline: LocalDate? = null,
   val serviceCategoryId: UUID? = null,
   val complexityLevelId: UUID? = null,
+  val furtherInformation: String? = null,
+  val additionalNeedsInformation: String? = null,
+  val accessibilityNeeds: String? = null,
+  val needsInterpreter: Boolean? = null,
+  val interpreterLanguage: String? = null,
+  val hasAdditionalResponsibilities: Boolean? = null,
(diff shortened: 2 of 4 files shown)


karenmillermoj (approved):
Great

src/main/resources/db/migration/V1_5__referral_further_information.sql
```diff
@@ -0,0 +1,9 @@
+alter table referral
+    add column further_information text,
+    add column accessibility_needs text,
+    add column additional_needs_information text,
+    add column needs_interpreter boolean,
+    add column interpreter_language text,
```
sldblog, line 6: I seem to recall that there was a recommendation to represent the `needs_interpreter` and `interpreter_language` as a single field, something similar to `required_language`.

For languages, could we use the two-letter [ISO 639-1](https://en.wikipedia.org/wiki/ISO_639-1) codes? (Assuming at this point we don't care about distinctions between English dialects)

src/main/resources/db/migration/V1_5__referral_further_information.sql
```diff
@@ -0,0 +1,9 @@
+alter table referral
+    add column further_information text,
```
sldblog, line 2: _data question_
I'm still catching up so I'm a bit out of context, but that may be good. Is it worth adding context to these field names?

For example, by analysing the data in `further_information`, the processor will have to know if it's in the context of risk, the person or the desired actions.

sldblog (commented):
Sorry for the post-merge questions -- related to interpreting the raw data

src/main/resources/db/migration/V1_5__referral_further_information.sql
lawrence-forooghian: Bringing conversation here from Slack. We need to know:

- whether the "is an interpreter required? question has already been answered  (to drive the UI)
- whether an interpreter is required
- if and only if an interpreter is required, then what language

I don’t see how we can represent all three with a single field.

As for language codes - currently it's just a free-text input in the design; since we can't ask the probation practitioner to provide a language code, we'd have to provide some other UX e.g. a dropdown for this.

src/main/resources/db/migration/V1_5__referral_further_information.sql
lawrence-forooghian: The question and hint are:

> Do you have further information for the ${this.serviceCategory.name} service provider? (optional)
> For example, relevant previous offences, previously completed programmes or further reasons for this referral

`further_information_for_service_provider`?

PR #51, opened 2021-01-06
Add Service Category Outcomes

karenmillermoj:

Adds Outcomes to the service category and outcome data against the existing categories.

Provide the Service Category outcomes set against a category.


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/jpa/entity/ServiceCategory.kt
```diff
@@ -14,4 +17,9 @@ data class ServiceCategory(
   @CreationTimestamp val created: OffsetDateTime,
   val name: String,
   @OneToMany @JoinColumn(name = "service_category_id") val complexityLevels: List<ComplexityLevel>,
+
+  @ElementCollection()
+  @CollectionTable(name = "service_category_outcome", joinColumns = [JoinColumn(name = "service_category_id")])
+  @Column(name = "outcome")
+  var outcomes: List<String>
```
tomsmyers, line 24: this should be a defined entity i.e. `List<Outcome>`. this follows the pattern used above for complexity levels. it also allows us to easily access the outcome id in this object, instead of just the outcome description. this is required in the DTO to satisfy the front end pacts for service category too.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/jpa/entity/Referral.kt
```diff
@@ -24,6 +26,10 @@ data class Referral(
   var whenUnavailable: String? = null,
   var complexityLevelID: UUID? = null,
   var serviceCategoryID: UUID? = null,
+
+  @OneToMany @JoinColumn(name = "service_category_id")
```
tomsmyers, line 30: i don't think this is required since the desired outccomes are set on the referral object explicitly via a call to PATCH /draft-referral

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/dto/DraftReferralDTO.kt
```diff
@@ -19,6 +20,7 @@ data class DraftReferralDTO(
   val hasAdditionalResponsibilities: Boolean? = null,
   val whenUnavailable: String? = null,
   val additionalRiskInformation: String? = null,
+  val desiredOutcomes: List<DesiredOutcome>? = null
```
tomsmyers, line 23: the type of this field should be a `List<UUID>` and the field name `desiredOutcomesIds` as per the contract test spec.

tomsmyers:
this needs a few things updating:

- the service category DTO needs to include `List<DesiredOutcome>` as `desiredOutcomes`
- the draft referral DTO needs to include `List<UUID>` as `desiredOutcomesIds`
- the pact tests need updating

src/main/resources/db/migration/V1_6_0__desired_outcome.sql
```diff
@@ -0,0 +1,16 @@
...
+
+  primary key (id),
+  constraint fk__desired_outcome__service_category foreign key(service_category_id) references service_category
+);
+
+create table referral_desired_outcomeids(
+    referral_id        uuid not null,
+    desired_outcomeids uuid,
+
+    constraint fk__referral_desired_outcomeids__referral foreign key (referral_id) references referral
```
sldblog, line 14: Given this table will be used to look up outcomes by referral ID, does this need an index on the `referral_id`?

sldblog (approved):
:+1: LGTM from a data perspective!

One non-blocker: efficient lookup outcome lookup would need an index on the linking table

src/main/resources/db/migration/V1_6_0__desired_outcome.sql
karenmillermoj: yes, it was created from the code hibernate generates.  It also needs to be not null.  I will update and push back - also renamed to remove the ids part of the name that was also automatically created

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/jpa/entity/ServiceCategory.kt
karenmillermoj: Updated

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/dto/DraftReferralDTO.kt
karenmillermoj: Updated

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/jpa/entity/Referral.kt
karenmillermoj: Updated

src/main/resources/db/migration/V1_6_0__desired_outcome.sql
karenmillermoj: Will add in the index separately

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/jpa/service/ReferralService.kt
```diff
@@ -124,6 +124,10 @@ class ReferralService(val repository: ReferralRepository) {
       referral.maximumRarDays = it
     }
 
+    if (!update.desiredOutcomeIds.isNullOrEmpty()) {
```
tomsmyers, line 127: this looks like it might make it impossible to explicitly set no desired outcomes. is that the correct behaviour?

tomsmyers:
since your rar days change has been merged first, the migration in this PR needs renaming to ensure correct application order

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/jpa/service/ReferralService.kt
karenmillermoj: Changed to ignore if null

tomsmyers (approved):

PR #52, opened 2021-01-06
Add rarDays to the Referral

karenmillermoj:

Adds rar days to the Referral and draftReferralDTO

Update the backend to reflect front end changes

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/jpa/service/ReferralService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/jpa/service/ReferralService.kt
@@ -58,6 +58,12 @@ class ReferralService(val repository: ReferralRepository) {
       }
     }
 
+    update.usingRarDays?.let {
+      if (it && update.maximumRarDays == null) {
+        errors.add(FieldError(field = "usingRarDays", error = Code.CONDITIONAL_FIELD_MUST_BE_SET))
+      }
+    }
+
     if (errors.isNotEmpty()) {
       throw ValidationError("draft referral update invalid", errors)
     }
@@ -108,6 +114,14 @@ class ReferralService(val repository: ReferralRepository) {
 
     update.additionalRiskInformation?.let {
       referral.additionalRiskInformation = it
+    }
+
+    update.usingRarDays?.let {
+      referral.usingRarDays = it
+    }
+
+    update.maximumRarDays?.let {
+      referral.maximumRarDays = it
     }
 
     return repository.save(referral)
--- a/src/main/resources/db/migration/V1_7__referral_rar_days.sql
+++ b/src/main/resources/db/migration/V1_7__referral_rar_days.sql
@@ -1,0 +1,3 @@
+alter table referral
+    add column using_rar_days boolean,
+    add column maximum_rar_days int;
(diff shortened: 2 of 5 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/jpa/service/ReferralService.kt
```diff
@@ -47,7 +47,7 @@ class ReferralService(val repository: ReferralRepository) {
     }
 
     update.needsInterpreter?.let {
-      // fixme: error if this is true and interpreterLangyage is missing
+      // fixme: error if this is true and interpreterLanguage is missing
```
tomsmyers, line 50: thanks !

src/main/resources/db/migration/V1_7__referral_rar_days.sql
```diff
@@ -0,0 +1,3 @@
+alter table referral
+    add column using_rar_days boolean,
+    add column maximum_rar_days int
```
tomsmyers, line 3: missing `;` at eol

karenmillermoj: good spot - cheers

tomsmyers (approved):
nice - thanks.

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/jpa/service/ReferralServiceTest.kt
```diff
@@ -220,6 +220,23 @@ class ReferralServiceTest @Autowired constructor(
     referralService.updateDraftReferral(referral, DraftReferralDTO(hasAdditionalResponsibilities = true, whenUnavailable = "wednesdays"))
   }
 
+  @Test
+  fun `when usingRarDays is true, maximumRarDays must be set`() {
+    val referral = Referral()
+    // this is fine
```
sldblog, line 226: Is there a negative variant for assertThrows? It may replace the comment if we could say `assertDoesNotThrow<ValidationError>` or similar 🤔 

(I'm a fan of self-explaining assertions as I always forget to update the comments 😅)

sldblog (approved):

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/jpa/service/ReferralService.kt
```diff
@@ -58,6 +58,12 @@ class ReferralService(val repository: ReferralRepository) {
       }
     }
 
+    update.usingRarDays?.let {
+      if (it && update.maximumRarDays == null) {
+        errors.add(FieldError(field = "usingRarDays", error = Code.CONDITIONAL_FIELD_MUST_BE_SET))
```
sldblog, line 63: We could use a custom code here, for example `rar_days_must_be_set`. Do we need to use generic ones?

sldblog (approved):
Extra question, not a blocker

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/jpa/service/ReferralService.kt
tomsmyers: this does mean there any many more error codes for the front end to manage though. isn't the field info with this message sufficient to convey what the issue is?

PR #53, opened 2021-01-06
IC-615 - more sensible error responses for validation errors 

tomsmyers:

there are three commits here 

- the first reverts a commit which pulls in a third party error handling library.
- the second changes the way we deal with errors so we have more control over the responses. the `ErrorResponse` object defines what the errors always look like. it's more stream lined than before, and allows the api consumer to simply check for existence of `validationErrors` to see if there are field level errors.
- the third adds validation logic for a bunch of existing fields, and the tests for that new logic. 

allow the front end to process form validation errors from the API


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/config/ErrorConfiguration.kt
```diff
@@ -1,60 +1,74 @@
...
-    val bean = LocalValidatorFactoryBean()
-    bean.setValidationMessageSource(messageSource())
-    return bean
+  companion object {
+    private val log = LoggerFactory.getLogger(ErrorConfiguration::class.java)
   }
 }
+
+@JsonInclude(JsonInclude.Include.NON_NULL)
+data class ErrorResponse(
```
tomsmyers, line 69: this is the key class which describes our error responses

mxco86: Have you considered using the HTTP problem details format for the error responses? The format is almost the same so it might be worth a look. Here is the RFC: https://tools.ietf.org/html/rfc7807

tomsmyers: thanks for sharing this document. as you mentioned, the format is similar so i think what we have is ok for now.

karenmillermoj (approved):
I like the enum for the error codes.

PR #55, opened 2021-01-07
store incoming user_id for newly created referrals

tomsmyers:

stored the user id of the logged in user who created the draft referral

allow us to show users all their in progress referrals


karenmillermoj (commented):
Should the created_by_userid  column on the referral table now be set to not null ?

tomsmyers:
> Should the created_by_userid column on the referral table now be set to not null ?

hmm what happens to existing referral entities which do not have created_by_userid set?

karenmillermoj (approved):
no problems

src/main/resources/db/migration/V1_6__created_by_update.sql
```diff
@@ -0,0 +1,11 @@
+alter table referral
+    add column created_by_user_auth_source varchar(255) not null;
+
+-- all the referrals that currently exist are in dev, so updating to a known dev user_id
+update referral
+    set created_by_userid = '2500128586',
+        created_by_user_auth_source = 'delius'
+    where created_by_userid = null;
```
sldblog, line 8: At this point, could we do this change manually on our current environments and/or put it in the `local` scope?

These scripts have to be treated as production code and reasoning like "this won't get executed on production because ..." can create confusion if we look back in a couple of months.

sldblog (commented):
:+1: Looks good to me. I'm a bit uneasy about putting a dev-only data migration in, raised a question about that.

src/main/resources/db/migration/V1_6__created_by_update.sql
tomsmyers: agree, i hesitated a lot over this. karen is right that having the `not null` modifier on the fields is correct, and especially important for prod. however we can't run this migration if any of the existing `created_by_userid`s are null. the reasoning here isn't "this won't get executed on production", it's "this will get executed on production in the furture, but from this point forward there will be no entries with `created_by_userid = null` so the update will not apply to any production data".

sldblog (approved):

src/main/resources/db/migration/V1_6__created_by_update.sql
sldblog: Yes, sorry, "there will be no entries" is what I meant by "won't get executed". We're on the same page :)

Thanks for updating the entry values to `unknown`; that'll make reasoning about values in environments simpler 🎉

PR #57, opened 2021-01-11
rename 'created' field in referral entity to 'createdAt' to make it's meaning clearer to the API consumer

tomsmyers:

rename 'created' field in referral entity to 'createdAt' to make it's meaning clearer to the API consumer

there was some confusion on the backend about the meaning of `created` in the DraftReferralDTO. it sounds like a boolean value, but really it's the timestamp when the referral was created.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/dto/DraftReferralDTO.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/dto/DraftReferralDTO.kt
@@ -7,7 +7,7 @@ import java.util.UUID
 
 data class DraftReferralDTO(
   val id: UUID? = null,
-  val created: OffsetDateTime? = null,
+  val createdAt: OffsetDateTime? = null,
   val createdByUserId: String? = null,
   val completionDeadline: LocalDate? = null,
   val serviceCategoryId: UUID? = null,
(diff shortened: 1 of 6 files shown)


Gweaton (approved):
Renaming looks sensible, but I haven't exactly grepped the codebase for any you might have missed, hope that's enough for a review!

src/main/resources/db/migration/V1_10__created_at_update.sql
```diff
@@ -0,0 +1,2 @@
+alter table referral
```
sldblog, line 1: Related to the previous versioning PRs -- would it make sense to switch to timestamp-based versions rather than semantic versioning?

`v20210111.210044__rename_created_at.sql`

sldblog (approved):
👍 Nice!

No blockers.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/dto/DraftReferralDTO.kt
```diff
@@ -7,7 +7,7 @@ import java.util.UUID
 
 data class DraftReferralDTO(
   val id: UUID? = null,
-  val created: OffsetDateTime? = null,
+  val createdAt: OffsetDateTime? = null,
```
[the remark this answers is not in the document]
tomsmyers: in the entity yes, but not in the DTO, since this is used to deserialize incoming PATCH referral requests.

src/main/resources/db/migration/V1_10__created_at_update.sql
tomsmyers: not seen this before in MOJ but i'm into it

PR #60, opened 2021-01-11
split not null migration into seperate transaction

tomsmyers:

split not null migration into seperate transaction. previously i split the migration into seperate files, but the naming convention meant the migration was still applied as one. this should finally fix this issue.

unblock the deploy


Gweaton (approved):
Let's see if this fixes the issue

sldblog (changes requested):
Locally, this still fails for me 🤔

sldblog (commented):
Pushed an `IS NULL` change up -- passes locally now

tomsmyers:
> Pushed an `IS NULL` change up -- migration works locally now
> 
> Totally missed this a few times in the past reviews 😅

thank you !

sldblog (approved):

sldblog:
naughty self-approval there 🙈

PR #61, opened 2021-01-12
Re-enable (non-blocking) Pact tests

sldblog:

Re-enables Pact tests that

- do not block deployments and development
- runnable locally (use `./gradlew pactTestPublish`)
- can be tested against arbitrary tags (use `PACTBROKER_CONSUMERVERSIONSELECTORS_TAGS ` env var)
- can be viewed in CircleCI (e.g. [here](https://1201-312544431-gh.circle-artifacts.com/0/build/reports/tests/pactTestPublish/classes/uk.gov.justice.digital.hmpps.hmppsinterventionsservice.integration.PactTest.html))

This **will** fail the `ci/circleci: pact_check_and_publish` job on CI but that will no longer block the deployment. This is **normal** for now.

## Where should the reviewer start?

Please review commit-by-commit, there's context there for each step.

To make the tests run against `main` on CI so we're aware how far away we are from mainline contracts.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -92,7 +92,7 @@ workflows:
   build_test_and_deploy:
     jobs:
       - validate
-#      - pact_check_and_publish
+      - pact_check_and_publish
       - hmpps/helm_lint:
           name: helm_lint
       - hmpps/build_docker:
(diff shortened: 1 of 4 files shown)


tomsmyers (approved):
lgtm

sldblog:
🙇 🎩 Thank you!

PR #62, opened 2021-01-12
some commits which bring the api more in line with the front end contracts

tomsmyers:

several small commits:
- add required pact states so we get proper test failures and not just "state missing" errors
- behaviour change to unset conditional fields when their boolean radio button is set to false

working towards getting latest contract tests green


sldblog:
Could we enable #61 before these changes?

karenmillermoj (approved):
Looks good

PR #63, opened 2021-01-13
Add service user info to draft referral and allow it to be set

tomsmyers:

Add service user info to draft referral and allow it to be set

allow the data pulled from delius about the SU to be stored with the draft referral

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/dto/ServiceUserDTO.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/dto/ServiceUserDTO.kt
@@ -1,11 +1,9 @@
 package uk.gov.justice.digital.hmpps.hmppsinterventionsservice.dto
 
-import com.fasterxml.jackson.annotation.JsonInclude
 import uk.gov.justice.digital.hmpps.hmppsinterventionsservice.jpa.entity.ServiceUser
 import uk.gov.justice.digital.hmpps.hmppsinterventionsservice.jpa.entity.Sex
 import java.util.Date
 
-@JsonInclude(JsonInclude.Include.NON_NULL)
 data class ServiceUserDTO(
   var pncNumber: String? = null,
   var crn: String? = null,
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/dto/ServiceUserDTO.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/dto/ServiceUserDTO.kt
@@ -1,25 +1,18 @@
 package uk.gov.justice.digital.hmpps.hmppsinterventionsservice.dto
 
 import uk.gov.justice.digital.hmpps.hmppsinterventionsservice.jpa.entity.ServiceUserData
-import uk.gov.justice.digital.hmpps.hmppsinterventionsservice.jpa.entity.Sex
 import java.util.Date
 
 data class ServiceUserDTO(
-  var pncNumber: String? = null,
   var crn: String? = null,
-  var nomisNumber: String? = null,
-  var dob: Date? = null,
-  var sex: Sex? = null,
-  var ethnicity: String? = null,
   var title: String? = null,
   var firstName: String? = null,
   var lastName: String? = null,
-  var otherNames: String? = null,
-  var address: String? = null,
-  var needs: String? = null,
+  var dob: Date? = null,
+  var gender: String? = null,
+  var ethnicity: String? = null,
   var preferredLanguage: String? = null,
   var religionOrBelief: String? = null,
-  var sexualOrientation: String? = null,
   var disabilities: String? = null,
 ) {
   companion object {
@@ -27,20 +20,14 @@ data class ServiceUserDTO(
       val dto = ServiceUserDTO(crn = crn)
       serviceUserData?.let {
         dto.crn = crn
-        dto.pncNumber = serviceUserData.pncNumber
-        dto.nomisNumber = serviceUserData.nomisNumber
-        dto.dob = serviceUserData.dob
-        dto.sex = serviceUserData.sex
-        dto.ethnicity = serviceUserData.ethnicity
         dto.title = serviceUserData.title
         dto.firstName = serviceUserData.firstName
         dto.lastName = serviceUserData.lastName
-        dto.otherNames = serviceUserData.otherNames
-        dto.address = serviceUserData.address
-        dto.needs = serviceUserData.needs
+        dto.dob = serviceUserData.dob
+        dto.gender = serviceUserData.gender
+        dto.ethnicity = serviceUserData.ethnicity
         dto.preferredLanguage = serviceUserData.preferredLanguage
         dto.religionOrBelief = serviceUserData.religionOrBelief
-        dto.sexualOrientation = serviceUserData.sexualOrientation
         dto.disabilities = serviceUserData.disabilities
       }
       return dto
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/dto/ServiceUserDTO.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/dto/ServiceUserDTO.kt
@@ -1,0 +1,51 @@
+package uk.gov.justice.digital.hmpps.hmppsinterventionsservice.dto
+
+import com.fasterxml.jackson.annotation.JsonInclude
+import uk.gov.justice.digital.hmpps.hmppsinterventionsservice.jpa.entity.ServiceUser
+import uk.gov.justice.digital.hmpps.hmppsinterventionsservice.jpa.entity.Sex
+import java.util.Date
+
+@JsonInclude(JsonInclude.Include.NON_NULL)
+data class ServiceUserDTO(
+  var pncNumber: String? = null,
+  var crn: String? = null,
+  var nomisNumber: String? = null,
+  var dob: Date? = null,
+  var sex: Sex? = null,
+  var ethnicity: String? = null,
+  var title: String? = null,
+  var firstName: String? = null,
+  var lastName: String? = null,
+  var otherNames: String? = null,
+  var address: String? = null,
+  var needs: String? = null,
+  var preferredLanguage: String? = null,
+  var religionOrBelief: String? = null,
+  var sexualOrientation: String? = null,
+  var disabilities: String? = null,
+) {
+  companion object {
+    fun from(serviceUser: ServiceUser?): ServiceUserDTO? {
+      return serviceUser?.let {
+        ServiceUserDTO(
+          pncNumber = serviceUser.pncNumber,
+          crn = serviceUser.crn,
+          nomisNumber = serviceUser.nomisNumber,
+          dob = serviceUser.dob,
+          sex = serviceUser.sex,
+          ethnicity = serviceUser.ethnicity,
+          title = serviceUser.title,
+          firstName = serviceUser.firstName,
+          lastName = serviceUser.lastName,
+          otherNames = serviceUser.otherNames,
+          address = serviceUser.address,
+          needs = serviceUser.needs,
+          preferredLanguage = serviceUser.preferredLanguage,
+          religionOrBelief = serviceUser.religionOrBelief,
+          sexualOrientation = serviceUser.sexualOrientation,
+          disabilities = serviceUser.disabilities,
+        )
+      }
+    }
+  }
+}
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/dto/ServiceUserDTO.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/dto/ServiceUserDTO.kt
@@ -1,6 +1,6 @@
 package uk.gov.justice.digital.hmpps.hmppsinterventionsservice.dto
 
-import uk.gov.justice.digital.hmpps.hmppsinterventionsservice.jpa.entity.ServiceUser
+import uk.gov.justice.digital.hmpps.hmppsinterventionsservice.jpa.entity.ServiceUserData
 import uk.gov.justice.digital.hmpps.hmppsinterventionsservice.jpa.entity.Sex
 import java.util.Date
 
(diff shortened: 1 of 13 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/jpa/service/ReferralService.kt
```diff
@@ -26,7 +27,7 @@ class ReferralService(val repository: ReferralRepository) {
 
     update.serviceCategoryId?.let {
       if (referral.serviceCategoryID != null) {
-        errors.add(FieldError(field = "serviceCategoryId", error = Code.SERVICE_CATEGORY_CANNOT_BE_CHANGED))
+        errors.add(FieldError(field = "serviceCategoryId", error = Code.FIELD_CANNOT_BE_CHANGED))
```
sldblog, line 30: This sounds like a business rule. Can we add -- for the API client and posterity's benefit -- the reason for having this rule, please? Perhaps as a message/reason?

Could this be `referral.serviceCategoryID != null  && referral.serviceCategoryID != update.serviceCategoryId` to allow idempotent updates from the client side?

src/main/resources/db/local/V99_0__draft_referrals.sql
```diff
@@ -3,3 +3,6 @@ values ('ac386c25-52c8-41fa-9213-fcf42e24b0b5', '2020-12-07 18:02:01.599803+00',
        ('dfb64747-f658-40e0-a827-87b4b0bdcfed', '2020-12-07 20:45:21.986389+00', '2021-03-01', '8751622134', 'delius', null),
        ('d496e4a7-7cc1-44ea-ba67-c295084f1962', '2020-12-24 09:32:32.871623+00', '2021-01-30', '2500128586', 'delius', '428ee70f-3001-4399-95a6-ad25eaaede16'),
        ('1219a064-709b-4b6c-a11e-10b8cb3966f6', '2021-01-12 14:46:21.987234+00', null, '2500128586', 'delius', '428ee70f-3001-4399-95a6-ad25eaaede16');
+
+insert into service_user (referral_id, address, crn, disabilities, dob, ethnicity, title, first_name, last_name, needs, nomis_number, other_names, pnc_number, preferred_language, religion_or_belief, sex, sexual_orientation)
```
sldblog, line 7: If we intend this to be 1-to-1 can we choose another name, please? Based on the name, I'd assume it's a primary table. What do you think about `referral_service_user_snapshot`?

src/main/resources/db/migration/V1_12__service_user.sql
```diff
@@ -0,0 +1,23 @@
+create table service_user (
+    referral_id uuid not null unique,
+    address text,
+    crn text,
```
sldblog, line 4: Can we treat CRN as a required (and lookup) key for users, please? We shouldn't have any without a CRN identifier (there are processes in court admin which guarantee this as far as I know)

sldblog (commented):
Bunch of random questions -- might be easier to have a quick call/pair review?

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/jpa/service/ReferralService.kt
tomsmyers: the reason for this is that things like outcomes and complexity level are based on the service category. so changing the service category could lead to inconsistent data in the referral, e.g. complexity levels for the wrong service category.

on the second part - yes, we can do that.

src/main/resources/db/local/V99_0__draft_referrals.sql
tomsmyers: sure thing

src/main/resources/db/migration/V1_12__service_user.sql
tomsmyers: yes

are there any such rules on nomis or PNC numbers? or are they sort of "bonus" identifiers?

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/jpa/entity/Referral.kt
```diff
@@ -4,20 +4,28 @@ import org.hibernate.annotations.CreationTimestamp
...
 import javax.validation.constraints.NotNull
 
 @Entity
 @Table(indexes = arrayOf(Index(columnList = "created_by_userid")))
 data class Referral(
+  // this seems like it should be a OneToMany relationship, however it is OneToOne
+  // for good reason. all service user information we pull from nDelius can change,
+  // and often does. storing a snapshot of the current information for each
+  // referral avoids any issues which may arise due to out of date information.
+  @OneToOne(mappedBy = "referral", cascade = arrayOf(CascadeType.ALL)) @PrimaryKeyJoinColumn var serviceUser: ServiceUser? = null,
```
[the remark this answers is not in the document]
tomsmyers: yeh this is a good idea. i played around with multiple versions of this type of thing (hence my questions in slack yesterday). but i like the idea of seperating the CRN as _the_ SU identifier

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/jpa/service/ReferralService.kt
```diff
@@ -71,8 +72,15 @@ class ReferralService(val repository: ReferralRepository) {
       if (referral.serviceCategoryID == null && update.serviceCategoryId == null) {
         errors.add(FieldError(field = "desiredOutcomeIds", error = Code.SERVICE_CATEGORY_MUST_BE_SET))
       }
+
+      // fixme: error if desiredOutcomeIds not valid for service category
+    }
+
+    update.serviceUser?.let {
+      if (referral.serviceUser != null) {
+        errors.add(FieldError(field = "serviceUser", error = Code.FIELD_CANNOT_BE_CHANGED))
```
tomsmyers, line 81: @sldblog your thing on CRN above also makes this check significantly better. 

the CRN cannot be change - the service user details can

src/main/resources/db/local/V99_0__draft_referrals.sql
tomsmyers: change to `referral_service_user_data`

src/main/resources/db/migration/V1_12__service_user.sql
tomsmyers: i've just taken CRN out of this and used it as the primary service user ID in the referral table.

tomsmyers: or do you want to be able to query all referral_service_user_data entities with a certain crn?

tomsmyers: perhaps if we have a requirement for this in the future we can add it then. we have all the data.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/dto/ServiceUserDTO.kt
```diff
@@ -0,0 +1,49 @@
...
+  var title: String? = null,
+  var firstName: String? = null,
+  var lastName: String? = null,
+  var otherNames: String? = null,
+  var address: String? = null,
+  var needs: String? = null,
+  var preferredLanguage: String? = null,
+  var religionOrBelief: String? = null,
+  var sexualOrientation: String? = null,
+  var disabilities: String? = null,
```
sldblog, line 23: I'm worried about creating this detailed API spec now. I think for the skeleton we may go with CRN, name and responsible LDU.

From a data handling perspective, we shouldn't copy more data than we need for processing the referrals.

Could we skip fields until we have a need for them through the UI or some business logic?
Is there added complexity if we want to add fields later?

src/main/resources/db/migration/V1_12__service_user_data.sql
```diff
@@ -1,16 +1,18 @@
-create table service_user (
+alter table referral
+    add column service_usercrn text not null default 'unknown';
+
+create table referral_service_user_data (
     referral_id uuid not null unique,
+    nomis_number text,
+    pnc_number text,
     address text,
-    crn text,
```
karenmillermoj: Does Address need to be in a separate table?

sldblog (commented):
Question about currently unused fields/skeleton scope/future changes

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/dto/ServiceUserDTO.kt
```diff
@@ -0,0 +1,49 @@
...
+import uk.gov.justice.digital.hmpps.hmppsinterventionsservice.jpa.entity.ServiceUserData
+import uk.gov.justice.digital.hmpps.hmppsinterventionsservice.jpa.entity.Sex
+import java.util.Date
+
+data class ServiceUserDTO(
+  var pncNumber: String? = null,
+  var crn: String? = null,
+  var nomisNumber: String? = null,
+  var dob: Date? = null,
+  var sex: Sex? = null,
```
sldblog, line 12: I saw some discussions about the difference between sex and gender -- which one do we need to capture?

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/jpa/service/ReferralService.kt
sldblog: Thank you, that's really useful context.

Would similar messaging make sense?
```kt
errors.add(FieldError(field = "serviceCategoryId", error = Code.FIELD_CANNOT_BE_CHANGED,
  reason = "service category can only be selected at the start of a new referral"))
```

The intent of this would be to be explicit to code readers and API clients about _why_ the field is immutable. (Could be done via data structures/predicate functions too)

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/dto/ServiceUserDTO.kt
tomsmyers: see comment on the enum. we use this to be able to determine if the SU needs a male or female specific service.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/dto/ServiceUserDTO.kt
tomsmyers: i just think we're making more work for ourselves down the line. i'm basing this on what's in the prototype, and all these fields are included. if something has changed we can remove the ones that aren't needed.

src/main/resources/db/migration/V1_12__service_user_data.sql
tomsmyers: i don't think so

karenmillermoj (approved):
great

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/jpa/entity/ServiceUserData.kt
```diff
@@ -0,0 +1,28 @@
...
+  var lastName: String? = null,
+  @Column(name = "dob", columnDefinition = "DATE") var dob: Date? = null,
+  var gender: String? = null,
+  var ethnicity: String? = null,
+  var preferredLanguage: String? = null,
+  var religionOrBelief: String? = null,
+  var disabilities: String? = null,
+
+  @OneToOne @MapsId @JoinColumn(name = "referral_id") var referral: Referral? = null,
+  @Id @Column(name = "referral_id") var id: UUID? = null,
```
sldblog, line 27: Could this be named `referralId` instead of `id`? Or can we skip it?
I don't trust myself to remember that `id` is a foreign key :)

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/jpa/entity/Referral.kt
```diff
@@ -40,6 +45,8 @@ data class Referral(
   @Column(name = "desired_outcome_id")
   var desiredOutcomeIDs: List<UUID>? = null,
   var completionDeadline: LocalDate? = null,
+
+  @NotNull val serviceUserCRN: String,
```
sldblog, line 49: Given `serviceUserCRN` is a required field, what do you think about moving it to the top?

sldblog (commented):
LGTM overall

One question before approval about `ServiceUserData.id`

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/jpa/entity/ServiceUserData.kt
tomsmyers: sure

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/jpa/entity/Referral.kt
tomsmyers: these three fields at the bottom are all sort of 'required', but postgres generates the other two. that's the reason for the grouping. can move to the top if you like though.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/jpa/entity/ServiceUserData.kt
```diff
@@ -0,0 +1,28 @@
...
+  var lastName: String? = null,
+  @Column(name = "dob", columnDefinition = "DATE") var dob: Date? = null,
+  var gender: String? = null,
+  var ethnicity: String? = null,
+  var preferredLanguage: String? = null,
+  var religionOrBelief: String? = null,
+  var disabilities: String? = null,
+
+  @OneToOne @MapsId @JoinColumn(name = "referral_id") var referral: Referral? = null,
+  @Id @Column(name = "referral_id") var referralID: UUID? = null,
```
sldblog, line 27: _not a blocker_
Sorry, should've commented this earlier. Can we use `@Id` here? From its documentation:

> The mapped column for the primary key of the entity is assumed to be the primary key of the primary table

Is this a matching use case?

sldblog (approved):
👍 LGTM, no blockers

PR #64, opened 2021-01-13
basic changes to allow converting draft referrals to sent referrals

tomsmyers:

- adds controller methods for sending a draft referral and retreiving a sent referral
- refactors 'createdByUserId' and 'createdByUserAuthSource' to a single 'createdBy' field which is a foreign key to an 'AuthUser' entity. this is also used to store 'sentBy'. 
- service methods to retrieve only draft or sent referrals, as well as only draft referrals for a certain user id.
- 

allow draft referrals to be sent


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/jpa/entity/Referral.kt
```diff
@@ -30,8 +38,7 @@ data class Referral(
   var serviceCategoryID: UUID? = null,
   var usingRarDays: Boolean? = null,
   var maximumRarDays: Int? = null,
-  @NotNull var createdByUserAuthSource: String? = null,
-  @Column(name = "created_by_userid") @NotNull var createdByUserID: String? = null,
+  @ManyToOne(fetch = FetchType.LAZY, cascade = [CascadeType.ALL]) var createdBy: AuthUser? = null,
```
tomsmyers, line 41: needs NotNull

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/dto/SentReferralDTO.kt
```diff
@@ -0,0 +1,23 @@
+package uk.gov.justice.digital.hmpps.hmppsinterventionsservice.dto
+
+import uk.gov.justice.digital.hmpps.hmppsinterventionsservice.jpa.entity.Referral
+import java.time.OffsetDateTime
+import java.util.UUID
+
+class SentReferralDTO(
+  val id: UUID,
+  val sentAt: OffsetDateTime,
+  val referenceNumber: String?,
```
tomsmyers, line 10: this is probably required

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/jpa/entity/AuthUser.kt
```diff
@@ -0,0 +1,11 @@
+package uk.gov.justice.digital.hmpps.hmppsinterventionsservice.jpa.entity
+
+import io.micrometer.core.lang.NonNull
```
karenmillermoj, line 3: Should we be using the spring import org.springframework.lang.NonNull ?

tomsmyers: yes - good spot. this is an IDE mishap.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/jpa/entity/Referral.kt
```diff
@@ -4,20 +4,28 @@ import org.hibernate.annotations.CreationTimestamp
...
 import javax.persistence.Table
 import javax.persistence.UniqueConstraint
 import javax.validation.constraints.NotNull
 
 @Entity
-@Table(indexes = arrayOf(Index(columnList = "created_by_userid")))
+@Table(indexes = arrayOf(Index(columnList = "created_by_id")))
 data class Referral(
+  var sentAt: OffsetDateTime? = null,
+  @ManyToOne(fetch = FetchType.LAZY, cascade = [CascadeType.ALL]) var sentBy: AuthUser? = null,
```
tomsmyers, line 25: check that this does not delete the user when the referral gets deleted

src/main/resources/db/migration/V1_13_1__created_by_auth_user.sql
```diff
@@ -0,0 +1,24 @@
+-- add the new column type that will replace both 'created_by_userid' and 'created_by_user_auth_source'
+alter table referral
+    add column created_by_id text,
```
tomsmyers, line 3: not null

karenmillermoj (approved):
lgtm

src/main/resources/db/migration/V1_13_1__created_by_auth_user.sql
```diff
@@ -0,0 +1,27 @@
...
+set created_by_id = created_by_userid;
+
+alter table referral
+    alter column created_by_id set not null;
+
+-- create any auth user entities that exist in the referral table but not the auth_user table.
+insert into auth_user (id, auth_source)
+    select created_by_userid, created_by_user_auth_source
+    from referral r
+    where not exists (select from auth_user where id = r.created_by_userid);
```
sldblog, line 17: _not a blocker_
💭 With only local and dev environments, we have the option to avoid doing these data migrations. We could collapse the migrations into a single migration per table before we roll up the pre-prod environment.

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/dto/DraftReferralDTOTest.kt
```diff
@@ -43,16 +43,13 @@ class DraftReferralDTOTest(@Autowired private val json: JacksonTester<DraftRefer
...
-      createdByUserAuthSource = "delius",
     )
     val out = json.write(DraftReferralDTO.from(referral))
     assertThat(out).isEqualToJson(
       """
       {
         "id": "3b9ed289-8412-41a9-8291-45e33e60276c", 
-        "createdAt": "2020-12-04T10:42:43Z",
-        "createdByUserId": "123456"
+        "createdAt": "2020-12-04T10:42:43Z"
```
sldblog, line 52: _not a blocker_
Is it worth using microseconds (as used by PostgreSQL) precision here?

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/dto/SentReferralDTOTest.kt
```diff
@@ -0,0 +1,66 @@
...
+      sentAt = sentAt,
+    )
+
+    val out = json.write(SentReferralDTO.from(referral))
+    Assertions.assertThat(out).isEqualToJson(
+      """
+      {
+        "sentAt": "2021-01-13T21:57:13Z",
+        "referral": {
+          "id": "3b9ed289-8412-41a9-8291-45e33e60276c",
```
sldblog, line 57: As clients probably expect an ID key for each entity, should this `id` be on the top level too?

(Side note: this matcher may get a lot of churn as the data models progress. Is there one that checks for the given fields rather than full equality?)

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/jpa/service/ReferralService.kt
```diff
@@ -1,24 +1,38 @@
...
+    return repository.save(referral)
+  }
+
+  fun createDraftReferral(user: AuthUser): Referral {
+    return repository.save(Referral(createdBy = user))
   }
 
   fun getDraftReferral(id: UUID): Referral? {
-    return repository.findByIdOrNull(id)
+    return repository.findByIdAndSentAtIsNull(id)
```
sldblog, line 35: _wider question, later_
Returning `null` for already sent referrals is safe but potentially confusing for clients. Getting a 404 on the previously valid request can be confusing.

Is there a way (later) to respond with "303 See Other" or "409 Conflict"?

sldblog (approved):
👍 LGTM, couple of questions, no blockers

sldblog:
Validate fails with
```
/.../jpa/repository/AuthUserRepository.kt:6:29: Missing spacing before ":"
"plain" report written to /home/circleci/app/build/reports/ktlint/ktlintMainSourceSetCheck/ktlintMainSourceSetCheck.txt
```

The report, unfortunately, isn't yet persisted in CircleCI

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/dto/SentReferralDTOTest.kt
tomsmyers: there is an id at the top level too. it's missing from this test because it's testing that the draft referral fields specifically are included. 

this is matching the id created in this test specifically - so it will not change.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/jpa/service/ReferralService.kt
tomsmyers: hmm let's discuss outside this pr. it definitely shouldn't be a 200 code IMO, but happy to think about others

PR #66, opened 2021-01-15
IC-916 send email notifications for sent referrals

tomsmyers:

this PR enables email notifications to be triggered when referrals are sent.

we are using spring's async events framework which allows multiple listeners to subscribe to these events which should provide a nice way to hook in other actions that need to be triggered for things like this.

NOTE: half way through these changes i refactored some paths so that the 'services' package is part of the JPA package. this means 'notification service' makes sense (since it's nothing to do with JPA). please review commit by commit to make your life easier when looking at these changes.

alert the service provider that a new referral has been sent to them

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/config/NotifyConfiguration.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/config/NotifyConfiguration.kt
@@ -1,0 +1,14 @@
+package uk.gov.justice.digital.hmpps.hmppsinterventionsservice.config
+
+import org.springframework.beans.factory.annotation.Value
+import org.springframework.context.annotation.Bean
+import org.springframework.context.annotation.Configuration
+import uk.gov.service.notify.NotificationClient
+
+@Configuration
+class NotifyConfiguration(@Value("\${notify.apikey:}") private val apiKey: String) {
+  @Bean
+  fun notifyClient(): NotificationClient {
+    return NotificationClient(apiKey)
+  }
+}
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/events/NotifyListeners.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/events/NotifyListeners.kt
@@ -1,19 +1,19 @@
+// ktlint-disable filename
 package uk.gov.justice.digital.hmpps.hmppsinterventionsservice.events
 
 import org.springframework.beans.factory.annotation.Value
 import org.springframework.context.ApplicationListener
 import org.springframework.stereotype.Component
-import org.springframework.web.servlet.support.ServletUriComponentsBuilder
 import org.springframework.web.util.UriComponentsBuilder
-import uk.gov.justice.digital.hmpps.hmppsinterventionsservice.jpa.service.NotifyService
+import uk.gov.justice.digital.hmpps.hmppsinterventionsservice.service.NotifyService
 
 @Component
 class NotifyReferralEventListener(
   @Value("\${interventions.ui.uri}") private val interventionsUIURI: String,
   private val notifyService: NotifyService,
-): ApplicationListener<ReferralEvent> {
-  override fun onApplicationEvent(event: ReferralEvent){
-    when(event.type) {
+) : ApplicationListener<ReferralEvent> {
+  override fun onApplicationEvent(event: ReferralEvent) {
+    when (event.type) {
       ReferralEventType.SENT -> {
         // fixme: is it ok that this class implicitly knows the URL for the sent referral location?
         val location = UriComponentsBuilder.fromHttpUrl(interventionsUIURI)
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/events/NotifyListeners.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/events/NotifyListeners.kt
@@ -9,15 +9,15 @@ import uk.gov.justice.digital.hmpps.hmppsinterventionsservice.service.NotifyServ
 
 @Component
 class NotifyReferralEventListener(
-  @Value("\${interventions.ui.uri}") private val interventionsUIURI: String,
+  @Value("\${interventions-ui.baseurl}") private val interventionsUIBaseURL: String,
+  @Value("\${interventions-ui.locations.sent-referral}") private val interventionsUISentReferralLocation: String,
   private val notifyService: NotifyService,
 ) : ApplicationListener<ReferralEvent> {
   override fun onApplicationEvent(event: ReferralEvent) {
     when (event.type) {
       ReferralEventType.SENT -> {
-        // fixme: is it ok that this class implicitly knows the URL for the sent referral location?
-        val location = UriComponentsBuilder.fromHttpUrl(interventionsUIURI)
-          .path("referrals/{id}/confirmation")
+        val location = UriComponentsBuilder.fromHttpUrl(interventionsUIBaseURL)
+          .path(interventionsUISentReferralLocation)
           .buildAndExpand(event.referral.id)
           .toUri()
 
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/events/NotifyListeners.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/events/NotifyListeners.kt
@@ -1,0 +1,29 @@
+package uk.gov.justice.digital.hmpps.hmppsinterventionsservice.events
+
+import org.springframework.beans.factory.annotation.Value
+import org.springframework.context.ApplicationListener
+import org.springframework.stereotype.Component
+import org.springframework.web.servlet.support.ServletUriComponentsBuilder
+import org.springframework.web.util.UriComponentsBuilder
+import uk.gov.justice.digital.hmpps.hmppsinterventionsservice.jpa.service.NotifyService
+
+@Component
+class NotifyReferralEventListener(
+  @Value("\${interventions.ui.uri}") private val interventionsUIURI: String,
+  private val notifyService: NotifyService,
+): ApplicationListener<ReferralEvent> {
+  override fun onApplicationEvent(event: ReferralEvent){
+    when(event.type) {
+      ReferralEventType.SENT -> {
+        // fixme: is it ok that this class implicitly knows the URL for the sent referral location?
+        val location = UriComponentsBuilder.fromHttpUrl(interventionsUIURI)
+          .path("referrals/{id}/confirmation")
+          .buildAndExpand(event.referral.id)
+          .toUri()
+
+        // fixme: this email address will eventually come from the provider associated with the referral
+        notifyService.referralSent(event.referral.referenceNumber!!, location, "tom.myers@digital.justice.gov.uk")
+      }
+    }
+  }
+}
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/events/ReferralEventPublisher.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/events/ReferralEventPublisher.kt
@@ -1,0 +1,19 @@
+package uk.gov.justice.digital.hmpps.hmppsinterventionsservice.events
+
+import org.springframework.context.ApplicationEvent
+import org.springframework.context.ApplicationEventPublisher
+import org.springframework.stereotype.Component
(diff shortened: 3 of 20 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/config/NotifyConfiguration.kt
```diff
@@ -0,0 +1,14 @@
+package uk.gov.justice.digital.hmpps.hmppsinterventionsservice.config
+
+import org.springframework.beans.factory.annotation.Value
+import org.springframework.context.annotation.Bean
+import org.springframework.context.annotation.Configuration
+import uk.gov.service.notify.NotificationClient
+
+@Configuration
+class NotifyConfiguration(@Value("\${notify.apikey:}") private val apiKey: String) {
```
tomsmyers, line 9: notice the trailing colon on the property here (it means default value of empty string). this is the only way i could get this to work without having the property name in the config file. we only want this value to come from env vars though.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/jpa/service/ReferralService.kt
```diff
@@ -24,7 +25,18 @@ class ReferralService(val repository: ReferralRepository) {
...
+    val sentReferral = repository.save(referral)
+
+    // fixme: this code is repeated from the controller. i only want to do it once, but where?
+    val location = ServletUriComponentsBuilder
+      .fromCurrentContextPath()
+      .path("/sent-referral/{id}")
+      .buildAndExpand(sentReferral.id)
+      .toUri()
+
+    notifyService.referralSent(location, "tom.myers@digital.justice.gov.uk")
```
tomsmyers, line 37: need to figure out where and what this email address is

src/main/resources/db/migration/V1_13_1__created_by_auth_user.sql
```diff
@@ -0,0 +1,24 @@
+-- add the new column type that will replace both 'created_by_userid' and 'created_by_user_auth_source'
+alter table referral
+    add column created_by_id text,
```
karenmillermoj, line 3: Should this be not null - similar to the created_by_userid we used to have?

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/jpa/service/ReferralService.kt
```diff
@@ -24,7 +25,18 @@ class ReferralService(val repository: ReferralRepository) {
 
     // fixme: hardcoded for now
     referral.referenceNumber = "HDJ2123F"
-    return repository.save(referral)
+    val sentReferral = repository.save(referral)
+
+    // fixme: this code is repeated from the controller. i only want to do it once, but where?
```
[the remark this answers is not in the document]
karenmillermoj: I think the controller should pass the location into the service if we need to pass that on.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/jpa/service/NotifyService.kt
```diff
@@ -0,0 +1,34 @@
...
+    if (enabled) {
+      try {
+        val response = client.sendEmail(
+          referralSentTemplateID,
+          recipientEmail,
+          mapOf("referralUrl" to location.toString()),
+          null,
+        )
+      } catch (e: NotificationClientException) {
+        log.error("referral sent notification failed", e)
```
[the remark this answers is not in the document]
sldblog: Can it crash without a `catch`?

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/events/ReferralEventPublisher.kt
```diff
@@ -0,0 +1,19 @@
...
+import org.springframework.context.ApplicationEvent
+import org.springframework.context.ApplicationEventPublisher
+import org.springframework.stereotype.Component
+import uk.gov.justice.digital.hmpps.hmppsinterventionsservice.jpa.entity.Referral
+
+enum class ReferralEventType {
+  SENT
+}
+
+class ReferralEvent(source: Any, val type: ReferralEventType, val referral: Referral) : ApplicationEvent(source)
```
sldblog, line 12: Do we need the `type`? We only have one event and we could name it `ReferralSentEvent(...)`

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/ReferralService.kt
```diff
@@ -26,7 +31,10 @@ class ReferralService(val repository: ReferralRepository, val authUserRepository
 
     // fixme: hardcoded for now
     referral.referenceNumber = "HDJ2123F"
-    return repository.save(referral)
+
+    val sentReferral = repository.save(referral)
+    eventPublisher.referralSentEvent(sentReferral)
```
sldblog, line 36: I'm wondering -- do we benefit from the `ReferralEventPublisher` abstraction?

It could also be
```kt
return repository.save(referral)
   .also { applicationEventPublisher.publishEvent(ReferralSentEvent(this, sentReferral)) }
```

sldblog (commented):
Sorry for the late review. Had a couple of questions -- it looks like we could reduce the number of abstractions (also very possible that I'm missing something)

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/events/NotifyListeners.kt
```diff
@@ -0,0 +1,29 @@
...
+  @Value("\${interventions.ui.uri}") private val interventionsUIURI: String,
+  private val notifyService: NotifyService,
+): ApplicationListener<ReferralEvent> {
+  override fun onApplicationEvent(event: ReferralEvent){
+    when(event.type) {
+      ReferralEventType.SENT -> {
+        // fixme: is it ok that this class implicitly knows the URL for the sent referral location?
+        val location = UriComponentsBuilder.fromHttpUrl(interventionsUIURI)
+          .path("referrals/{id}/confirmation")
+          .buildAndExpand(event.referral.id)
```
karenmillermoj, line 21: I think it's ok for the listener to know the URL as it's deciding what to do with the event (and where it should go).  Alternatively pick it up from a properties file but I still think it's fine.

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/NotifyServiceTest.kt
```diff
@@ -0,0 +1,45 @@
...
+import org.springframework.test.context.ActiveProfiles
+import uk.gov.service.notify.NotificationClient
+import uk.gov.service.notify.NotificationClientException
+import java.net.URI
+
+@SpringBootTest(properties = ["spring.main.lazy-initialization=true"])
+@ActiveProfiles("test", "local")
+class NotifyServiceTest {
+  @MockBean private lateinit var notificationClient: NotificationClient
+  private val url = URI.create("http://interventions.com")
```
sldblog, line 23: In case of any accidental actual call during the future life of the app -- it's usually good practice to keep these URLs fake, i.e. `https://interventions.example.com`

> The domain names example.com, example.net, example.org, and example.edu are second-level domain names in the Domain Name System of the Internet. They are reserved by the Internet Assigned Numbers Authority (IANA) at the direction of the Internet Engineering Task Force (IETF) as special-use domain names for documentation purposes.

source: https://en.wikipedia.org/wiki/Example.com

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/NotifyServiceTest.kt
```diff
@@ -0,0 +1,45 @@
...
+    val notifyService = NotifyService(true, "templateID", notificationClient)
+    notifyService.referralSent("", url, "tom@tom.tom")
+    verify(notificationClient).sendEmail(eq("templateID"), eq("tom@tom.tom"), any(), isNull())
+  }
+
+  @Test
+  fun `test referralSent doesnt blow up on notify errors`() {
+    whenever(notificationClient.sendEmail(any(), any(), any(), anyOrNull())).thenThrow(NotificationClientException::class.java)
+    val notifyService = NotifyService(true, "templateID", notificationClient)
+    notifyService.referralSent("", url, "tom@tom.tom")
```
sldblog, line 43: 🙋 What do you think about using `assertDoesNotThrow` here to reinforce the test title?

In these cases, as a reader, I would also wonder "ok, but what _does_ happen if an error occurs?" Is that worth testing?

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/events/NotifyReferralEventListenerTest.kt
```diff
@@ -0,0 +1,41 @@
...
+    val referenceNumber = "HAS71263"
+    val referral = Referral(id = id, referenceNumber = referenceNumber, serviceUserCRN = "X123456")
+    val event = ReferralEvent("source", ReferralEventType.SENT, referral)
+    listener.onApplicationEvent(event)
+
+    val locationCaptor = ArgumentCaptor.forClass(URI::class.java)
+    verify(notifyService).referralSent(eq(referenceNumber), capture(URICaptor), anyString())
+
+    // we don't care about the actual URL, but we can assume that it requires the referral id
+    assertThat(URICaptor.value.toString()).contains(id.toString())
```
sldblog, line 39: Related to

> fixme: is it ok that this class implicitly knows the URL for the sent referral location?

I think it's ok as long as we test the logic of how we constructed that URL pattern. Could this change to test substring of `/referrals/{id}/confirmation`?

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/events/ReferralEventPublisher.kt
tomsmyers: the listener has to be scoped to the event class, so having each event as a seperate class means a seperate listener for each event, which just means more boilerplate code when adding new events, instead of just adding a new type to the list.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/ReferralService.kt
tomsmyers: `ApplicationEventPublisher` is in interface only.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/events/NotifyListeners.kt
tomsmyers: i think including in properties is a good idea 👍

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/NotifyServiceTest.kt
tomsmyers: done. 

also tested that an error is logs when the email fails.

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/events/NotifyReferralEventListenerTest.kt
tomsmyers: i'm not sure what that would really test. the string `/referrals/{id}/confirmation` isn't constructed, it comes from config.

tomsmyers: i've changed this test so that your last point is covered

karenmillermoj (approved):
lgtm

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/events/ReferralEventPublisher.kt
sldblog: 🤔 Thinking more about this, two further things popped up

- I'm not sure if we will have anything else as "referral event" as the subsequent ones will be "related" to the referral but happen about "action plan", "appointment", "end of service report"
- We don't know how different listeners have to be in the future -- it doesn't sound like we can yet make a call on what the reasonable categorisation might be. (Branching should be built in when there's more than 1 branch)

I think it's better to revisit this when we have 2 separate events.

PR #68, opened 2021-01-18
create user entities before adding their ids to the new referral column to prevent violating FK constraint

tomsmyers:

create user entities before adding their ids to the new referral column to prevent violating FK constraint

fix broken migration and unblock deploy


karenmillermoj (approved):
lgtm

sldblog (approved):

PR #70, opened 2021-01-19
Upload csv

karenmillermoj:

Provides an API to add interventions via csv format.

Set up the initial Intervention and provide the ability to upload a basic Intervention via CSV to enable current interventions for service providers to be easily uploaded into the system.


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/controller/InterventionCsvController.kt
```diff
@@ -0,0 +1,23 @@
...
+import org.springframework.web.bind.annotation.PostMapping
+import org.springframework.web.bind.annotation.RequestBody
+import org.springframework.web.bind.annotation.RequestParam
+import org.springframework.web.bind.annotation.RestController
+import org.springframework.web.multipart.MultipartFile
+import uk.gov.justice.digital.hmpps.hmppsinterventionsservice.dto.InterventionDTO
+import uk.gov.justice.digital.hmpps.hmppsinterventionsservice.jpa.service.InterventionCsvService
+
+@RestController
+class InterventionCsvController(
```
tomsmyers, line 12: i don't think this should be a seperate controller

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/controller/InterventionController.kt
```diff
@@ -0,0 +1,34 @@
...
+  @GetMapping("/intervention/{id}")
+  fun getInterventionByID(@PathVariable id: String): InterventionDTO {
+    val uuid = parseID(id)
+
+    return interventionService.getIntervention(uuid)
+      ?.let { InterventionDTO.from(it) }
+      ?: throw ResponseStatusException(HttpStatus.NOT_FOUND, "intervention not found [id=$uuid]")
+  }
+
+  private fun parseID(id: String): UUID {
```
tomsmyers, line 27: this is repeated from elsewhere. let's refactor to avoid this.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/controller/InterventionCsvController.kt
```diff
@@ -0,0 +1,23 @@
...
+  private val csvService: InterventionCsvService
+) {
+
+  @PostMapping("/intervention/csv/uploadfile")
+  fun uploadCsvFile(@RequestParam("file") file: MultipartFile): ResponseEntity<List<InterventionDTO>> {
```
tomsmyers, line 17: we should discuss if a file is the most appropriate way to represent this information.

body with content-type `text/csv` may be simpler and more explicit.

@sldblog what do you think ?

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/controller/InterventionCsvController.kt
```diff
@@ -0,0 +1,23 @@
...
+import org.springframework.web.multipart.MultipartFile
+import uk.gov.justice.digital.hmpps.hmppsinterventionsservice.dto.InterventionDTO
+import uk.gov.justice.digital.hmpps.hmppsinterventionsservice.jpa.service.InterventionCsvService
+
+@RestController
+class InterventionCsvController(
+  private val csvService: InterventionCsvService
+) {
+
+  @PostMapping("/intervention/csv/uploadfile")
```
tomsmyers, line 16: we should align on the new api endpoints in general, but i'd expect this one to be something like `POST /intervention/add` or `intervention/new`.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/controller/InterventionCsvController.kt
```diff
@@ -0,0 +1,23 @@
...
+  fun uploadCsvFile(@RequestParam("file") file: MultipartFile): ResponseEntity<List<InterventionDTO>> {
+    val importedEntries = csvService.uploadCsvFile(file)
+    return ResponseEntity.ok(importedEntries)
```
tomsmyers, line 19: i'd expect there to be some other return codes present here for case like "invalid input data" for example

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/dto/InterventionDTO.kt
```diff
@@ -0,0 +1,23 @@
+package uk.gov.justice.digital.hmpps.hmppsinterventionsservice.dto
+
+import uk.gov.justice.digital.hmpps.hmppsinterventionsservice.jpa.entity.Intervention
+import java.time.OffsetDateTime
+import java.util.UUID
+
+data class InterventionDTO(
+  val id: UUID? = null,
+  val name: String? = null,
+  val createdAt: OffsetDateTime? = null
```
tomsmyers: i wonder if `lastUpdated` might be a more useful field here instead of `createdAt`?

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/dto/InterventionDTO.kt
```diff
@@ -0,0 +1,23 @@
+package uk.gov.justice.digital.hmpps.hmppsinterventionsservice.dto
+
+import uk.gov.justice.digital.hmpps.hmppsinterventionsservice.jpa.entity.Intervention
+import java.time.OffsetDateTime
+import java.util.UUID
+
+data class InterventionDTO(
+  val id: UUID? = null,
```
tomsmyers, line 8: i don't think any of these field should be nullable

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/jpa/entity/Intervention.kt
```diff
@@ -0,0 +1,16 @@
...
+import javax.persistence.Entity
+import javax.persistence.GeneratedValue
+import javax.persistence.Id
+
+@Entity
+data class Intervention(
+  @Id @GeneratedValue val id: UUID? = null,
+  @CreationTimestamp val createdAt: OffsetDateTime? = null,
+
+  val name: String? = null
```
tomsmyers: i think this should be a required field

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/jpa/service/InterventionService.kt
```diff
@@ -0,0 +1,24 @@
...
+
+  fun createIntervention(interventionDto: InterventionDTO): Intervention {
+    return repository.save(Intervention(name = interventionDto.name))
+  }
+
+  fun getIntervention(id: UUID): Intervention? {
+    return repository.findByIdOrNull(id)
+  }
+
+  fun createInterventions(interventions: List<Intervention>): List<Intervention> {
```
tomsmyers, line 29: i'm slightly confused reading this between the method signatures of `createInterventions` vs `createIntervention`. why do they take different intervention objects? can we just slim this down to a single endpoint?

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/exception/BadRequestException.kt
```diff
@@ -0,0 +1,7 @@
+package uk.gov.justice.digital.hmpps.hmppsinterventionsservice.exception
```
tomsmyers, line 1: this isn't nessecary. spring has one already https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/web/server/ServerWebInputException.html

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/dto/InterventionDTO.kt
```diff
@@ -6,17 +6,18 @@ import java.util.UUID
 
 data class InterventionDTO(
   val id: UUID? = null,
-  val name: String? = null,
-  val createdAt: OffsetDateTime? = null
-
+  val title: String? = null,
+  val createdAt: OffsetDateTime? = null,
+  val regionList: String,
```
tomsmyers, line 11: i think this would be better as an actual `List<String>`

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/jpa/entity/Region.kt
```diff
@@ -0,0 +1,12 @@
+package uk.gov.justice.digital.hmpps.hmppsinterventionsservice.jpa.entity
+
+import javax.persistence.Entity
+import javax.persistence.Id
+import javax.persistence.Table
+
+@Entity
+@Table(name = "region")
+data class Region(
+  @Id val id: Char?,
```
tomsmyers, line 10: let's match the region ids to the ones in delius instead of using our own "A,B,C,D" ids

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/dto/InterventionDTO.kt
karenmillermoj: I agree - it was initially a String to help with the mapping from csv.  I didn't quite get this working (as a list) so reverted back....I'll swap back as that is definitely what we want.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/jpa/entity/Region.kt
karenmillermoj: I picked this up from the spreadsheet David provided - I assumed that's what it was using.  I am unsure how to get these values from delius.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/jpa/service/InterventionService.kt
karenmillermoj: It wasdone to map thru to the save and saveAll repository calls.  Happy to switch to just one.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/jpa/entity/Intervention.kt
karenmillermoj: yes agree - newer version it is also called title.  I'm thinking all of these should be non-null too.  There are additional fields in the newer version

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/exception/BadRequestException.kt
karenmillermoj: updated

src/main/resources/db/migration/V1_16__intervention.sql
```diff
@@ -0,0 +1,23 @@
+create table intervention (
+    id uuid not null,
+    created_at timestamp with time zone not null,
+    type text not null,
+    title text not null,
```
tomsmyers, line 5: there isn't a corresponding field in the entity class for this column

karenmillermoj: Will add in - a few need matching up (wasn't sure how many to add in and so pushed some wip back) - they are also needed for dto

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/dto/InterventionDTO.kt
karenmillermoj: Not sure tbh - will we ever let the intervention be updated?  if so would we want both? For auditing, do we need to link through to createdBy and updatedBy  through to the new auth_user table?  Or are we not bothered at this stage?

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/controller/InterventionCsvController.kt
karenmillermoj: ok - I have no preference - moved to same controller but kept service separate

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/jpa/entity/Intervention.kt
```diff
@@ -0,0 +1,40 @@
...
+import javax.persistence.JoinTable
+import javax.persistence.ManyToMany
+import javax.persistence.ManyToOne
+
+enum class AgeGroup {
+  EIGHTEEN_TO_25, OVER_25
+}
+
+@Entity
+data class Intervention(
```
sldblog, line 21: Sorry for the late review.

"Intervention" will be a confusing term because it can mean a lot of things, depending on what is happening: it can mean

- overall idea: "we use interventions to reduce reoffending"
- specific offering: "we offer these interventions"
- what is happening: "I am referring you to this intervention"

The meaning of the word depends heavily on who is speaking and point of time.

This entity is meant to describe the service offered by Dynamic Framework providers, right?

> This "thing" is offered in "these regions" and covers "this category".

We have called this previously as service configuration or contract scopes. Can we call these `ProviderContract` or `ProviderScope` or similar?

sldblog (commented):
Sorry, a potentially huge impact question

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/jpa/entity/Intervention.kt
tomsmyers: to answer just the last part of the question - it's my understanding that a provider can offer a service in the same service category for different groups of SUs, i.e. women specific accommodation service in west yorks or 18-25 ETE service in south wales.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/jpa/entity/Intervention.kt
```diff
@@ -0,0 +1,40 @@
...
+  @ManyToOne @JoinColumn(name = "service_category_id")
+  var serviceCategory: ServiceCategory? = null,
+
+  @ManyToMany
+  @JoinTable(
+    name = "intervention_region",
+    joinColumns = [JoinColumn(name = "intervention_id")],
+    inverseJoinColumns = [JoinColumn(name = "region_id") ]
+  )
+  var regions: List<Region>? = emptyList()
```
tomsmyers, line 39: since regions is unordered, it seems like (according to the post below) using a `Set` may produce more efficient sql for this field

https://vladmihalcea.com/the-best-way-to-use-the-manytomany-annotation-with-jpa-and-hibernate/

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/controller/InterventionController.kt
karenmillermoj: Pulled into separate StringHelper file.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/dto/InterventionDTO.kt
karenmillermoj: Yeah, the only issue is that we use InterventionDTO for opencsv to map the csv file to a bean.  So I can set everything apart from id (and possibly created - if we keep that field) to non-null.  The only other option would be to have incoming/outgoing dtos.

PR #72, opened 2021-01-20
refactor error response generation to make adding new error types simpler

tomsmyers:

refactor error response generation to make addind new error types simpler

make adding new error types require less boilerplate


karenmillermoj (approved):
great

PR #73, opened 2021-01-21
Add Interventions and its associated entities

karenmillermoj:

Adds the data and model configuration for Interventions and its related entities, including ServiceProvider, Contract, Region and PCCRegion

Introduce interventions into the system, forming a base to build a more detailed entity.


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/utils/StringHelper.kt
```diff
@@ -0,0 +1,15 @@
+package uk.gov.justice.digital.hmpps.hmppsinterventionsservice.utils
+
+import org.springframework.web.server.ServerWebInputException
+import java.util.UUID
+
+object StringHelper {
+  @JvmStatic
```
tomsmyers, line 7: the idiomatic way to do this in kotlin is using a companion object

i.e. 
```
class StringHelper {
  companion object {
    fun parseId() {
    ...
```

this way there's no need for the `@JvmStatic` annotation either

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/dto/InterventionDTO.kt
```diff
@@ -0,0 +1,34 @@
...
+
+import uk.gov.justice.digital.hmpps.hmppsinterventionsservice.jpa.entity.AgeGroup
+import uk.gov.justice.digital.hmpps.hmppsinterventionsservice.jpa.entity.Intervention
+import uk.gov.justice.digital.hmpps.hmppsinterventionsservice.jpa.entity.InterventionType
+import java.time.OffsetDateTime
+import java.util.UUID
+
+data class InterventionDTO(
+  val id: UUID? = null,
+  val createdAt: OffsetDateTime? = null,
```
tomsmyers, line 11: is there a use case that requires `id` or `createdAt` fields in the DTO?

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/jpa/entity/Intervention.kt
```diff
@@ -0,0 +1,38 @@
...
+
+@Entity
+data class Intervention(
+  @Id @GeneratedValue val id: UUID? = null,
+  @CreationTimestamp val createdAt: OffsetDateTime? = null,
+
+  @NotNull val title: String,
+  @NotNull val description: String,
+
+  @NotNull @Enumerated(EnumType.STRING) var ageGroup: AgeGroup,
```
tomsmyers, line 30: is there a reason for these being `var` not `val`?

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/jpa/entity/ServiceProvider.kt
```diff
@@ -0,0 +1,11 @@
+package uk.gov.justice.digital.hmpps.hmppsinterventionsservice.jpa.entity
+
+import java.util.UUID
+import javax.persistence.Entity
+import javax.persistence.Id
+
+@Entity
+data class ServiceProvider(
+  @Id val id: UUID,
```
tomsmyers, line 9: there's an open question about the service provider id. it seems sensible we should map it to the hmpps auth 'groupCode' for the provider. in which case it would be a string data type. let's confirm when we talk through this in person.

src/main/resources/db/local/V99_0__draft_referrals.sql
```diff
@@ -6,7 +6,15 @@ insert into referral (id, created_at, service_usercrn, completion_deadline, crea
 values ('ac386c25-52c8-41fa-9213-fcf42e24b0b5', '2020-12-07 18:02:01.599803+00', 'X987623', '2021-02-14', '2500128586', null),
        ('dfb64747-f658-40e0-a827-87b4b0bdcfed', '2020-12-07 20:45:21.986389+00', 'X456234', '2021-03-01', '8751622134', null),
        ('d496e4a7-7cc1-44ea-ba67-c295084f1962', '2020-12-24 09:32:32.871623+00', 'X542343', '2021-01-30', '2500128586', '428ee70f-3001-4399-95a6-ad25eaaede16'),
-       ('1219a064-709b-4b6c-a11e-10b8cb3966f6', '2021-01-12 14:46:21.987234+00', 'X862134', null, '2500128586', '428ee70f-3001-4399-95a6-ad25eaaede16');
+       ('1219a064-709b-4b6c-a11e-10b8cb3966f6', '2021-01-12 14:46:21.987234+00', 'X862134', null, '2500128586', '428ee70f-3001-4399-95a6-ad25eaaede16'),
+       ('037cc90b-beaa-4a32-9ab7-7f79136e1d27', '2021-01-12 14:46:21.987234+00', 'X872391', null, '2500128586', '428ee70f-3001-4399-95a6-ad25eaaede16'),
+       ('81d754aa-d868-4347-9c0f-50690773014e', '2020-12-07T20:45:21.986389+00', 'X903472', null, '8751622134', '428ee70f-3001-4399-95a6-ad25eaaede16');
```
tomsmyers, line 11: this will break any pact tests which use state "Given a single referral for user with ID 8751622134 exists" (since after this migration there will be two referrals for this user id)

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/dto/InterventionDTO.kt
```diff
@@ -0,0 +1,34 @@
...
+  val title: String,
+  val description: String,
+  val ageGroup: AgeGroup,
+  var type: InterventionType,
+  var serviceCategory: String,
+  var serviceProvider: String,
```
tomsmyers, line 17: i think it would make sense to have a service provider DTO which potentially includes an id that can be used by the front end to get more information about the SP.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/utils/StringHelper.kt
tomsmyers: nitpick: since this is no longer a private method in a single controller, a more descriptive method name or small docstring comment might be useful.

karenmillermoj: Thanks - I didn't know the kotlin way

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/dto/InterventionDTO.kt
karenmillermoj: yeah- wasn't sure tbh.  Will remove until we actually need them

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/jpa/entity/Intervention.kt
karenmillermoj: good spot

src/main/resources/db/local/V99_0__draft_referrals.sql
karenmillermoj: swapped over to user 2500128586

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/dto/InterventionDTO.kt
karenmillermoj: yeah - I wasn't sure about this tbh...but left as a string as I wasn't sure if it would be needed. I'll add that in if we're gonna need it.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/jpa/entity/ServiceProvider.kt
karenmillermoj: updated to string type (var char(30) in sql)

src/main/resources/db/migration/V1_18__contract.sql
```diff
@@ -0,0 +1,34 @@
...
+
+  region_id char,
+  pcc_region_id text,
+
+  primary key (id),
+  constraint fk__contract__service_category_id foreign key (service_category_id) references service_category,
+  constraint fk__contract__service_provider_id foreign key (service_provider_id) references service_provider,
+  constraint fk__contract__pcc_region_id foreign key (pcc_region_id) references pcc_region,
+  constraint fk__contract__region_id foreign key (region_id) references region,
+  constraint contract_dates_valid check (start_date < end_date),
```
tomsmyers, line 19: cool - but is this something we want to validate here, or in the code?

src/main/resources/db/migration/V1_18__contract.sql
```diff
@@ -0,0 +1,34 @@
...
+  constraint contract_region_valid check
+      ((pcc_region_id is not null and region_id is null) or (pcc_region_id is null and region_id is not null))
+);
+
+create table contract_detail (
+    contract_id uuid not null,
+    age_group varchar(30) not null,
```
tomsmyers, line 26: could this be two integer fields, min_age and max_age, to make it clearer?

src/main/resources/db/migration/V1_15__service_provider.sql
```diff
@@ -0,0 +1,7 @@
+create table service_provider (
+  id varchar(30) not null,
+  name text not null,
+  email_address text,
```
tomsmyers, line 4: i think we should more tightly scope what the email is e.g. incoming_referral_distribution_email or something

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/jpa/entity/Region.kt
```diff
@@ -0,0 +1,11 @@
+package uk.gov.justice.digital.hmpps.hmppsinterventionsservice.jpa.entity
+
+import javax.persistence.Entity
+import javax.persistence.Id
+import javax.validation.constraints.NotNull
+
+@Entity
+data class Region(
```
tomsmyers, line 8: can we call this `NPSRegion` - it more closely aligns with the terminology i've seen in the DF documentation

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/jpa/entity/PCCRegion.kt
```diff
@@ -0,0 +1,13 @@
+package uk.gov.justice.digital.hmpps.hmppsinterventionsservice.jpa.entity
+
+import javax.persistence.*
```
tomsmyers, line 3: heads up - i think this may fail the ktlint source checks

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/jpa/entity/Intervention.kt
```diff
@@ -0,0 +1,26 @@
...
+
+@Entity
+data class Intervention(
+  @NotNull @Id @GeneratedValue val id: UUID? = null,
+  @NotNull @CreationTimestamp val createdAt: OffsetDateTime? = null,
+
+  @NotNull val title: String,
+  @NotNull val description: String,
+
+  @NotNull @OneToOne(fetch = FetchType.EAGER, cascade = [CascadeType.ALL]) @JoinColumn(name = "contract_id")
```
tomsmyers, line 23: what's the reason for eager fetching on this field?

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/jpa/entity/Contract.kt
```diff
@@ -0,0 +1,63 @@
...
+
+  @ManyToOne @JoinColumn(name = "region_id")
+  val region: Region? = null,
+
+  @ManyToOne @JoinColumn(name = "pcc_region_id")
+  val pccRegion: PCCRegion? = null,
+)
+
+@Embeddable
+data class ContractDetail(
```
tomsmyers, line 55: this is neat - but what's the reason these fields aren't just part of the contract entity ?

src/main/resources/db/migration/V1_18__contract.sql
karenmillermoj: I'm not sure tbh - should we just be adding in the code and not here?

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/jpa/entity/Region.kt
karenmillermoj: sure

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/jpa/entity/PCCRegion.kt
karenmillermoj: will update

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/jpa/entity/Intervention.kt
karenmillermoj: Wouldn't we always want to retrieve a contract when an intervention is retrieved?  There is so much information that the intervention relies upon.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/jpa/entity/Contract.kt
karenmillermoj: I think the "optional" fields were pulled out separately from an initial suggestion.  Happy either way tbh

tomsmyers (approved):
all good except the constraints thing we talked about yesterday

src/main/resources/db/migration/V1_18__contract.sql
karenmillermoj: Business logic will be written in the code - constraints removed

src/main/resources/db/migration/V1_15__service_provider.sql
karenmillermoj: renamed

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/jpa/entity/Intervention.kt
karenmillermoj: yes do not want the intervention to remove the contract - changed mapping

src/main/resources/db/migration/V1_18__contract.sql
karenmillermoj: Separated out

PR #74, opened 2021-01-21
🐛 Re-enable multiple referrals by the same user

sldblog:

Re-enables starting referrals by the same user -- currently 🐛 not working due to

```
'could not execute statement; SQL [n/a]; constraint [auth_user_pkey];
⮑ nested exception is org.hibernate.exception.ConstraintViolationException: could not execute statement'
```

The above error happens when the same user is trying to create a new draft referral.

This is due to
- the primary key constraint on the `auth_user` table
- the `cascade = [CascadeType.PERSIST]` setting on `Referral` for the
  `createdBy` field

It seems the persistence framework cannot or did not decide whether
the `AuthUser` instance has been persisted yet or not.

I've tried a couple of things:
- overriding the repository's `save` function
- playing with other cascade types
- persisting the user directly

Decided on the last one as it's the simplest to understand solution.

With explicit `authUserRepository.save(...)` calls in the
`ReferralService`,  we can guarantee that the `Referral` object does not
have transient linked data.


tomsmyers (approved):
thanks for this - good solution.

PR #75, opened 2021-01-21
Re-enable migration and exception logs

sldblog:

Make 🗒️ migrations and 💥 exceptions visible in
- local development console
- and on Kubernetes pod logs

| Before | After |
| --- | --- |
| image | image <br> Plus exceptions are printed, not shown on the screenshot |

**On the Kubernetes environments, we should always log to stdout.**

Currently, the migration logs and raised exceptions are not visible locally or on the development Kubernetes pod logs.

It looks like a lot of messages are redirected to `ApplicationInsightsAppender` -- which we do not yet use.

I like seeing the migration status and exceptions on the screen, so: this commit reverts back to default log settings until we configure AppInsights.


tomsmyers:
thanks for doing this. happy to merge for now, but i'd like to make it a priority to put proper logback config back in ASAP.

tomsmyers (approved):

sldblog:
> proper logback config back in ASAP

Sure! Do we have it defined somewhere?

PR #77, opened 2021-01-22
IC-972 store SU disabilities as list of strings 

tomsmyers:

stores disabilities as a list of strings instead of a single string. 

see this thread for more details https://github.com/ministryofjustice/hmpps-interventions-ui/pull/82#discussion_r562628835

allow the front end to store structured data about SU disabilities instead of squashing them together in a single string.

### DIFF
--- a/src/main/resources/db/migration/V1_14__disabilities_list.sql
+++ b/src/main/resources/db/migration/V1_14__disabilities_list.sql
@@ -1,0 +1,2 @@
+alter table referral_service_user_data
+    alter disabilities type text[] using array[disabilities];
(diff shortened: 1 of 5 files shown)


Gweaton (approved):
Thanks, sorry for the faff!

src/main/resources/db/migration/V1_14__disabilities_list.sql
```diff
@@ -0,0 +1,2 @@
+alter table referral_service_user_data
+    alter disabilities type text[] using array[disabilities];
```
sldblog, line 2: TIL about `using array[...]`

sldblog (approved):

PR #78, opened 2021-01-22
remove unneeded 'CascadeType' annotations 

tomsmyers:
this is related to 97d8aa2a90ba850acd87cccdc8ccfadef37d8f2f - these should probably have been taken out then

remove unneeded 'CascadeType' annotations 

code clean up

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/jpa/entity/Referral.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/jpa/entity/Referral.kt
@@ -24,12 +24,11 @@ import javax.validation.constraints.NotNull
 @Table(indexes = arrayOf(Index(columnList = "created_by_id")))
 data class Referral(
   var sentAt: OffsetDateTime? = null,
-  @ManyToOne(fetch = FetchType.LAZY, cascade = [CascadeType.PERSIST]) var sentBy: AuthUser? = null,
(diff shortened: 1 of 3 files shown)


src/main/resources/application-local.yml
```diff
@@ -16,3 +16,4 @@ logging:
 spring:
   flyway:
     locations: classpath:db/migration,classpath:db/local
+    clean-on-validation-error: true
```
sldblog, line 19: I'm assuming this "clean if migrations are invalid" is back because it's useful in the local env, especially when it comes to seed data changing.

_Not part of this PR_
Would it make sense to create a standalone command for seeding, not treating it as a migration?

For example, in Rails and Phoenix there's a `bin/rake db:seed` and `mix run priv/repo/seeds.exs` for similar purposes, and it's "on demand" rather than automatic. (And then people usually have a shortcut "reset" commands that drop/create/migrate/seed.)

sldblog (approved):
👍 Nice. No blockers.

src/main/resources/application-local.yml
sldblog: _Edit_
Seems like repeatable migrations might have similar semantics (emphasis mine):

> Repeatable migrations have a description and a checksum, but no version. Instead of being run just once, they are (re-)applied **every time their checksum changes**

This might work well for local data.

tomsmyers: ah no - this is a mistake - it should not be back! removing now. good spot.

PR #79, opened 2021-01-25
Remove data engineering review requirement

sldblog:

Remove data engineering PR review requirement for database changes.
I can't say if this change is temporary or not.

Currently, this review feels like a roadblock rather than a collaboration.


tomsmyers (approved):
thanks - i agree this often blocks reviews since you are the only person really able to review on that github group. 

however i do want to ensure you get eyes on these reviews where approriate, we'll just to ensure we alert each other in slack when nessecary

sldblog:
🙇‍♂️ 🎩

PR #84, opened 2021-01-28
refactor the notify event listener so it's completely decoupled and isolated

tomsmyers:

implement the event listener interface in the notify service instead of a seperate class

improve the structure of the async event listeners and decouple all the event actions.


sldblog (approved):
👍 LGTM

PR #86, opened 2021-02-01
IC-1012 add ServiceProviderDTO definition and include it in the DraftReferralDTO

tomsmyers:

IC-1012 add ServiceProviderDTO definition and include it in the DraftReferralDTO.

this allows the UI to display service provider name in the 'confirm answers' page without an additional API call. 

reduce number of API calls required to get information about the service provider.


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/ReferralService.kt
```diff
@@ -32,17 +34,24 @@ class ReferralService(
...
   }
 
   fun createDraftReferral(user: AuthUser, crn: String): Referral {
-    return repository.save(Referral(serviceUserCRN = crn, createdBy = authUserRepository.save(user)))
+    val interventionId = UUID.fromString("98a42c61-c30f-4beb-8062-04033c376e2d")
+    return referralRepository.save(
+      Referral(
+        serviceUserCRN = crn,
+        createdBy = authUserRepository.save(user),
+        intervention = interventionRepository.getOne(interventionId)
```
karenmillermoj, line 48: Do we need to ensure that the intervention actually exists and throw an exception/Validation error if not?

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/jpa/entity/Referral.kt
```diff
@@ -51,6 +51,7 @@ data class Referral(
   var desiredOutcomesIDs: List<UUID>? = null,
   var completionDeadline: LocalDate? = null,
 
+  @NotNull @ManyToOne(fetch = FetchType.LAZY) val intervention: Intervention,
```
karenmillermoj, line 54: The @NotNull means the intervention can't be null - but doesn't a referral get created without assigning an intervention to it?

src/main/resources/db/migration/V1_21__associate_referral_with_intervention.sql
```diff
@@ -0,0 +1,3 @@
+alter table referral
+    add column intervention_id uuid,
+    add constraint fk_intervention_id foreign key (intervention_id) references intervention;
```
karenmillermoj, line 3: Should this have a not null?

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/ReferralService.kt
tomsmyers: yes probably on the first part. but we can do that in the controller (i.e. validate the request body)

on the second part (contractual validity). i'm not sure if we need to do that. at least not right now.

@sldblog any thoughts?

src/main/resources/db/migration/V1_21__associate_referral_with_intervention.sql
tomsmyers: now we have dev data, this is done.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/ReferralService.kt
tomsmyers: adding code to parse this in the controller so i can do the first part

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/jpa/entity/PCCRegion.kt
```diff
@@ -12,6 +13,6 @@ import javax.validation.constraints.NotNull
 data class PCCRegion(
   @NotNull @Id val id: String,
   @NotNull val name: String,
-  @NotNull @ManyToOne @JoinColumn(name = "nps_region_id")
+  @NotNull @ManyToOne(fetch = FetchType.LAZY) @JoinColumn(name = "nps_region_id")
```
karenmillermoj, line 16: Yep, good spot on this

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/jpa/repository/InterventionRepository.kt
```diff
@@ -1,9 +1,9 @@
 package uk.gov.justice.digital.hmpps.hmppsinterventionsservice.jpa.repository
 
-import org.springframework.data.repository.CrudRepository
+import org.springframework.data.jpa.repository.JpaRepository
 import uk.gov.justice.digital.hmpps.hmppsinterventionsservice.jpa.entity.Intervention
 import java.util.UUID
 
-interface InterventionRepository : CrudRepository<Intervention, UUID> {
+interface InterventionRepository : JpaRepository<Intervention, UUID> {
```
tomsmyers, line 7: this is required for `repository.getOne` which returns a shallow reference to the entity

Gweaton (approved):
The DTO looks like what we're expecting from a UI perspective. Can you see any reason we might want to include an `id`, or do you think just the `name` is fine for now?

src/main/resources/db/local/V99_2__draft_referrals.sql
```diff
@@ -0,0 +1,23 @@
...
+       ('2a67075a-9c77-4103-9de0-63c4cfe3e8d6', null, null, null, null, 'Alex', null, null, null, null),
+       ('81d754aa-d868-4347-9c0f-50690773014e', '{Autism spectrum condition, sciatica}', TO_DATE('1980-01-01', 'YYYY-MM-DD'), 'British', 'Mr', 'Alex', 'River', 'English', 'Agnostic', 'Male');
+
+insert into referral_desired_outcome (referral_id, desired_outcome_id)
+values ('81d754aa-d868-4347-9c0f-50690773014e', '301ead30-30a4-4c7c-8296-2768abfb59b5'),
+       ('81d754aa-d868-4347-9c0f-50690773014e', '65924ac6-9724-455b-ad30-906936291421');
+--        ('037cc90b-beaa-4a32-9ab7-7f79136e1d27', '301ead30-30a4-4c7c-8296-2768abfb59b5'),
+--        ('037cc90b-beaa-4a32-9ab7-7f79136e1d27', '65924ac6-9724-455b-ad30-906936291421'),
+--        ('d496e4a7-7cc1-44ea-ba67-c295084f1962', '9b30ffad-dfcb-44ce-bdca-0ea49239a21a'),
+--        ('d496e4a7-7cc1-44ea-ba67-c295084f1962', 'e7f199de-eee1-4f57-a8c9-69281ea6cd4d');
```
karenmillermoj, line 23: Aren't the removed desired outcomes required for the Pact Tests:
There is an existing draft referral with ID of 037cc90b-beaa-4a32-9ab7-7f79136e1d27, and it has had desired outcomes selected 
- the bottom two may not be needed (They were probably added for additional testing)

src/main/resources/db/migration/V1_21__associate_referral_with_intervention.sql
karenmillermoj: I don't understand the default value on the intervention_id - wouldn't this fail anyway as the intervention with that id ('0000...') doesn't exist?

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/dto/ServiceProviderDTO.kt
```diff
@@ -0,0 +1,13 @@
...
+import uk.gov.justice.digital.hmpps.hmppsinterventionsservice.jpa.entity.ServiceProvider
+
+class ServiceProviderDTO(
+  val name: String,
+) {
+  companion object {
+    fun from(serviceProvider: ServiceProvider): ServiceProviderDTO {
+      return ServiceProviderDTO(name = serviceProvider.name)
+    }
+  }
```
karenmillermoj, line 12: Do we need a DTO just for a String ?

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/jpa/entity/DynamicFrameworkContract.kt
```diff
@@ -32,10 +33,10 @@ data class DynamicFrameworkContract(
   @Embedded
   val contractEligibility: ContractEligibility,
 
-  @ManyToOne @JoinColumn(name = "nps_region_id")
+  @ManyToOne(fetch = FetchType.LAZY) @JoinColumn(name = "nps_region_id")
   val npsRegion: NPSRegion? = null,
 
-  @ManyToOne @JoinColumn(name = "pcc_region_id")
+  @ManyToOne(fetch = FetchType.LAZY) @JoinColumn(name = "pcc_region_id")
   val pccRegion: PCCRegion? = null,
```
karenmillermoj, line 40: I'm not sure we want lazy loading.  The current service calls to get interventions and contracts are immediately accessing the service category/provider/nps and pcc regions and manipulating them into DTOs - so wouldn't the lazy loading just result in multiple calls to the database rather than just one eager load?
I can understand for a oneToMany - but these are small objects.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/dto/ServiceProviderDTO.kt
tomsmyers: good point. this matches with the UI expectations of the response though. and if we add more more SP fields in the future this existing one won't have to change

src/main/resources/db/local/V99_2__draft_referrals.sql
tomsmyers: sorry this is probably a merge error. loads of the pacts are still failing. let's address the pact issues outside this PR

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/jpa/entity/DynamicFrameworkContract.kt
tomsmyers: yeh good point. i think you're right that we will likely always want to access the service provider and service category information when returning results about interventions. for regions i'm not sure.  our existing endpoints do not pull region data. but i also agree they're small objects. will pull these changes out for now.

src/main/resources/db/migration/V1_21__associate_referral_with_intervention.sql
karenmillermoj: Has the not null been removed accidentally in this last update?

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/dto/DraftReferralDTO.kt
```diff
@@ -44,6 +46,8 @@ data class DraftReferralDTO(
         maximumRarDays = referral.maximumRarDays,
         desiredOutcomesIds = referral.desiredOutcomesIDs,
         serviceUser = ServiceUserDTO.from(referral.serviceUserCRN, referral.serviceUserData),
+        serviceProvider = ServiceProviderDTO.from(contract.serviceProvider),
```
sldblog, line 49: _suggestion, not a blocker_
The full chain is `referral.intervention.dynamicFrameworkContract.serviceProvider`; would it make sense to add a service to look up contract or provider by a referral?

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/ReferralService.kt
sldblog: > does the intervention contract need to be valid - i.e. check contract dates

👍 probably in a separate PR though

I guess we could also remove the start/end fields... I forgot that if we don't have testable behaviour against it, we shouldn't have data

src/main/resources/db/local/V99_2__draft_referrals.sql
```diff
@@ -0,0 +1,23 @@
+insert into referral (id, intervention_id, created_at, sent_at, service_usercrn, completion_deadline, created_by_id, further_information, additional_needs_information, accessibility_needs, needs_interpreter, interpreter_language, has_additional_responsibilities, when_unavailable, additional_risk_information, using_rar_days, maximum_rar_days, complexity_levelid, reference_number)
+values ('ac386c25-52c8-41fa-9213-fcf42e24b0b5', '98a42c61-c30f-4beb-8062-04033c376e2d', '2020-12-07 18:02:01.599803+00', null, 'X987623', '2021-02-14', '2500128586', null, null, null, null, null, null, null, null, null, null, null, null),
```
sldblog, line 2: _not a blocker_
Hmm, perhaps these would be easier to create in code now? As fixtures/sample data/seeds?

This is reaching 'too much detail to follow' level for me.

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/jpa/entity/SampleData.kt
```diff
@@ -1,12 +1,54 @@
...
 import java.time.LocalDate
 import java.time.OffsetDateTime
 import java.util.UUID
 
 class SampleData {
-
   companion object {
+    // there are tonnes of related tables that need to exist to successfully persist an intervention,
+    // this is a helper method that persists them all
+    fun persistIntervention(em: TestEntityManager, intervention: Intervention): Intervention {
```
sldblog, line 12: Should a `Service` be able to do all this?

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/utils.kt
```diff
@@ -0,0 +1,20 @@
+package uk.gov.justice.digital.hmpps.hmppsinterventionsservice
+
+import org.springframework.security.core.GrantedAuthority
+import org.springframework.security.core.authority.AuthorityUtils
+import org.springframework.security.oauth2.jwt.Jwt
+import org.springframework.security.oauth2.server.resource.authentication.JwtAuthenticationToken
+
+fun createJwtAuthenticationToken(
```
sldblog, line 8: Could this file be renamed to `tokenHelper` or similar, please? "Utilities" usually erodes a lot.

sldblog (approved):
👍 Nice! No blockers from me

Concern around persisting helpers: it feels like a service or fixture library would solve this without explicit test code.

In other frameworks, usually, it's possible to do something like `Factory.create(:intervention, contract: Factory.create(:accommodation_south_wales))` which would then create individually valid fixture records

tomsmyers:
@sldblog 

thanks for the comments, i'm going to address in a new PR to avoid this review dragging on. i think all of the comments are around test code. so i've created this ticket to track, and included your thoughts.

PR #87, opened 2021-02-04
Get interventions

karenmillermoj:

Adds the ability to return all (unfiltered) interventions 

Provide the consumer with a list of all interventions available


karenmillermoj:
I have opened this as a draft request to get feedback on the InterventionDTO.
Note.  Additional tests still need to be added

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/jpa/entity/PCCRegion.kt
```diff
@@ -13,5 +13,5 @@ data class PCCRegion(
   @NotNull @Id val id: String,
   @NotNull val name: String,
   @NotNull @ManyToOne @JoinColumn(name = "nps_region_id")
-  val region: NPSRegion
+  val npsRegion: NPSRegion
```
sldblog, line 16: 👏

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/dto/InterventionDTO.kt
```diff
@@ -5,13 +5,30 @@ import uk.gov.justice.digital.hmpps.hmppsinterventionsservice.jpa.entity.Interve
...
       return InterventionDTO(
         title = intervention.title,
         description = intervention.description,
+        pccRegions = pccRegions,
+        serviceCategory = contract.serviceCategory.name,
+        serviceProvider = contract.serviceProvider.name,
+        minimumAge = contractEligibility.minimumAge,
+        maximumAge = contractEligibility.maximumAge,
+        allowsFemale = contractEligibility.allowsFemale,
+        allowsMale = contractEligibility.allowsMale,
```
sldblog, line 31: The code hints that these belong together -- it's probably ok to model them as
```json
"eligibility": {
  "minimumAge": 18,
  "maximumAge": null,
  "allowsMale": true,
  "allowsFemale": true,
}
```
What do you think @lawrence-forooghian, @Gweaton? What'd be simpler to understand and use from a UI perspective?

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/dto/InterventionDTO.kt
```diff
@@ -5,13 +5,30 @@ import uk.gov.justice.digital.hmpps.hmppsinterventionsservice.jpa.entity.Interve
 data class InterventionDTO(
   val title: String,
   val description: String,
+  val pccRegions: List<String>,
+  val serviceCategory: String,
+  val serviceProvider: String,
```
Gweaton, line 10: Do you think we could append `name` to these? (`serviceCategoryName`, `serviceProviderName`)

sldblog (commented):
> Do we need to return PCCRegionId (as well as name)?

I would say yes, as querying subsequent details or POST requests are only possible with the IDs. I think this is true for all entities (service category, regions, provider)

The intervention's ID would be required for starting a referral, so can we add that too, please?

> Would we ever need to return NPS Region?

🤔 If the contract is NPS region-specific, that's the only data point that we would have. Do we resolve it to all related PCC regions or just display the NPS region?

We don't have to solve this now. For now, what do you think about deferring it until it becomes important?

> Do we need to return the "Type" - DF/Structured (maybe only when Structured Interventions are introduced)?

No. Let's not worry anything about the potential future. So many things can change 😛

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/dto/InterventionDTO.kt
sldblog: Would the object work?
```json
"serviceCategory": {
  "id": ...,
  "name": ...,
}
```

I'm conscious we already have endpoints that return a representation of these concepts, so seems like we could reuse them. Is that valid?

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/dto/InterventionDTO.kt
Gweaton: That's a good point, hadn't thought of that, I think your suggestion looks good @sldblog

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/dto/InterventionDTO.kt
karenmillermoj: I am unsure the best way forward with these,  I am happy to either return as serviceCategoryName/serviceProviderName or as their own objects.
@Gweaton do you have a preference?

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/jpa/entity/PCCRegion.kt
karenmillermoj: 1. Agree - with pccRegionid (also George has raised this) I will add that in - we have a PCCRegionDTO which holds these 2 fields
2. I have assumed (maybe wrong to assume) that we will need to return PCC Regions.  Code in this PR does that in the service layer (although I know I didn't ask you to review that bit yet).  It determines if the NPSRegion is set, and if so, does a call back to the db to get all PCCRegions associated with it.  There might be a need JPA way to do this- but will check.

sldblog: 👍 Good call, it would be good to figure out preloading the linked category entities (service category, NPS regions, PCC regions, etc.) otherwise we'll have a massive N+1 query problem.

karenmillermoj: One other question @sldblog -
What should the value of maximumAge be if we have no upper limit?  null/99/-1 ?

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/dto/InterventionDTO.kt
Gweaton: Sorry @sldblog I'm a bit confused, are you saying we return the whole `serviceCategory` object that's returned from existing endpoints (e.g complexity level, desired outcomes too), or that we should just include `name` and `id` and use the `id` to make requests to the existing `service-category/{id}` endpoints?

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/jpa/entity/PCCRegion.kt
sldblog: for me, semantically, `null` would mean "unset", which I'd interpret as "no limit"
it's a bit harder to implement on client-side but the explicit value is worth it imo

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/dto/InterventionDTO.kt
sldblog: Just IDs would push the necessity of resolving those IDs to the client when displaying the list. That's also fine, but more complex on UI.

I was thinking of the full payload (embedding) because it's simpler for us in theory as we can reuse code that already exists and will be easier to document.

Feels like the complexity levels and desired outcomes could have their own endpoint (not for this PR though)

Gweaton: Ah ok, that makes sense - shall we go for full payload then?

tomsmyers: how about `Summary` DTOs for a few of these entities. e.g. `ServiceCategorySummaryDTO` which are strict subsets of the full DTO. functionally this would be good to have for the interventions themselves too, where endpoints to query interventions return summaries, whereas endpoints to retrieve a specific intervention returns the full entity.

lawrence-forooghian: If we are going to go for the summary approach, then I like having them as explicitly separate DTOs, instead of making a bunch of properties optional in the original DTO.

lawrence-forooghian: After today's dev chat, we decided we'll use the full entities here for now.

tomsmyers (approved):

PR #89, opened 2021-02-08
IC-914 show referral notification in delius

rgforsyth:

When a referral event is triggered, a post request will be sent to community api to insert a row into delius db - contact table with some details of the referral. 

To update the delius contact log with the referral sent notification

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -10,7 +10,10 @@ repositories {
 }
 
 configurations {
-  testImplementation { exclude(group = "org.junit.vintage") }
+  testImplementation {
+    exclude(group = "org.junit.vintage")
+    exclude("ch.qos.logback")
+  }
 }
 
 tasks {
@@ -41,6 +44,7 @@ dependencies {
   // security
   implementation("org.springframework.boot:spring-boot-starter-webflux")
   implementation("org.springframework.boot:spring-boot-starter-oauth2-resource-server")
+  implementation("org.springframework.boot:spring-boot-starter-oauth2-client")
 
   // database
   implementation("org.springframework.boot:spring-boot-starter-data-jpa")
@@ -52,4 +56,5 @@ dependencies {
 
   testImplementation("au.com.dius.pact.provider:junit5spring:4.1.14")
   testImplementation("com.h2database:h2:1.4.200")
+  testImplementation("uk.org.lidalia:slf4j-test:1.0.1")
 }
--- a/src/main/resources/application-local.yml
+++ b/src/main/resources/application-local.yml
@@ -4,6 +4,16 @@ notify:
 interventions-ui:
   baseurl: http://localhost:3000
 
+community-api:
+  baseUrl: http://localhost:8081
+#  Temporarily stored in config until it is decided where these values will come from. TeamCode may come from the community api
+#  in the future. The rest may be hard coded as generic values.
+  temporaryReferralRequest:
+    probationAreaCode: A00
+    referralType: C385
+    staffCode: ASPUATU
+    teamCode: ASPUAT
+
 aws:
   sns:
     enabled: false
@@ -31,3 +41,10 @@ logging:
 spring:
   flyway:
     locations: classpath:db/migration,classpath:db/local
+  security:
+    oauth2:
+      client:
+        registration:
+          community-api:
+            client-id: interventions
+            client-secret: clientsecret
--- a/src/main/resources/application.yml
+++ b/src/main/resources/application.yml
@@ -13,6 +13,15 @@ spring:
       WRITE_DATES_AS_TIMESTAMPS: false
   security:
     oauth2:
+      client:
+        provider:
+          hmppsauth:
+            token-uri: ${hmppsauth.baseurl}/oauth/token
+        registration:
+          community-api:
+            provider: hmppsauth
+            authorization-grant-type: client_credentials
+            scope: read
       resourceserver:
         jwt:
           issuer-uri: ${hmppsauth.baseurl}/issuer
@@ -67,6 +76,10 @@ notify:
 interventions-ui:
   locations:
     sent-referral: "/referrals/{id}/confirmation"
+
+community-api:
+  locations:
+    sent-referral: "/secure/offenders/crn/{crn}/referral/sent"
 
 aws:
   sns:
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/config/WebClientConfiguration.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/config/WebClientConfiguration.kt
@@ -1,0 +1,53 @@
+package uk.gov.justice.digital.hmpps.hmppsinterventionsservice.config
+
+import org.springframework.beans.factory.annotation.Value
+import org.springframework.context.annotation.Bean
+import org.springframework.context.annotation.Configuration
+import org.springframework.security.oauth2.client.AuthorizedClientServiceOAuth2AuthorizedClientManager
+import org.springframework.security.oauth2.client.OAuth2AuthorizedClientManager
+import org.springframework.security.oauth2.client.OAuth2AuthorizedClientProviderBuilder
+import org.springframework.security.oauth2.client.OAuth2AuthorizedClientService
+import org.springframework.security.oauth2.client.registration.ClientRegistrationRepository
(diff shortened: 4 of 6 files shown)


build.gradle.kts
```diff
@@ -10,7 +10,10 @@ repositories {
 }
 
 configurations {
-  testImplementation { exclude(group = "org.junit.vintage") }
+  testImplementation {
+    exclude(group = "org.junit.vintage")
+    exclude("ch.qos.logback")
```
rgforsyth, line 15: Excluded for the tests to prevent conflict with slf4j-test. (Is this an issue?)
Details here: https://projects.lidalia.org.uk/slf4j-test/

build.gradle.kts
```diff
@@ -52,4 +56,5 @@ dependencies {
 
   testImplementation("au.com.dius.pact.provider:junit5spring:4.1.14")
   testImplementation("com.h2database:h2:1.4.200")
+  testImplementation("uk.org.lidalia:slf4j-test:1.0.1")
```
rgforsyth, line 59: allows log messages to be stored in memory for a test, meaning assertions on logs can be made

src/main/resources/application-local.yml
```diff
@@ -4,6 +4,14 @@ notify:
 interventions-ui:
   baseurl: http://localhost:3000
 
+communityapi:
+  baseUrl: http://localhost:8081
+  temporaryReferralRequest:
```
rgforsyth, line 9: Used until we can decide on the default data. Team code made be retrieved from community-api in the future.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/CommunityService.kt
```diff
@@ -0,0 +1,63 @@
...
+  @Value("\${communityapi.temporaryReferralRequest.staffCode}") private val staffCode: String,
+  @Value("\${communityapi.temporaryReferralRequest.teamCode}") private val teamCode: String,
+  @Value("\${interventions-ui.baseurl}") private val interventionUiBaseUrl: String,
+  private val communityApiWebClient: WebClient,
+) : ApplicationListener<ReferralEvent> {
+
+  override fun onApplicationEvent(event: ReferralEvent) {
+    when (event.type) {
+      ReferralEventType.SENT -> {
+        val url = "$interventionUiBaseUrl/referrals/${event.referral.id}/confirmation"
```
tomsmyers, line 27: can we use UriComponentsBuilder (?) to build this url?

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/CommunityService.kt
```diff
@@ -0,0 +1,63 @@
...
+        val body = ReferRequest(
+          probationAreaCode,
+          referralType,
+          staffCode,
+          teamCode,
+          url,
+          event.referral.sentAt?.toLocalDate()
```
tomsmyers, line 34: if we are saying this can't be null, let's make a non-null assertion, this way if it _is_ null, the code blows up

`event.referral.sentAt!!.toLocalDate()`

of course we need to ensure we are handling the potential exception though

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/CommunityService.kt
```diff
@@ -0,0 +1,63 @@
...
+        )
+
+        communityApiWebClient.post().uri("secure/offenders/crn/${event.referral.serviceUserCRN}/referral/sent")
```
tomsmyers, line 37: i would have this in config as `"secure/offenders/crn/{crn}/referral/sent"`

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/CommunityService.kt
```diff
@@ -0,0 +1,63 @@
...
+          .body(Mono.just(body), ReferRequest::class.java)
+          .retrieve()
+          .bodyToMono(Unit::class.java)
+          .onErrorResume { e ->
+            log.error("{}", e.toString())
```
tomsmyers, line 42: more informative log message some thing smiple just like "community api call failed"

src/main/resources/application.yml
```diff
@@ -13,6 +13,17 @@ spring:
       WRITE_DATES_AS_TIMESTAMPS: false
   security:
     oauth2:
+      client:
+        provider:
+          hmppsauth:
+            token-uri: ${hmppsauth.baseurl}/oauth/token
+        registration:
```
tomsmyers, line 20: this config is only valid for 'local' , it shouldn't be at the top level

tomsmyers (commented):
few small things

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/CommunityService.kt
```diff
@@ -0,0 +1,74 @@
...
+        val communityApiSentReferralPath = UriComponentsBuilder.fromPath(communityAPISentReferralLocation)
+          .buildAndExpand(event.referral.serviceUserCRN)
+          .toString()
+
+        communityApiWebClient.post().uri(communityApiSentReferralPath)
+          .body(Mono.just(body), ReferRequest::class.java)
+          .retrieve()
+          .bodyToMono(Unit::class.java)
+          .onErrorResume { e ->
+            log.error("Call to community api to update contact log failed: {}", e.toString())
```
tomsmyers, line 53: sorry missed this before - the second argument here should be the error itself, `e`.

src/main/resources/application-local.yml
```diff
@@ -31,3 +41,16 @@ logging:
 spring:
   flyway:
     locations: classpath:db/migration,classpath:db/local
+  security:
```
tomsmyers, line 44: this config should be split into bits that are common across all environments, and bits that are specific to local/dev etc 

so i would pop ```
security:
  oauth2:
      client:
        provider:
          hmppsauth:
            token-uri: ${hmppsauth.baseurl}/oauth/token
        registration:
          communityapi:
            provider: hmppsauth
            authorization-grant-type: client_credentials
            scope: read
            ```
            in the top-level application.yml file,
            
            and ```
security: 
  oauth2:
      client:
        registration:
          communityapi:
            client-id: interventions
            client-secret: clientsecret
            ```
            in this file.

src/main/resources/application.yml
```diff
@@ -68,6 +68,10 @@ interventions-ui:
   locations:
     sent-referral: "/referrals/{id}/confirmation"
 
+community-api:
```
tomsmyers, line 71: this should match with the toplevel config section name in `application-local.yml`, which is `communityapi`

rgforsyth: Changed to community-api

tomsmyers (approved):
awesome - thanks

src/main/resources/application.yml
```diff
@@ -68,6 +77,10 @@ interventions-ui:
...
+  locations:
+    sent-referral: "/secure/offenders/crn/{crn}/referral/sent"
```
sldblog, line 82: 🙋‍♂️ Could we inline this to the code, please?

The request URL configuration won't change between environments and has the same pattern as `interventions-ui.locations.sent-referral`, which is for a different purpose, which will be confusing

src/main/resources/application-local.yml
sldblog: Could we call it "notification", instead of "referral", please? Maybe `contactNotificationContext`?

It'll be the same for all notification purposes later, not only for referrals.

sldblog (commented):
👏 Nice!

Just a couple of comments about naming and configuration

src/main/resources/application.yml
sldblog: (Also just noticed there's `communityapi.baseUrl` and `community-api.locations....` (different root key)

tomsmyers: hm this is my bad - i think i requested it to go into config to mirror the interventions-ui location url

tomsmyers: i still think it's suitable to live in config

i think by it's nature this is "configuration" (vs code)

PR #90, opened 2021-02-08
IC-1107 endpoint to get list of sent referrals with optional 'service provider id' query param

tomsmyers:

Adds a `GET /sent-referrals` endpoint which accepts an optional `serviceProviderID` query param to allow retrieving sent referrals for a particular service provider.

1. power the new SP dashboard screen
2. enable future filtering of referrals by SP


rgforsyth (approved):

rgforsyth (commented):
Looks good

PR #91, opened 2021-02-08
IC-1108 Improvements to pact tests to make them more maintainable

tomsmyers:

It alters the way we approach the pact testing on the backend, by moving the test data out of the 'local' database migrations, and into the state setup code for each individual test.

there are some functional changes in this PR, which are isolated in their own commits. these are things we have missed because the contract tests have been red for so long.

make the tests easier to add, easier to maintain, and less prone to failing due to data being in inconsistent states.


src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/integration/PactTest.kt
```diff
@@ -5,69 +5,147 @@ import au.com.dius.pact.provider.junitsupport.Provider
...
+      user = deliusUser,
+      crn = "X123456",
+      interventionId = accommodationInterventionID,
+      UUID.fromString("d496e4a7-7cc1-44ea-ba67-c295084f1962"),
+    )
+  }
 
   @State("There is an existing draft referral with ID of d496e4a7-7cc1-44ea-ba67-c295084f1962, and it has had a service provider selected")
-  fun `use referral d496e4a7 from the seed, since it also has a service provider`() {}
+  fun `create a new draft referral with HARMONY_LIVING service provider (default)`() {
```
rgforsyth, line 92: It looks like you combined two states to a function above(Line 66). As this function is doing the exact same as the state above, should you do the same thing here? Or is it separate for a reason.

rgforsyth (approved):

PR #92, opened 2021-02-08
IC-622: add openapi autogenerated docs and swagger ui

tomsmyers:

add openapi autogenerated docs and swagger ui

make the API and DTO schemas visible for interested parties,

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -35,6 +35,10 @@ tasks {
 }
 
 dependencies {
+  // openapi
+  implementation("org.springdoc:springdoc-openapi-ui:1.5.3")
+
+  // notifications
   implementation("uk.gov.service.notify:notifications-java-client:3.17.0-RELEASE")
   implementation("software.amazon.awssdk:sns:2.15.71")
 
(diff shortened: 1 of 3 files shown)


build.gradle.kts
```diff
@@ -35,6 +35,10 @@ tasks {
 }
 
 dependencies {
+  // openapi
+  implementation("org.springdoc:springdoc-openapi-ui:1.5.2")
```
sldblog, line 39: _nitpick_
🙋‍♂️ There seems to be a newer patch version that was released recently
```suggestion
  implementation("org.springdoc:springdoc-openapi-ui:1.5.3")
```

sldblog (approved):
😍 Love it

sldblog (approved):
Just occurred to me:

Could we add an API docs badge, similar to the one in [hmpps-auth](https://github.com/ministryofjustice/hmpps-auth/blob/main/README.md), please?

That way the API would get auto-published to the [Published APIs page](https://structurizr.com/share/56937/documentation#%2F:Published%20APIs) when the [arch-as-code](https://github.com/ministryofjustice/hmpps-architecture-as-code repo) gets built

PR #95, opened 2021-02-10
change config variables from camel casing to dashed case to work with…

rgforsyth:
… helm

Changes the config to dashed casing rather than camel casing.

Solve errors in the deployment to the dev region.


sldblog (commented):
Interesting, how does this work? Will camel casing get converted to underscored env names?

rgforsyth:
I tried your suggestion of running from the command line and passing the variables in that way. I replicated the error we see in the dev logs. Changing from camel casing to dashed typing locally allowed me to start up the application successfully.

tomsmyers (approved):

tomsmyers:
not sure why, but it's documented here 

https://docs.spring.io/spring-boot/docs/current/reference/html/spring-boot-features.html#boot-features-external-config-vs-value

PR #99, opened 2021-02-15
Filter by location

rgforsyth:

Allows the option query parameter 'pccRegionIds' to be passed to the Get: 'Interventions' endpoint. Doing this allows the api to return a filtered list of interventions, whose pcc region only contain those passed in, or whose nps region contains a pcc region that does.

To provide a way of returning a filtered list of interventions by pcc region as per IC-1060


tomsmyers (commented):
this is looking good to me. how come there are a few commented out tests?

tomsmyers:
please include a JIRA reference in the PR description :)

rgforsyth:
Some tests were left commented until your pr goes in, as it would be easier to set up the test data using the Intervention factories you have created.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/InterventionService.kt
```diff
@@ -13,20 +13,27 @@ import java.util.UUID
...
   fun getAllInterventions(): List<InterventionDTO> {
-    return repository.findAll().map {
+    return interventionRepository.findAll().map {
       InterventionDTO.from(it, getPCCRegions(it))
     }
   }
+
+  fun getInterventions(pccRegionIds: List<String>): List<InterventionDTO> {
+    return interventionRepository.findByCriteria(pccRegionIds).map {
+      InterventionDTO.from(it, getPCCRegions(it))
```
tomsmyers, line 33: @mahalrmoj suggested we return entities from the service layer and convert to DTOs in the controller. we can do outside this PR, so just a note.

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/util/DynamicFrameworkContractFactory.kt
```diff
@@ -0,0 +1,47 @@
...
+import java.util.UUID
+
+class DynamicFrameworkContractFactory(em: TestEntityManager) : EntityFactory(em) {
+  private val serviceCategoryFactory = ServiceCategoryFactory(em)
+  private val serviceProviderFactory = ServiceProviderFactory(em)
+  private val npsRegionFactory = NPSRegionFactory(em)
+  private val pccRegionFactory = PCCRegionFactory(em)
+
+  fun create(
+    id: UUID? = UUID.randomUUID(),
```
tomsmyers, line 19: no need for nullable id here

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/util/DynamicFrameworkContractFactory.kt
```diff
@@ -0,0 +1,47 @@
...
+    minimumAge: Int = 18,
+    maximumAge: Int? = null,
+    allowsMale: Boolean = true,
+    allowsFemale: Boolean = true,
+    npsRegion: NPSRegion? = null,
+    pccRegion: PCCRegion? = null,
+  ): DynamicFrameworkContract {
+    return save(
+      DynamicFrameworkContract(
+        id = id ?: UUID.randomUUID(),
```
tomsmyers, line 33: no need to specify a fallback here if we remove the nullable modifier above

tomsmyers (approved):
lgtm 

question - what does the sql look like that the filter generates? is it sane?

PR #101, opened 2021-02-15
WIP refactor intervention regions

tomsmyers:

_Required._

_Required._

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/jpa/entity/DynamicFrameworkContract.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/jpa/entity/DynamicFrameworkContract.kt
@@ -5,6 +5,7 @@ import java.util.UUID
 import javax.persistence.Entity
 import javax.persistence.Id
 import javax.persistence.JoinColumn
+import javax.persistence.ManyToMany
 import javax.persistence.ManyToOne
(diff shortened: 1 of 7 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/jpa/entity/DynamicFrameworkContract.kt
```diff
@@ -21,11 +22,12 @@ data class DynamicFrameworkContract(
...
   @NotNull val endDate: LocalDate,
 
-  @ManyToOne @JoinColumn(name = "nps_region_id")
-  val npsRegion: NPSRegion? = null,
-
-  @ManyToOne @JoinColumn(name = "pcc_region_id")
-  val pccRegion: PCCRegion? = null,
+  // a contract is offered for a single NPS region,
+  // or one-or-more PCC regions within an NPS region.
+  // if `npsRegion` is set, `pccRegions` should contain
```
mahalrmoj, line 27: Can we make the model more explicit in that we make both fields mandatory, so that a npsRegion always applies, and the pccRegion list contains all or some of the pccRegions. Where intervention applies to all pccRegions within the npsRegion, then all are present, otherwise only the subset that applies are present? Apologies if I've missed a constraint.

mahalrmoj (commented):
Like the movement of the DTOs into the controller.

PR #102, opened 2021-02-15
add ability to filter Get /interventions by gender. AllowsFemale & Al…

rgforsyth:
…lowsMale

Provides the ability to pass 'allowsMale' & 'allowsFemale' boolean parameters into the GET /interventions endpoint. This will allows the search results to filter on these parameters. This will also work in combination with a location filter.

To allow the filtering of intervention search results, as per IC-1026

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/jpa/repository/InterventionFilterRepositoryImpl.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/jpa/repository/InterventionFilterRepositoryImpl.kt
@@ -1,5 +1,6 @@
 package uk.gov.justice.digital.hmpps.hmppsinterventionsservice.jpa.repository
 
+import uk.gov.justice.digital.hmpps.hmppsinterventionsservice.jpa.entity.DynamicFrameworkContract
 import uk.gov.justice.digital.hmpps.hmppsinterventionsservice.jpa.entity.Intervention
 import uk.gov.justice.digital.hmpps.hmppsinterventionsservice.jpa.entity.NPSRegion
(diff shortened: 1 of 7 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/jpa/repository/InterventionFilterRepositoryImpl.kt
```diff
@@ -61,13 +62,18 @@ class InterventionFilterRepositoryImpl(
...
   }
 
-  private fun getGenderPredicate(criteriaBuilder: CriteriaBuilder?, root: Root<Intervention>?, gender: String?): Predicate? {
-
-    if (gender.isNullOrEmpty()) {
+  private fun getAllowsFemalePredicate(criteriaBuilder: CriteriaBuilder, root: Root<Intervention>, allowsFemale: Boolean?): Predicate? {
+    if (allowsFemale == null) {
       return null
     }
+    return criteriaBuilder.equal(root.get<Any>("dynamicFrameworkContract")?.get<Any>("allowsFemale"), allowsFemale)
```
tomsmyers, line 69: do the types matter here? in the other predicate we specify some of the types in the calls to `get`

tomsmyers (approved):
nice. the tests are easy to read too which is great.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/jpa/repository/InterventionFilterRepositoryImpl.kt
rgforsyth: Good spot. I don't think it actually affected anything, but i've updated all of them in this file to be more specific.

PR #107, opened 2021-02-18
IC-1113 only return sent referrals for the SP organization that the requesting user belongs to

tomsmyers:
note: this was originally done in the front end, but see the comments here about why this PR has been raised instead. https://github.com/ministryofjustice/hmpps-interventions-ui/pull/126

only return sent referrals for the SP organization that the requesting user belongs to

ensure SPs don't see referrals sent to other orgs

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/HMPPSAuthService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/HMPPSAuthService.kt
@@ -1,0 +1,61 @@
+package uk.gov.justice.digital.hmpps.hmppsinterventionsservice.service
+
+import org.slf4j.LoggerFactory
+import org.springframework.beans.factory.annotation.Value
+import org.springframework.http.HttpStatus
+import org.springframework.stereotype.Service
+import org.springframework.web.reactive.function.client.WebClient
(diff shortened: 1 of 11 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/HMPPSAuthService.kt
```diff
@@ -0,0 +1,61 @@
...
+            // we expect 404s for non-auth users
+            if (!e.statusCode.equals(HttpStatus.NOT_FOUND)) {
+              log.error("could not get groups for user", e)
+            }
+          }
+          else -> log.error("could not get groups for user", e)
+        }
+        Flux.empty()
+      }
+      .collectList().block()
```
rgforsyth, line 46: Is there a reason we are blocking here rather than using subscribe?

rgforsyth (approved):

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/HMPPSAuthService.kt
tomsmyers: yeh, we need the result right away

sldblog (commented):
Great stuff, obvs late review, just some musing

PR #108, opened 2021-02-18
create referral assignment event & send sns upon assignment event

rgforsyth:

Allows an event to be published upon a referral assignment, as per IC-1182

Send an sns message with referral assignment details when a referral assignment event is received


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/SNSService.kt
```diff
@@ -29,6 +29,16 @@ class SNSService(
         )
         publish(objectMapper.writeValueAsString(message))
       }
+      ReferralEventType.ASSIGNED -> {
+        val message = mapOf(
+          "eventType" to "intervention.referral.assigned",
+          "description" to "A referral has been assigned to a service user",
+//          "assigned_to" to event.referral.assignedTo,
```
rgforsyth, line 36: Exact values & callback url will be added once referral assignment code is merged

tomsmyers:
looking good - thanks !

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/component/SNSPublisher.kt
```diff
@@ -0,0 +1,46 @@
...
+@Component
+class SNSPublisher(
+  private val client: SnsClient,
+  private val objectMapper: ObjectMapper,
+  @Value("\${aws.sns.enabled}") private val enabled: Boolean,
+  @Value("\${aws.sns.topic.arn}") private val arn: String,
+) {
+  fun publish(event: EventDTO) {
+    if (enabled) {
+      return buildRequestAndPublish(event)
```
mahalrmoj, line 21: Should we log that it wasn't enabled just to help in debugging going forward?
Don't think return is necessary here

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/dto/EventDTO.kt
```diff
@@ -0,0 +1,12 @@
...
+import java.time.OffsetDateTime
+
+data class EventDTO(
+  val eventType: String,
+  val description: String,
+  val detailUrl: String,
+  val occurredAt: OffsetDateTime,
+  val version: Int,
+  val additionalInformation: Map<String, Any>
+)
```
mahalrmoj, line 12: Should additional be defaulted to an empty list?
Any looks a bit strange but can't think of anything better

mahalrmoj (commented):
Nice to see the actual publisher concern/details move out of the service into its own class
.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/component/SNSPublisher.kt
```diff
@@ -0,0 +1,46 @@
...
+    try {
+      val result = client.publish(request)
+      log.debug("message sent; messageId: ${result.messageId()}, status: ${result.sdkHttpResponse().statusCode()}")
+    } catch (e: Exception) {
+      log.error("message send failure", e)
+    }
+  }
+
+  companion object {
+    private val log = LoggerFactory.getLogger(SNSService::class.java)
```
tomsmyers, line 44: logger name is incorrect

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/component/SNSPublisher.kt
tomsmyers: i'd be happy with a log event at DEBUG level if the publisher is disabled. any higher level i think would be too much noise.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/SNSService.kt
```diff
@@ -1,56 +1,52 @@
...
-      log.debug("message sent; messageId: ${result.messageId()}, status: ${result.sdkHttpResponse().statusCode()}")
-    } catch (e: Exception) {
-      log.error("message send failure", e)
-    }
-  }
-
-  companion object {
-    private val log = LoggerFactory.getLogger(SNSService::class.java)
+  private fun createDetailUrl(event: ReferralEvent): String {
+    return UriComponentsBuilder.fromHttpUrl(interventionsServiceBaseURL)
```
tomsmyers, line 47: you can get this from `ServletUriComponentsBuilder.fromCurrentContextPath()` without having to get it from config

src/main/resources/application-local.yml
```diff
@@ -4,6 +4,11 @@ notify:
 interventions-ui:
   baseurl: http://localhost:3000
 
+interventions-service:
+  baseurl: http://localhost:8080
+  locations:
+    sent-referral: "/sent-referral/{id}"
```
tomsmyers, line 10: i know this follows the pattern here, but i'm not sure having internal urls in config is a good idea. 

these URLs are defined in the controllers, and it would be very easy to forget to update the URL here too. is there some method we can use to get the url by controller method name or something so that it would either be correct or cause a compile time error?

tomsmyers (commented):
couple of changes requested

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/ReferralService.kt
```diff
@@ -31,7 +31,9 @@ class ReferralService(
     referral.assignedAt = OffsetDateTime.now()
     referral.assignedBy = authUserRepository.save(assignedBy)
     referral.assignedTo = authUserRepository.save(assignedTo)
-    return referralRepository.save(referral)
+    val assignedReferral = referralRepository.save(referral)
+    eventPublisher.referralSentEvent(assignedReferral)
```
tomsmyers, line 35: this is the wrong event ! :)

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/dto/EventDTO.kt
```diff
@@ -0,0 +1,13 @@
...
+import java.time.OffsetDateTime
+
+data class EventDTO(
+  val eventType: String,
+  val description: String,
+  val detailUrl: String,
+  val occurredAt: OffsetDateTime,
+  val additionalInformation: Map<String, Any>
+) {
+  val version: Int = 1
```
rgforsyth, line 12: This is here to be hard coded into the DTO. It is not up to the caller to decide the version

tomsmyers (approved):
all good - thanks! 

i would like a unit test for the 'detail url' generation method - but not a blocker

PR #113, opened 2021-02-19
IC-1142 Allow referrals to be assigned

tomsmyers:

Adds service methods and endpoints for allowing sent referrals to be assigned to a SP user.

please review commit by commit as there are some changes in here with associated refactoring or test sample data updates.

Critical outcome in which a referral can be assigned


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/controller/ReferralController.kt
```diff
@@ -31,12 +32,30 @@ class ReferralController(
...
-    val uuid = parseID(id)
+  @PostMapping("/sent-referral/{id}/assign")
+  fun assignSentReferral(
+    @PathVariable id: UUID,
+    @RequestBody referralAssignment: ReferralAssignmentDTO,
+    authentication: JwtAuthenticationToken,
+  ): SentReferralDTO {
+    val sentReferral = referralService.getSentReferral(id)
+      ?: throw ResponseStatusException(HttpStatus.NOT_FOUND, "sent referral not found [id=$id]")
+
```
mahalrmoj, line 43: The retrieval of the existing referral can confined to the service tier. Can the not found exception be raised by the service tier and converted in the controller to a responseStatusException.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/dto/AuthUserDTO.kt
```diff
@@ -5,12 +5,14 @@ import uk.gov.justice.digital.hmpps.hmppsinterventionsservice.jpa.entity.AuthUse
 class AuthUserDTO(
```
mahalrmoj, line 5: Should this be a data class?

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/controller/ReferralController.kt
```diff
@@ -120,20 +134,10 @@ class ReferralController(
...
-    val uuid = parseID(id)
-
-    return serviceCategoryService.getServiceCategoryByID(uuid)
+  fun getServiceCategoryByID(@PathVariable id: UUID): ServiceCategoryDTO {
+    return serviceCategoryService.getServiceCategoryByID(id)
       ?.let { ServiceCategoryDTO.from(it) }
-      ?: throw ResponseStatusException(HttpStatus.NOT_FOUND, "service category not found [id=$uuid]")
-  }
-
-  private fun parseID(id: String): UUID {
```
mahalrmoj, line 131: Just FYI I've moved this to function so it can be reused by all controllers

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/jpa/entity/Referral.kt
```diff
@@ -21,6 +21,12 @@ import javax.validation.constraints.NotNull
 @Entity
 @Table(indexes = arrayOf(Index(columnList = "created_by_id")))
 data class Referral(
+  // assigned referral fields
+  var assignedAt: OffsetDateTime? = null,
+  @ManyToOne(fetch = FetchType.LAZY) var assignedBy: AuthUser? = null,
```
mahalrmoj, line 26: Should we make these eager and they are used most of the time by the DTOs?
@Fetch(value = FetchMode.JOIN)

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/ReferralService.kt
```diff
@@ -25,6 +25,13 @@ class ReferralService(
   val interventionRepository: InterventionRepository,
   val eventPublisher: ReferralEventPublisher
 ) {
+  fun assignSentReferral(referral: Referral, assignedBy: AuthUser, assignedTo: AuthUser): Referral {
```
mahalrmoj, line 28: Remove referral as a supplied parameter and retrieve as part of method and through exception if entity not found

mahalrmoj:
What should happen if a plan already exists for a referral? It wil fail

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/controller/ReferralController.kt
tomsmyers: as in the service method takes an id and throws an error if it doesn't exist?

yeh i'm not against it. if we decide to do it let's do it in another commit though and change all at once.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/dto/AuthUserDTO.kt
tomsmyers: yes good spot - saw you handled this so not changing here.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/jpa/entity/Referral.kt
tomsmyers: done

mahalrmoj (approved):
Thanks

PR #118, opened 2021-02-24
IC-1243 email notification on assigned referrals

tomsmyers:

send an email to the assignee when a referral is assigned.

alert the service provider case worker that a referral has been assigned to them

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/CommunityAPIService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/CommunityAPIService.kt
@@ -55,6 +55,7 @@ class CommunityAPIService(
           }
           .subscribe()
       }
+      else -> {}
     }
   }
 
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/HMPPSAuthService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/HMPPSAuthService.kt
@@ -2,22 +2,48 @@ package uk.gov.justice.digital.hmpps.hmppsinterventionsservice.service
 
 import org.slf4j.LoggerFactory
 import org.springframework.beans.factory.annotation.Value
+import org.springframework.http.HttpStatus
 import org.springframework.stereotype.Service
 import org.springframework.web.reactive.function.client.WebClient
 import org.springframework.web.util.UriComponentsBuilder
+import reactor.core.publisher.Mono
 import uk.gov.justice.digital.hmpps.hmppsinterventionsservice.jpa.entity.AuthGroupID
 import uk.gov.justice.digital.hmpps.hmppsinterventionsservice.jpa.entity.AuthUser
 
-val AuthServiceProviderGroupPrefix = "INT_SP_"
+class UnverifiedEmailException : RuntimeException()
 
-data class AuthGroup(
+data class UserDetail(
+  val firstName: String,
+  val email: String,
+)
+
+private const val AuthServiceProviderGroupPrefix = "INT_SP_"
+
+private data class AuthGroupResponse(
   val groupCode: String,
   val groupName: String,
 )
 
+private data class AuthUserDetailResponse(
+  val firstName: String,
+  val email: String,
+  val verified: Boolean,
+)
+
+private data class UserEmailResponse(
+  val email: String,
+)
+
+private data class UserDetailResponse(
+  val name: String,
+)
+
 @Service
 class HMPPSAuthService(
-  @Value("\${hmppsauth.api.locations.user-groups}") private val userGroupsLocation: String,
+  @Value("\${hmppsauth.api.locations.auth-user-groups}") private val authUserGroupsLocation: String,
+  @Value("\${hmppsauth.api.locations.auth-user-detail}") private val authUserDetailLocation: String,
+  @Value("\${hmppsauth.api.locations.user-email}") private val userEmailLocation: String,
+  @Value("\${hmppsauth.api.locations.user-detail}") private val userDetailLocation: String,
   private val hmppsAuthApiWebClient: WebClient,
 ) {
   fun getServiceProviderOrganizationForUser(user: AuthUser): AuthGroupID? {
@@ -25,13 +51,13 @@ class HMPPSAuthService(
       return null
     }
 
-    val url = UriComponentsBuilder.fromPath(userGroupsLocation)
+    val url = UriComponentsBuilder.fromPath(authUserGroupsLocation)
       .buildAndExpand(user.userName)
       .toString()
 
     val groups = hmppsAuthApiWebClient.get().uri(url)
       .retrieve()
-      .bodyToFlux(AuthGroup::class.java)
+      .bodyToFlux(AuthGroupResponse::class.java)
       .collectList().block()
 
     val serviceProviderOrgs = groups
@@ -42,6 +68,38 @@ class HMPPSAuthService(
     // which represent SP admins managing subcontractor organizations.
     // for now we can assume there is only one group per user.
     return if (serviceProviderOrgs.isEmpty()) null else serviceProviderOrgs[0]
+  }
+
+  fun getUserDetail(user: AuthUser): UserDetail {
+    return if (user.authSource == "auth") {
+      val url = UriComponentsBuilder.fromPath(authUserDetailLocation).buildAndExpand(user.userName).toString()
+      hmppsAuthApiWebClient.get().uri(url)
+        .retrieve()
+        .bodyToMono(AuthUserDetailResponse::class.java)
+        .map {
+          if (!it.verified) {
+            throw UnverifiedEmailException()
+          }
+          UserDetail(it.firstName, it.email)
+        }
+        .block()
+    } else {
+      val detailUrl = UriComponentsBuilder.fromPath(userDetailLocation).buildAndExpand(user.userName).toString()
+      val emailUrl = UriComponentsBuilder.fromPath(userEmailLocation).buildAndExpand(user.userName).toString()
+      Mono.zip(
+        hmppsAuthApiWebClient.get().uri(detailUrl)
+          .retrieve()
+          .bodyToMono(UserDetailResponse::class.java)
+          .map { it.name.substringBefore(' ') },
+        hmppsAuthApiWebClient.get().uri(emailUrl)
+          .retrieve()
+          .onStatus({ it.equals(HttpStatus.NO_CONTENT) }, { Mono.error(UnverifiedEmailException()) })
+          .bodyToMono(UserEmailResponse::class.java)
+          .map { it.email }
+      )
+        .map { UserDetail(it.t1, it.t2) }
+        .block()
+    }
   }
 
   companion object {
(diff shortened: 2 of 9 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/HMPPSAuthService.kt
```diff
@@ -44,6 +70,38 @@ class HMPPSAuthService(
...
+        }
+        .block()
+    } else {
+      val detailUrl = UriComponentsBuilder.fromPath(userDetailLocation).buildAndExpand(user.userName).toString()
+      val emailUrl = UriComponentsBuilder.fromPath(userEmailLocation).buildAndExpand(user.userName).toString()
+      Mono.zip(
+        hmppsAuthApiWebClient.get().uri(detailUrl)
+          .retrieve()
+          .bodyToMono(UserDetailResponse::class.java)
+          .map { it.name.substringBefore(' ') },
```
tomsmyers, line 93: this is a sucky way to get the firstname, but i don't know what else to do

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/HMPPSAuthService.kt
```diff
@@ -2,36 +2,62 @@ package uk.gov.justice.digital.hmpps.hmppsinterventionsservice.service
...
+  val name: String,
+)
+
 @Service
 class HMPPSAuthService(
-  @Value("\${hmppsauth.api.locations.user-groups}") private val userGroupsLocation: String,
+  @Value("\${hmppsauth.api.locations.auth-user-groups}") private val authUserGroupsLocation: String,
+  @Value("\${hmppsauth.api.locations.auth-user-detail}") private val authUserDetailLocation: String,
+  @Value("\${hmppsauth.api.locations.user-email}") private val userEmailLocation: String,
+  @Value("\${hmppsauth.api.locations.user-detail}") private val userDetailLocation: String,
```
sldblog, line 46: _not a blocker_

🙋‍♂️ Is there a need to have these locations as configuration? I'm assuming they won't change between environments, we don't need to use them elsewhere and if they change in _hmpps-auth_, we most likely to have other code changes to do as well.

There is probably an additional benefit with using the 2nd option: the reader does not have to look up a value from the config to understand what it's doing:

```kt
// 1
val url = UriComponentsBuilder.fromPath(authUserGroupsLocation)
  .buildAndExpand(user.userName)

// 2
val url = UriComponentsBuilder.fromPath("/api/authuser/{username}/groups")
  .buildAndExpand(user.userName)
```

What do you think?

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/HMPPSAuthService.kt
```diff
@@ -44,6 +70,38 @@ class HMPPSAuthService(
...
 
+  fun getUserDetail(user: AuthUser): UserDetail {
+    return if (user.authSource == "auth") {
+      val url = UriComponentsBuilder.fromPath(authUserDetailLocation).buildAndExpand(user.userName).toString()
+      hmppsAuthApiWebClient.get().uri(url)
+        .retrieve()
+        .bodyToMono(AuthUserDetailResponse::class.java)
+        .map {
+          if (!it.verified) {
+            throw UnverifiedEmailException()
```
sldblog, line 81: Could we validate that the person has a verified e-mail address in other control flows? This (`getUserDetail`) seems like an unusual place to do it

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/HMPPSAuthService.kt
sldblog: could we poll the user details from its source when we first validate a token like this?
i.e. `AuthUser` &rarr; pull details from source? `userDetailsFromAuth`, `userDetailsFromDelius`?

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/NotifyService.kt
```diff
@@ -8,32 +8,63 @@ import org.springframework.web.util.UriComponentsBuilder
...
 class NotifyService(
   @Value("\${notify.enabled}") private val enabled: Boolean,
   @Value("\${notify.templates.referral-sent}") private val referralSentTemplateID: String,
+  @Value("\${notify.templates.referral-assigned}") private val referralAssignedTemplateID: String,
   @Value("\${interventions-ui.baseurl}") private val interventionsUIBaseURL: String,
   @Value("\${interventions-ui.locations.sent-referral}") private val interventionsUISentReferralLocation: String,
   private val client: NotificationClient,
+  private val hmppsAuthService: HMPPSAuthService,
 ) : ApplicationListener<ReferralEvent> {
   override fun onApplicationEvent(event: ReferralEvent) {
```
sldblog, line 24: _observation_
🤔 this method would now probably trip a cyclomatic complexity or function lines-of-code static check

do we want those to run on the code? (I'm not suggesting to rewrite, just interested in if we want to setup something like that)

sldblog (commented):
👍 Just a couple of questions

The biggest one I suppose is around verifying emails, doesn't sound like we should allow those users to even access the service, which could simplify the code we have here

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/HMPPSAuthService.kt
tomsmyers: yeh i kind of agree. originally i liked having all the URLs in config because they kinda feel like config to me. but i agree that it's easier to read them in place and you're less likely to use the wrong one. 

can i put up a seperate PR to change all the configured 'locations' in this way (there are some for community api too)?

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/HMPPSAuthService.kt
tomsmyers: i don't fully understand the 403 idea. but the thinking here is that the caller of this method wants the user details, including the email address. if the email address is unverified, then the method blows up because the email address isn't reliable. so it seems like an ok place to check this to me.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/NotifyService.kt
tomsmyers: good point. yes i'd be up for running something like this. rajiv is adding to this class as we speak so he could include some refactoring there? @mahalrmoj

tomsmyers:
@sldblog i can't reply to this comment for some reason ... here is the reply

https://github.com/ministryofjustice/hmpps-interventions-service/pull/118/#discussion_r582068711

at this point we have no token for the user or anything like it. this user was passed to us in the body of the 'assign' request. so we just have username, userid and authsource. we could use community api to get the user details, but i'm not sure it's worth it. the first name is only used in the email to say "hi {firstName}", so it's not super important.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/HMPPSAuthService.kt
tomsmyers: at this point we have no token for the user or anything like it. this user was passed to us in the body of the 'assign' request. so we just have username, userid and authsource. we could use community api to get the user details, but i'm not sure it's worth it. the first name is only used in the email to say "hi {firstName}", so it's not super important.

tomsmyers: a downside of doing this is that community api doesn't store email address for user. so to get the information we require to send the email we would have to make a call to hmpps auth for the email and delius for the name. which is added complexity.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/HMPPSAuthService.kt
tomsmyers: actually - i'll just change it in this PR now

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/HMPPSAuthService.kt
sldblog: Sure, sorry, I meant: should we explicitly say no one with an unverified email can access or be assigned in our service?

tomsmyers: ok i see what you mean.

first off this is just a method to get user details for a username. so i don't think we can make any assumptions about whether the username has a verified email in this class. 

second, we will do the validation you are ttalking about on the front end anyway. we could do it in the API as well, but i still don't think it changes the fact that this method should be agnostic and handle unverified emails. it just means these UnverifiedEmailExceptions should never be thrown :) !

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/HMPPSAuthService.kt
tomsmyers: so as i've gone through making this change i've identified a couple of positives to having these URLs in config.

1. they are reusable. in one instance the value was used in multiple places
2. allows injection of the url by test classes which makes using mock webservers more reliable in tests.

considering these - i'm going to leave this change as it is for now and we can consider these pros and cons separately.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/CommunityAPIService.kt
```diff
@@ -55,6 +55,7 @@ class CommunityAPIService(
           }
           .subscribe()
       }
+      else -> {}
```
rgforsyth, line 58: Is this serving any purpose?

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/HMPPSAuthService.kt
```diff
@@ -2,36 +2,62 @@ package uk.gov.justice.digital.hmpps.hmppsinterventionsservice.service
...
       .buildAndExpand(user.userName)
       .toString()
 
     val groups = hmppsAuthApiWebClient.get().uri(url)
       .retrieve()
-      .bodyToFlux(AuthGroup::class.java)
+      .bodyToFlux(AuthGroupResponse::class.java)
       .collectList().block()
 
     val serviceProviderOrgs = groups
```
rgforsyth, line 63: not something you have changed, but unless we are wanting to explicitly state that a service provider organisation is a group (which we may?), can this not just be assigned to the above request?

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/CommunityAPIService.kt
tomsmyers: oh it's a compiler warning i guess i fixed on a whim. not really is the answer

rgforsyth (approved):
LGTM - Shame that Auth doesn't return the name as first/last for non-auth users.

sldblog (approved):

sldblog:
> Shame that Auth doesn't return the name as first/last for non-auth users.

Could we raise a PR for it there?

sldblog:
`ci/circleci: build_docker` will be solved by upgrading the hmpps orb to 2.3.2

PR #122, opened 2021-02-26
IC-1172 add application instrumentation key to helm files

rgforsyth:

Add the application instrumentation key/secret.

Allow exceptions/logs to be sent to application insights


helm_deploy/hmpps-interventions-service/templates/_envs.tpl
```diff
@@ -34,6 +34,12 @@ env:
   - name: COMMUNITYAPI_CONTACTNOTIFICATIONCONTEXT_TEAMCODE
     value: "{{ .Values.env.COMMUNITYAPI_CONTACTNOTIFICATIONCONTEXT_TEAMCODE }}"
 
+  - name: APPLICATION_INSIGHTS_IKEY
+    valueFrom:
+      secretKeyRef:
+        name: {{ template "app.name" . }}
```
sldblog, line 40: This would look up the secret from `hmpps-interventions-service` secret, is that correct?

helm_deploy/secrets-example.yaml
```diff
@@ -2,3 +2,4 @@
 secrets:
   INTERVENTIONS_SERVICE_CLIENT_ID: clientid
   INTERVENTIONS_SERVICE_CLIENT_SECRET: clientsecret
+  APPINSIGHTS_INSTRUMENTATIONKEY: somesecretvalue
```
sldblog, line 5: The helm secret definition creates an environment variable called `APPLICATION_INSIGHTS_IKEY`, but this one is called `APPINSIGHTS_INSTRUMENTATIONKEY` -- which one should be used?

sldblog (commented):
Nice, thanks! Just some clarification questions

helm_deploy/secrets-example.yaml
rgforsyth: Nice catch. I initially just reimplemented the code that was removed a while ago.
I've updated to match the variable within _envs.tpl

helm_deploy/hmpps-interventions-service/templates/_envs.tpl
```diff
@@ -34,6 +34,12 @@ env:
...
+        key: APPINSIGHTS_INSTRUMENTATIONKEY
```
sldblog, line 41: 🙋‍♂️ we essentially have key/value pairs namespaced under the secret name, so if you'd like we can call it

[secret name] `application-insights` &rarr; [key inside secret] `instrumentation-key`
vs current
[secret name] `application-insights-key` &rarr; [key inside secret] `APPINSIGHTS_INSTRUMENTATIONKEY `

absolutely not a blocker, just an opportunity to lean on the namespace/key/value semantics

helm_deploy/secrets-example.yaml
```diff
@@ -2,3 +2,4 @@
 secrets:
   INTERVENTIONS_SERVICE_CLIENT_ID: clientid
```
sldblog, line 3: Totally not related: `SPRING_SECURITY_OAUTH2_CLIENT_REGISTRATION_INTERVENTIONSCLIENT_CLIENTID` is the one we may need, I guess? 🤷‍♂️

sldblog (approved):
LGTM 👍

PR #126, opened 2021-03-02
fix sent referral URL which is sent in email notifications

tomsmyers:

fix sent referral URL which is sent in email notifications

point the recipient of the email to the correct page in the UI

### DIFF
--- a/src/main/resources/application.yml
+++ b/src/main/resources/application.yml
@@ -76,7 +76,7 @@ notify:
 
 interventions-ui:
   locations:
-    sent-referral: "/referrals/{id}/confirmation"
+    sent-referral: "/service-provider/referrals/{id}"
 
 community-api:
   locations:


mahalrmoj (approved):

Gweaton (approved):
Looks good to me.

Once we have the action plan routing in place it might be a bit strange as users will still be able to visit this page via the link in the email notification, but not via any links in the UI, so we'll probably want to have a chat about what to do when that's in place in future.

src/main/resources/application.yml
```diff
@@ -76,7 +76,7 @@ notify:
 
 interventions-ui:
   locations:
-    sent-referral: "/referrals/{id}/confirmation"
+    sent-referral: "/service-provider/referrals/{id}"
```
sldblog, line 79: Sorry, getting out of touch again. Is this page only for providers?

sldblog: I guess what I'm pondering: should we have `locations.probation.sent-referral`, `locations.providers.sent-referral`?

sldblog: (Another not-really-relevant-right-now thought: this sounds like shared mapping; we could maintain a mapping in Kubernetes/Redis, seeded by the UI app, what the service could query)

tomsmyers: thanks for comments. yeh right now the only 'referral detail' page is targeted for providers. as a PP if you head to this page you see the 'assign to case worker' form. but the existing confirmation page doesn't give you any information about the referral at all. this is something we should loop around on and see if we need to make the referral landing page for PPs at this point too. maybe this will fall out of the action plan approval/EOSR stuff next sprint.

sldblog: thanks for explaining!

PR #128, opened 2021-03-03
IC-1219 IC-1230 Update plan activities incrementally and Contract Tests

mahalrmoj:

IC-1230 Implementation of contract tests as expected by the UI
IC-1219 To make the update to a draft plan incremental, i.e. either the addition of a new activity or an update to the number of sessions.

To implement automated tests to verify that the api always conforms to the contract and can be run as part of the build.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/component/ActionPlanValidator.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/component/ActionPlanValidator.kt
@@ -15,7 +15,7 @@ class ActionPlanValidator {
       if (it <= 0) {
         errors.add(FieldError(field = "numberOfSessions", error = Code.CANNOT_BE_NEGATIVE_OR_ZERO))
       }
-    } ?: errors.add(FieldError(field = "numberOfSessions", error = Code.CANNOT_BE_EMPTY))
+    }
 
     if (errors.isNotEmpty()) {
       throw ValidationError("draft action plan update invalid", errors)
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/component/ActionPlanValidator.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/component/ActionPlanValidator.kt
@@ -1,7 +1,8 @@
 package uk.gov.justice.digital.hmpps.hmppsinterventionsservice.component
 
 import org.springframework.stereotype.Component
-import uk.gov.justice.digital.hmpps.hmppsinterventionsservice.config.Code
+import uk.gov.justice.digital.hmpps.hmppsinterventionsservice.config.Code.CANNOT_BE_EMPTY
+import uk.gov.justice.digital.hmpps.hmppsinterventionsservice.config.Code.CANNOT_BE_NEGATIVE_OR_ZERO
 import uk.gov.justice.digital.hmpps.hmppsinterventionsservice.config.FieldError
 import uk.gov.justice.digital.hmpps.hmppsinterventionsservice.config.ValidationError
 import uk.gov.justice.digital.hmpps.hmppsinterventionsservice.jpa.entity.ActionPlan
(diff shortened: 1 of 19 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/dto/UpdateActionPlanDTO.kt
```diff
@@ -0,0 +1,6 @@
+package uk.gov.justice.digital.hmpps.hmppsinterventionsservice.dto
+
+data class UpdateActionPlanDTO(
+  val numberOfSessions: Int?,
```
tomsmyers, line 4: ah i see why this was removed from the validation logic. it's optional to allow updates to just include new activities.

can we change the validation logic then to say "if numberOfSessions is not already set, and the request does not contain number of sessions, then error"? or is that handled elsewhere in the 'create action plan' stuf?

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/jpa/entity/ActionPlanActivity.kt
```diff
@@ -13,8 +15,13 @@ data class ActionPlanActivity(
...
 
   // For ordering
-  @NotNull val createdAt: OffsetDateTime,
+  @NotNull val createdAt: OffsetDateTime = OffsetDateTime.now(),
 
   // required fields
   @NotNull @ManyToOne(fetch = FetchType.LAZY) val desiredOutcome: DesiredOutcome,
+
+  // as an embedded collection entry each instance have no unique identifier
+  // this uuid can be though of as a unique handle to each entry for future use
```
tomsmyers, line 24: thanks for this comment

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/ActionPlanService.kt
```diff
@@ -47,10 +47,14 @@ class ActionPlanService(
...
-    actionPlanValidator.validateDraftActionPlanUpdate(update)
-    updateDraftActivityPlan(draftActionPlan, update)
+  fun updateActionPlan(
+    actionPlanId: UUID,
+    numberOfSessions: Int?,
+    activityUpdate: ActionPlanActivity?,
+  ): ActionPlan {
+    val draftActionPlan = getDraftActionPlan(actionPlanId)
+    updateDraftActivityPlan(draftActionPlan, numberOfSessions, activityUpdate)
+    actionPlanValidator.validateDraftActionPlanUpdate(draftActionPlan)
```
tomsmyers, line 57: shouldn't we validate before updating? then if it's invalid we save a few cycles.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/ActionPlanService.kt
```diff
@@ -71,10 +75,18 @@ class ActionPlanService(
...
+  private fun updateDraftActivityPlan(
+    draftActionPlan: ActionPlan,
+    numberOfSessions: Int?,
+    activityUpdate: ActionPlanActivity?
+  ) {
+    numberOfSessions?.let {
       draftActionPlan.numberOfSessions = it
     }
+
+    activityUpdate?.let {
```
tomsmyers, line 87: is it possible to remove an activity using this api?

tomsmyers: i.e. is it clear in the contract that this is additive, as opposed to replacing all the activities on the action plan with content of the request body ?

tomsmyers: (this would affect the mutability of the list in the action plan entity, as changed above)

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/ActionPlanService.kt
mahalrmoj: The validator is an ActionPlan validator as apposed a request object validator which means it can only be done once the update has been applied to the entity object. I do take your point but the validator has greater reuse as an entity validator.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/component/ActionPlanValidator.kt
```diff
@@ -15,7 +15,7 @@ class ActionPlanValidator {
       if (it <= 0) {
         errors.add(FieldError(field = "numberOfSessions", error = Code.CANNOT_BE_NEGATIVE_OR_ZERO))
       }
-    } ?: errors.add(FieldError(field = "numberOfSessions", error = Code.CANNOT_BE_EMPTY))
```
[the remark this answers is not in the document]
mahalrmoj: This can only be checked at the point of submitting the action plan when it is fully entered as apposed to validation that can be done on incremental changes.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/dto/UpdateActionPlanDTO.kt
mahalrmoj: This can only be done when "submitting the action plan" because up until that point the plan remains incomplete and can be added too incrementally. 
At some point we may have a number of "rules" to ensure the validity of an action plan at the point of submission, e.g. an activity must not exist for an outcome that is not part of the referral. When we build this type of logic we can add the rule for ensuring a number of session exists.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/ActionPlanService.kt
mahalrmoj: Delete activity can't be specified using the current request object design. An update can be done but not implemented for skeleton. The request object reflects this.

data class UpdateActionPlanDTO(
  val numberOfSessions: Int?,
  val activity: CreateActionPlanActivityDTO?
)

However, I can make more clear by change the name of the variable derived from activity to newActivity rather than activityUpdate. This would give clarity. Like it.

mahalrmoj: Done

tomsmyers (approved):

PR #129, opened 2021-03-04
IC-653 Improve logging configuration for production environments.

tomsmyers:

Add logback configuration which splits the logging config into local/prod
sections. local config behaves the same as normal, coloured output pretty printed at DEBUG level.
the non locl config formats messages as json at WARN level.

structured log messages in production environments.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -38,6 +38,10 @@ tasks {
 }
 
 dependencies {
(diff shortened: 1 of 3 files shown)


mahalrmoj (approved):

sldblog:
🎉

PR #130, opened 2021-03-04
IC-1138: Expose our schema on /meta/schema URL

sldblog:

This change exposes our database schema on `/meta/schema` as a data dictionary.

This is intended as a plug-in enhancement:

1. It's an addiitional CircleCI job
2. The deployment is mostly an extension of the current deployment
3. The page can be plugged in via the Kubernetes ingress: no app changes needed

be8c31dab93cd06586ff4db22aaaf3b51e0c8a08 also provides the first documentation for the contract table.

We are gravitating towards "self-describing" services, which we can interrogate for metadata.

We can use our database schema for the "data dictionary" as the comments on the tables and columns, together with their types and names give us a reasonable starting ground.


sldblog (commented):
I see, this requires `graphviz` now to run the unit test suite. Let me have a think.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/config/ResourceServerConfiguration.kt
```diff
@@ -24,7 +24,7 @@ class ResourceServerConfiguration : WebSecurityConfigurerAdapter() {
       .sessionCreationPolicy(SessionCreationPolicy.STATELESS)
       .and().csrf().disable()
       .authorizeRequests {
-        it.antMatchers("/health/**", "/info", "/v3/api-docs/**", "/swagger-ui/**", "/swagger-ui.html").permitAll()
+        it.antMatchers("/health/**", "/meta/schema", "/info", "/v3/api-docs/**", "/swagger-ui/**", "/swagger-ui.html").permitAll()
```
tomsmyers, line 27: do we want this to be accessible to the public?

sldblog: I think so. It's the schema of the data, without the data itself. Should we restrict it?

tomsmyers: i guess our source code is public so it makes no difference 👍

sldblog:
I'm thinking about generating an HTML schema page as part of the build and bundling that in the Docker image. Would that be better?

sldblog:
@ministryofjustice/hmpps-interventions This is now ready for re-review 🙇‍♂️

.circleci/config.yml
```diff
@@ -44,6 +44,44 @@ jobs:
       - store_artifacts:
           path: build/reports/tests
 
+  publish_data:
+    executor: hmpps/java
+    docker:
+      - image: cimg/openjdk:11.0
```
tomsmyers, line 50: what's this image for?

.circleci/config.yml
```diff
@@ -44,6 +44,44 @@ jobs:
...
+      - image: circleci/postgres:10-alpine-ram
+        environment:
+          POSTGRES_PASSWORD: password
+    steps:
+      - checkout
+      - hmpps/create_app_version
+      - run:
+          name: Migrate the database
+          command: |
+            wget -qO- https://repo1.maven.org/maven2/org/flywaydb/flyway-commandline/7.7.0/flyway-commandline-7.7.0-linux-x64.tar.gz | \
```
tomsmyers, line 62: not a blocker: instead of doing this we could add the gradle plugin for flyway

.circleci/config.yml
```diff
@@ -44,6 +44,44 @@ jobs:
...
+              tar xvz && sudo ln -s `pwd`/flyway-7.7.0/flyway /usr/local/bin
+            flyway migrate -url="jdbc:postgresql://localhost:5432/$POSTGRES_DB" -user="postgres" -password="password" -locations="src/main/resources/db/migration"
+      - run:
+          name: Build data dictionary
+          command: |
+            sudo apt update && sudo apt install graphviz
+            script/build-data-dictionary.sh
+      - setup_remote_docker
```
tomsmyers, line 70: could we use the build-docker job provided by the hmpps orb for these subsequent tasks?

script/.dockerignore
```diff
@@ -0,0 +1 @@
+libs/
```
tomsmyers, line 1: somewhat confused by this directory being called 'script'

tomsmyers (approved):
great work. a few comments, no blockers.

.circleci/config.yml
sldblog: I needed a `java` executable to run the `schemaspy.jar`

script/.dockerignore
sldblog: would `bin` be better?

PR #133, opened 2021-03-05
IC-1358 Sent Referrals Contains Action Plan Id and new endpoint /acti…

mahalrmoj:
…on-plan?referralId

1) The existing endpoint /sent-referrals now includes an actionPlanId for each referral found should it be set.
2) An action plan can be retrieved using a referralId /action-plan?referralId= 

These additions facilitate the flow in the ui from a referral to an acto plan


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/controller/ReferralController.kt
```diff
@@ -79,7 +80,7 @@ class ReferralController(
   }
 
   @GetMapping("/sent-referrals")
-  fun getSentReferrals(authentication: JwtAuthenticationToken): List<SentReferralDTO> {
+  fun getSentReferrals(authentication: JwtAuthenticationToken): List<SentReferralWithActionPlanIdDTO> {
```
tomsmyers, line 83: what's the reason this endpoint and `/sent-referral/{id}` return different DTOs?

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/dto/SentReferralWithActionPlanIdDTO.kt
```diff
@@ -0,0 +1,29 @@
+package uk.gov.justice.digital.hmpps.hmppsinterventionsservice.dto
+
+import uk.gov.justice.digital.hmpps.hmppsinterventionsservice.jpa.entity.Referral
+import java.time.OffsetDateTime
+import java.util.UUID
+
+class SentReferralWithActionPlanIdDTO(
```
tomsmyers, line 7: why do we need this extra class? can't we just have the optional actionPlanId on the `SentReferralDTO`?

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/jpa/entity/ActionPlan.kt
```diff
@@ -35,4 +35,9 @@ data class ActionPlan(
   // Required
   @NotNull @OneToOne @Fetch(FetchMode.JOIN) val referral: Referral,
   @Id val id: UUID,
-)
+) {
+  // This is to avoid stack overflow issues with the bi-directional association with referral
+  override fun toString(): String {
+    return "ActionPlan(referralId=${referral.id}, numberOfSessions=$numberOfSessions, activities=$activities, createdBy=$createdBy, createdAt=$createdAt, submittedBy=$submittedBy, submittedAt=$submittedAt, id=$id)"
```
tomsmyers, line 41: i understand the reason this exists, but shouldn't it be JSON?

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/ReferralService.kt
```diff
@@ -45,9 +46,9 @@ class ReferralService(
     return referralRepository.findBySentAtIsNotNull().map { SentReferralDTO.from(it) }
   }
 
-  fun getSentReferralsForServiceProviderID(serviceProviderID: AuthGroupID): List<SentReferralDTO> {
```
tomsmyers, line 48: not a blocker because i see this is consistent with out methods in this class, but aren't we leaving the controllers to do all the DTO mappings?

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/controller/ReferralController.kt
mahalrmoj: Will do - was worried we are implicitly changing the contract in other places by changing the SentReferralDTO response object - but the client should ignore an extra field

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/dto/SentReferralWithActionPlanIdDTO.kt
mahalrmoj: Will revert back to SentReferralDTO and make change in that

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/ReferralService.kt
mahalrmoj: Conscious JIRA work was unplanned so didn't want it's scope to increase. Plus we're planning to have a specific mop up task to purge the DTOs from the business tier.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/jpa/entity/ActionPlan.kt
mahalrmoj: The toString is as per Kotlin which isn't strictly json - e.g. AuthUser(id=referral_repository_test_user_id, authSource=delius, userName=bernard.beaks)

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/controller/ReferralController.kt
tomsmyers: yep 👍

tomsmyers (approved):
nice work

PR #134, opened 2021-03-05
update app insights to use v3 agent

tomsmyers:

update app insights to use v3 agent
sends an exception event to app insights on uncaught exceptions being handled

improve application telemetry


.gitignore
```diff
@@ -66,3 +66,4 @@ dist
 ecr.repo
 /dps-gradle-spring-boot-suppressions.xml
 ddl_jpa_creation.sql
+applicationinsights-agent-*.jar
```
tomsmyers, line 69: included here for easier local testing of the agent

build.gradle.kts
```diff
@@ -38,9 +38,10 @@ tasks {
 }
 
 dependencies {
-  // logging
+  // logging + telemetry
   implementation("net.logstash.logback:logstash-logback-encoder:6.6")
   implementation("ch.qos.logback:logback-classic:1.2.3")
+  implementation("com.microsoft.azure:applicationinsights-agent:3.0.2")
```
rgforsyth, line 44: Is this not already brought in by the dps spring boot plugin? 
https://github.com/ministryofjustice/dps-gradle-spring-boot/blob/main/src/main/kotlin/uk/gov/justice/digital/hmpps/gradle/configmanagers/AppInsightsConfigManager.kt

rgforsyth (approved):

build.gradle.kts
tomsmyers: yeh good spot thanks

PR #135, opened 2021-03-08
IC-1314 cancel referral

rgforsyth:

Updated the referral table with cancelledAt & cancelledBy columns.
Adds a new endpoint that can be used to set the cancellation fields on a referral

To allow the cancellation of a referral.


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/controller/ReferralController.kt
```diff
@@ -87,6 +90,17 @@ class ReferralController(
     } ?: throw AccessDeniedException("user is not associated with a service provider organization")
   }
 
+  @PostMapping("/sent-referral/{id}/cancel")
```
sldblog, line 93: 🙋‍♂️ "cancel" is a context-sensitive name in the domain, it only applies until the first delivery appointment is attended, after that, it's different.

To me, it feels like this domain-specific name is unhelpful on an API level (we'd have to reject "cancel" after the first appointment).

Could we use a neutral term, like `/conclude` or `/finish` or `/end`?

sldblog: (From a process perspective, we'll have to call the action behind this internally when the last appointment is attended -- as a business rule, we will close the case when the last delivery session has been fulfilled)

rgforsyth: Are we not creating a separate endpoint for ended(more than one session). Which would set an ended at/by field on the referral?
Are you saying there should be a general purpose endpoint for both?

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/controller/ReferralController.kt
```diff
@@ -87,6 +90,17 @@ class ReferralController(
...
+  fun cancelSentReferral(@PathVariable id: UUID, authentication: JwtAuthenticationToken): CancelledReferralDTO {
+    val sentReferral = referralService.getSentReferral(id)
+      ?: throw ResponseStatusException(HttpStatus.NOT_FOUND, "draft referral not found [id=$id]")
```
mahalrmoj, line 96: Does the word draft apply here?

mahalrmoj (approved):
No more cancels - good one

PR #136, opened 2021-03-09
IC-1352 Created Appointments Database schema and repo

mahalrmoj:

Create the database structures necessary to implement sessions and appointments first cut functionality.

Provide the persistence tier on which to build the API functionality

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/jpa/entity/ActionPlan.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/jpa/entity/ActionPlan.kt
@@ -35,7 +35,7 @@ data class ActionPlan(
 
   // This isn't set directly and hence it's a 'val'. Once an action plan appointment has been persisted, on retrieval
   // of the associated action plan this list will be automatically set by hibernate.
-  @OneToMany(mappedBy = "actionPlan") val appointments: List<ActionPlanAppointment>,
+  @OneToMany(mappedBy = "actionPlan") val appointments: Set<ActionPlanAppointment>,
 
   // Required
   @NotNull @OneToOne @Fetch(FetchMode.JOIN) val referral: Referral,
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/jpa/entity/ActionPlan.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/jpa/entity/ActionPlan.kt
@@ -10,7 +10,6 @@ import javax.persistence.Entity
 import javax.persistence.Id
 import javax.persistence.JoinColumn
 import javax.persistence.ManyToOne
-import javax.persistence.OneToMany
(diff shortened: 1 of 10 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/jpa/entity/ActionPlanAppointment.kt
```diff
@@ -0,0 +1,31 @@
...
+  // Activities
+  var appointmentTime: OffsetDateTime? = null,
+  var durationInMinutes: Int? = null,
+
+  // Status
+  @NotNull @ManyToOne @Fetch(FetchMode.JOIN) val createdBy: AuthUser,
+  @NotNull val createdAt: OffsetDateTime,
+
+  // Required
+  @NotNull @ManyToOne val actionPlan: ActionPlan,
```
tomsmyers, line 29: do we want to specify the fetchtype here? it's likely we will not want the full action plan when looking up appointments? so lazy?

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/jpa/entity/ActionPlan.kt
```diff
@@ -32,6 +33,10 @@ data class ActionPlan(
   @ManyToOne @Fetch(FetchMode.JOIN) var submittedBy: AuthUser? = null,
   var submittedAt: OffsetDateTime? = null,
 
+  // This isn't set directly and hence it's a 'val'. Once an action plan appointment has been persisted, on retrieval
+  // of the associated action plan this list will be automatically set by hibernate.
+  @OneToMany(mappedBy = "actionPlan") val appointments: List<ActionPlanAppointment>,
```
tomsmyers, line 38: i read somewhere that using `Set` results in more efficient SQL than `List` https://vladmihalcea.com/hibernate-facts-favoring-sets-vs-bags/

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/jpa/entity/ActionPlanAppointment.kt
mahalrmoj: In "maintaining" appointments, I'm not sure we need to eagerly load action plans (and joined activities) as we don't use anything from action plan - so the action plan wont need to get hydrated.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/jpa/entity/ActionPlan.kt
mahalrmoj: On reflection a bi-directional relationship is not necessary. At present, an action plan doesn't need to know about its appointments as it's not part of the action plan dto returned to the ui. Appointments can be retrieved directly given the requirements of the current set of endpoints. So I have removed the list (which should have been a set) for now.

tomsmyers (approved):

PR #139, opened 2021-03-11
IC-1109 API for creating, listing and updating appointments

mahalrmoj:

Implements an API for creating, updating and listing appointments in an Action Plan

Provide functionality to allow sessions to be managed with service users to achieve the outcomes of a referral.


tomsmyers (approved):
nice

PR #140, opened 2021-03-11
use the same DTO for draft and submitted action plans

tomsmyers:

use the same DTO for draft and submitted action plans

make GET endpoints return the same data structure regardless of the state of the entity


rgforsyth (approved):
Looks good

mahalrmoj (approved):
Like it - much simpler

PR #142, opened 2021-03-12
IC-1381 Removed GET /draft-action-plan

mahalrmoj:

Removes GET /draft-action-plan/{id} 

The intent was so that a single URL can be used for getting the current state of an action plan without it going stale once the action plan has moved on from being a draft. This is particularly useful when retrieving links against old events.


sldblog:
> ake the URLs consistent and not distinguish by state for GET

For posterity, could we also expand on why we stopped doing it, please? (We could document and refer to these in https://github.com/ministryofjustice/hmpps-interventions-service/tree/main/doc/adr?)

tomsmyers (approved):

PR #144, opened 2021-03-12
IC-1312 get eosr by referral

rgforsyth:

Adds an endpoint to get an end of service report by referral id

To be able to retrieve an end of service report for a referral.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/controller/EndOfServiceReportController.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/controller/EndOfServiceReportController.kt
@@ -5,6 +5,7 @@ import org.springframework.security.oauth2.server.resource.authentication.JwtAut
 import org.springframework.web.bind.annotation.GetMapping
 import org.springframework.web.bind.annotation.PathVariable
 import org.springframework.web.bind.annotation.PostMapping
+import org.springframework.web.bind.annotation.RequestParam
 import org.springframework.web.bind.annotation.RestController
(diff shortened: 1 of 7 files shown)


sldblog:
I'm a bit confused, does this overlap with #143?

rgforsyth:
> I'm a bit confused, does this overlap with #143?

This story is to retrieve an end of service report by a referral id. The other story is to retrieve a eosr by the id of an eosr.
Each of these stories has been branched off the previous one. As they haven't been merged yet, the work from each is included in each pr if that's what you mean?

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/controller/EndOfServiceReportController.kt
```diff
@@ -0,0 +1,52 @@
...
+import java.util.UUID
+
+@RestController
+class EndOfServiceReportController(
+  val jwtAuthUserMapper: JwtAuthUserMapper,
+  val locationMapper: LocationMapper,
+  val endOfServiceReportService: EndOfServiceReportService,
+) {
+
+  @PostMapping("/referral/{id}/end-of-service-report")
```
sldblog, line 23: Given our intent here is to "start an end of service report for a referral", that relate to similar intents like "start an action plan for a referral", which we implemented as `POST /draft-action-plan`

Should we use that pattern here, too, to be consistent?

sldblog:
> As they haven't been merged yet, the work from each is included in each pr if that's what you mean?

Yes, thanks for clarifying. Using the other PR's branch as the base branch for this PR might be useful in these situations, if that's helpful for explicit ordering?

> wants to merge 1 commit into `IC-1309-get-eosr` from `IC-1312-get-eosr-by-referral-id`

PR #145, opened 2021-03-17
match service user names with local delius data for more constistent UX

tomsmyers:

changes the service user info in the seed data

make the data match the data held for the CRN in test delius so the UI has matching names


src/main/resources/db/local/V99_2_1__sent_referrals.sql
```diff
@@ -2,7 +2,7 @@ insert into referral (id, intervention_id, created_at, sent_at, sent_by_id, serv
 values ('81d754aa-d868-4347-9c0f-50690773014e', '98a42c61-c30f-4beb-8062-04033c376e2d', '2021-01-11 10:32:12.382884+00', '2021-01-14 15:56:45.382884+00', '2500099998', 'CRN23', '2021-04-01', '2500099998', 'Some information about the service user', 'Alex is currently sleeping on her aunt''s sofa', 'She uses a wheelchair', true, 'Spanish', true, 'She works Mondays 9am - midday', 'A danger to the elderly', true, 10, 'd0db50b0-4a50-4fc7-a006-9c97530e38b2', 'HD2123AC');
```
sldblog, line 2: 🙋‍♂️ The description has terms like "Alex is currently sleeping" which will be inconsistent with the new names

sldblog (approved):
LGTM 👍 

Noticed one source of inconsistency that might be offputting for users

PR #149, opened 2021-03-19
move previous migration change into new migration

rgforsyth:
…ommar from app insights json

Moves the previous pr change of adding a primary key into a new migration.

Attempt to fix deployment into dev region, due to a change to an existing migration.

### DIFF
--- a/src/main/resources/db/migration/v1_33__end_of_service_report_outcome_primary_key.sql
+++ b/src/main/resources/db/migration/v1_33__end_of_service_report_outcome_primary_key.sql
@@ -1,2 +1,0 @@
-alter table end_of_service_report_outcome
-    constraint pk_end_of_service_report_outcome primary key (end_of_service_report_id, desired_outcome_id);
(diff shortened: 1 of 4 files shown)


src/main/resources/db/migration/v1_33__end_of_service_report_outcome_primary_key.sql
```diff
@@ -0,0 +1,2 @@
+alter table end_of_service_report_outcome
+    constraint pk_end_of_service_report_outcome primary key (end_of_service_report_id, desired_outcome_id);
```
tomsmyers, line 2: we are missing `add` from this line

tomsmyers (changes requested):

src/main/resources/db/migration/v1_33__end_of_service_report_outcome_primary_key.sql
tomsmyers: i think?

tomsmyers (approved):

PR #153, opened 2021-03-21
IC-1405 add sample data for user research testing environment

tomsmyers:

adds a bunch of sample data for use in the user research environment

since we need lots of data for this i wrote a script to generate the referral migrations. it's a bit of a mess but it gets the job done.

make the UR environment useful


sldblog (commented):
LGTM so far!

Is it worth adding a script to apply the seed file?

Thinking about `script/seed_user_research_env.sh` which would spin up a new pod in the user research namespace to run `flyway` with this specific folder (and then quit normally)?

(Configuring the env to always use this seed via a profile is probably not what we want)

tomsmyers:
> LGTM so far!
> 
> Is it worth adding a script to apply the seed file?
> 
> Thinking about `script/seed_user_research_env.sh` which would spin up a new pod in the user research namespace to run `flyway` with this specific folder (and then quit normally)?
> 
> (Configuring the env to always use this seed via a profile is probably not what we want)

yeh - we can do this change as part of IC-1406 though. this ticket is focused soley on the seed data.

PR #159, opened 2021-03-24
ensure 'SPRING_PROFILES_ACTIVE' env var is set in the k8s environment

tomsmyers:

ensure 'SPRING_PROFILES_ACTIVE' env var is set in the k8s environment

apply the research migrations on deploy

### DIFF
--- a/helm_deploy/hmpps-interventions-service/templates/_envs.tpl
+++ b/helm_deploy/hmpps-interventions-service/templates/_envs.tpl
@@ -7,32 +7,10 @@ env:
   - name: SERVER_PORT
     value: "{{ .Values.image.port }}"
 
-  - name: SPRING_PROFILES_ACTIVE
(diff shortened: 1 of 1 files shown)


helm_deploy/hmpps-interventions-service/templates/_envs.tpl
```diff
@@ -7,6 +7,9 @@ env:
   - name: SERVER_PORT
     value: "{{ .Values.image.port }}"
 
+  - name: SPRING_PROFILES_ACTIVE
+    value: "{{ .Values.env.SPRING_PROFILES_ACTIVE }}"
```
sldblog, line 11: I think this will break if this isn't defined in `helm_deploy/values-dev.yaml`

tomsmyers: argh

sldblog (approved):

PR #160, opened 2021-03-24
Switch to "London (Metropolitan Police)" as the London PCC

tomsmyers:

Switch to Met Police PCC region for London

To make the 'find' page more user friendly.

The Met Police PCC covers almost the entire Greater London and is more applicable to use as the only PCC to cover the NPS region "London".

### DIFF
--- a/src/main/resources/db/migration/V1_37__use_different_london_pcc.sql
+++ b/src/main/resources/db/migration/V1_37__use_different_london_pcc.sql
@@ -1,0 +1,9 @@
+insert into pcc_region (id, name, nps_region_id) values
+    ('metropolitan', 'London (Metropolitan Police)', 'J');
+
+update dynamic_framework_contract
+    set pcc_region_id = 'metropolitan'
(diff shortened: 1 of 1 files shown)


sldblog (approved):
Could we also do this in https://github.com/ministryofjustice/hmpps-reference-data/blob/main/pcc-regions-for-probation-v0.csv, please?

sldblog:
I've asked around and the current proposal is:

- Remove City of London PCC
- Add Met Police PCC with the name of "London (Metropolitan Police)" and the ID used in https://data.police.uk/api/forces

I'll amend the PR here

rgforsyth (approved):

rgforsyth (changes requested):
Version number on migration will need to be updated. Sorry

sldblog:
@rgforsyth Done!

rgforsyth (approved):
👍

sldblog:
🙇‍♂️ Thank you

PR #162, opened 2021-03-24
IC-1432: Add longer descriptions

tomsmyers:

Change text field in intervention entity to hold a longer description, with new lines and Unicode characters.

Updates the research data migration generator to handle multiline text fields.

Adds long descriptions for the user research interventions.

Allow interventions to contain more data in 'find'


sldblog (dismissed):
LGTM

I've removed the `clean-on-validation-errors` configuration as we run the deployment with `SPRING_PROFILES_ACTIVE: research` and that would mean a new app migration would reset the entire thing, which is not good on that env

mahalrmoj (approved):
Makes sense to me

src/main/resources/db/migration/V1_36__add_long_description_to_intervention.sql
```diff
@@ -0,0 +1,6 @@
+alter table intervention
+    add column long_description text not null default '';
+
+update intervention
+    set long_description = description
+    where long_description = '';
```
rgforsyth, line 6: Not sure if no newline will cause an issue in circle. It usually fails for Kotlin files.

rgforsyth (approved):
Looks good. Small question about missing newline in .sql files

sldblog (commented):
@Gweaton highlighted in the UI PR (https://github.com/ministryofjustice/hmpps-interventions-ui/pull/188) that the service's intent is to have a single description field

We'll have to decide I guess to add this and squash later or change the long content to be in a single description

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/InterventionServiceTest.kt
```diff
@@ -85,6 +87,23 @@ class InterventionServiceTest @Autowired constructor(
 //    assertThat(interventionDTO.pccRegions.size).isEqualTo(1)
   }
 
+  @Test
```
rgforsyth, line 90: Should testing retrieving data from the db be done as a repository test. I think elsewhere service tests have mocked out the repositories?

Although, I can see that other tests in this file are doing it the way you have below

rgforsyth (approved):

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/InterventionServiceTest.kt
tomsmyers: i'm really keen for this sort of test to actually involve the database, especially since we want to test that the db itself does not modify the character encoding or formatting.

sldblog (commented):
Thanks for the review request -- I've changed the PR so I'll defer to others :)
The changes to descriptions and generator are good from my perspective

PR #166, opened 2021-03-29
IC-1480: Document simpler tables for data dictionary

sldblog:

Document simpler tables for data dictionary.

To extend https://hmpps-interventions-service-dev.apps.live-1.cloud-platform.service.justice.gov.uk/meta/schema/ with definitions for more tables.

### DIFF
--- a/src/main/resources/db/migration/V1_38__document_simpler_tables.sql
+++ b/src/main/resources/db/migration/V1_38__document_simpler_tables.sql
@@ -1,0 +1,29 @@
+COMMENT ON TABLE auth_user IS 'details about the user from hmpps-auth';
+COMMENT ON COLUMN auth_user.id IS 'the user ID';
+COMMENT ON COLUMN auth_user.auth_source IS 'where the user has come from';
+COMMENT ON COLUMN auth_user.user_name IS 'the username';
+
+COMMENT ON TABLE service_category IS '[reference data] intervention service categories, which relate to service user needs';
+COMMENT ON COLUMN service_category.id IS 'service-owned unique identifier';
+COMMENT ON COLUMN service_category.created IS 'when the record was added';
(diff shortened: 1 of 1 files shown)


src/main/resources/db/migration/V1_38__document_simpler_tables.sql
```diff
@@ -0,0 +1,29 @@
+COMMENT ON TABLE auth_user IS 'details about the user from hmpps-auth';
+COMMENT ON COLUMN auth_user.id IS 'the user ID';
+COMMENT ON COLUMN auth_user.auth_source IS 'where the user has come from';
```
tomsmyers, line 3: i think this description is lacking in clarity, but explaining this field in a few words is tricky

tomsmyers (approved):

src/main/resources/db/migration/V1_38__document_simpler_tables.sql
sldblog: would be great to iterate on it!

src/main/resources/db/migration/V1_38__document_simpler_tables.sql
```diff
@@ -0,0 +1,29 @@
...
+COMMENT ON COLUMN auth_user.user_name IS 'the username';
+
+COMMENT ON TABLE service_category IS '[reference data] intervention service categories, which relate to service user needs';
+COMMENT ON COLUMN service_category.id IS 'service-owned unique identifier';
+COMMENT ON COLUMN service_category.created IS 'when the record was added';
+COMMENT ON COLUMN service_category.name IS 'intervention service category';
+
+COMMENT ON TABLE complexity_level IS '[reference data] complexity levels for each intervention service category';
```
sldblog, line 11: TIL this appears as a link in the app -- so I suppose we can use Markdown here
image

PR #168, opened 2021-03-29
IC-1296 send email when appointment not attended

rgforsyth:

When recording appointment attendance, if the appointment was not attended, email the pp.

Email the PP to let them know if an appointment was not attended

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/component/LocationMapper.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/component/LocationMapper.kt
@@ -1,12 +1,22 @@
 package uk.gov.justice.digital.hmpps.hmppsinterventionsservice.component
 
 import org.springframework.stereotype.Component
+import org.springframework.web.bind.annotation.GetMapping
 import org.springframework.web.servlet.support.ServletUriComponentsBuilder
+import java.lang.RuntimeException
 import java.net.URI
+import kotlin.reflect.KFunction
+import kotlin.reflect.full.findAnnotation
 
 @Component
 class LocationMapper {
   fun expandPathToCurrentRequestBaseUrl(path: String, vararg uriVariableValues: Any): URI {
     return ServletUriComponentsBuilder.fromCurrentRequestUri().path(path).buildAndExpand(*uriVariableValues).toUri()
+  }
+
+  fun getPathFromControllerMethod(method: KFunction<*>): String {
+    val annotation = method.findAnnotation<GetMapping>()
+      ?: throw RuntimeException("method '${method.name}' does not have a GetMapping annotation")
+    return annotation.value.first()
   }
 }
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/events/AppointmentEventPublisher.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/events/AppointmentEventPublisher.kt
@@ -1,0 +1,34 @@
+package uk.gov.justice.digital.hmpps.hmppsinterventionsservice.events
+
+import org.springframework.context.ApplicationEvent
+import org.springframework.context.ApplicationEventPublisher
+import org.springframework.stereotype.Component
+import uk.gov.justice.digital.hmpps.hmppsinterventionsservice.component.LocationMapper
+import uk.gov.justice.digital.hmpps.hmppsinterventionsservice.controller.AppointmentsController
+import uk.gov.justice.digital.hmpps.hmppsinterventionsservice.jpa.entity.ActionPlanAppointment
+
+enum class AppointmentEventType {
+  NO_ATTENDANCE,
+}
+
+class AppointmentEvent(source: Any, val type: AppointmentEventType, val appointment: ActionPlanAppointment, val detailUrl: String) : ApplicationEvent(source) {
+  override fun toString(): String {
+    return "AppointmentEvent(type=$type, appointment=${appointment.id}, detailUrl='$detailUrl', source=$source)"
+  }
+}
+
+@Component
+class AppointmentEventPublisher(
+  private val applicationEventPublisher: ApplicationEventPublisher,
+  private val locationMapper: LocationMapper
+) {
+
+  fun appointmentNotAttendedEvent(appointment: ActionPlanAppointment) {
+    applicationEventPublisher.publishEvent(AppointmentEvent(this, AppointmentEventType.NO_ATTENDANCE, appointment, getAppointmentAttendanceURL(appointment)))
+  }
+
+  private fun getAppointmentAttendanceURL(appointment: ActionPlanAppointment): String {
+    val path = locationMapper.getPathFromControllerMethod(AppointmentsController::getAppointment)
+    return locationMapper.expandPathToCurrentRequestBaseUrl(path, appointment.actionPlan.id, appointment.sessionNumber).toString()
+  }
+}
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/AppointmentsService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/AppointmentsService.kt
@@ -1,6 +1,7 @@
 package uk.gov.justice.digital.hmpps.hmppsinterventionsservice.service
 
 import org.springframework.stereotype.Service
+import uk.gov.justice.digital.hmpps.hmppsinterventionsservice.events.AppointmentEventPublisher
 import uk.gov.justice.digital.hmpps.hmppsinterventionsservice.jpa.entity.ActionPlanAppointment
 import uk.gov.justice.digital.hmpps.hmppsinterventionsservice.jpa.entity.Attended
 import uk.gov.justice.digital.hmpps.hmppsinterventionsservice.jpa.entity.AuthUser
@@ -17,6 +18,7 @@ class AppointmentsService(
   val actionPlanAppointmentRepository: ActionPlanAppointmentRepository,
   val actionPlanRepository: ActionPlanRepository,
   val authUserRepository: AuthUserRepository,
+  val appointmentEventPublisher: AppointmentEventPublisher
 ) {
   fun createAppointment(
     actionPlanId: UUID,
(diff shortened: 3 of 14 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/component/LocationMapper.kt
```diff
@@ -19,4 +19,8 @@ class LocationMapper {
   fun mapToCurrentContextPathAsString(path: String, id: UUID): UriComponents {
     return ServletUriComponentsBuilder.fromCurrentContextPath().path(path).buildAndExpand(id)
   }
+
+  fun mapToCurrentContextPathAsAppointment(path: String, id: UUID, sessionsNumber: Int): UriComponents {
+    return ServletUriComponentsBuilder.fromCurrentContextPath().path(path).buildAndExpand(id, id)
```
tomsmyers, line 24: there's a typo here, we are expanding `id` twice. a unit test would catch this. 

i also think the method name here is really unhelpful. i wouldn't be able to guess what this method does from the name. same with the method above.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/events/AppointmentEventPublisher.kt
```diff
@@ -0,0 +1,39 @@
...
+import uk.gov.justice.digital.hmpps.hmppsinterventionsservice.jpa.entity.ActionPlanAppointment
+import kotlin.reflect.KFunction
+
+enum class AppointmentEventType {
+  NO_ATTENDANCE,
+}
+
+class AppointmentEvent(source: Any, val type: AppointmentEventType, val appointment: ActionPlanAppointment, val detailUrl: String) : ApplicationEvent(source) {
+  override fun toString(): String {
+    return "AppointmentEvent(type=$type, appointment=${appointment.id}, detailUrl='$detailUrl', source=$source)"
```
tomsmyers, line 18: why the single quotes around the url?

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/AppointmentsService.kt
```diff
@@ -63,9 +65,16 @@ class AppointmentsService(
   ): ActionPlanAppointment {
     val appointment = getActionPlanAppointmentOrThrowException(actionPlanId, sessionNumber)
     setAttendanceFields(appointment, attended, additionalInformation)
+    if (attended == Attended.NO) {
+      notifyNoAttendance(appointment)
+    }
     return actionPlanAppointmentRepository.save(appointment)
   }
 
+  private fun notifyNoAttendance(appointment: ActionPlanAppointment) {
```
tomsmyers, line 74: does this need to be in its own method?

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/NotifyService.kt
```diff
@@ -49,6 +51,35 @@ class NotifyActionPlanService(
...
+  @Value("\${interventions-ui.locations.appointment-not-attended}") private val interventionsUIAppointmentNotAttendedLocation: String,
+  private val emailSender: EmailSender,
+  private val hmppsAuthService: HMPPSAuthService,
+) : ApplicationListener<AppointmentEvent>, NotifyService {
+
+  @AsyncEventExceptionHandling
+  override fun onApplicationEvent(event: AppointmentEvent) {
+    when (event.type) {
+      AppointmentEventType.NO_ATTENDANCE -> {
+        val userDetails = hmppsAuthService.getUserDetail(event.appointment.actionPlan.referral.createdBy)
```
tomsmyers, line 67: i'm sure we talked about this but i can't remember the outcome - do we want the notification to go to the PP who created the referral, or the one who sent it? 

either way, we should ensure we are consistent with these types of notifications. could we add a helper method to the referral for 'get responsible pp' or something?

tomsmyers: nitpick: i would also rename this variable so it's clear who's user details these are.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/NotifyService.kt
```diff
@@ -49,6 +51,35 @@ class NotifyActionPlanService(
...
+        val location = generateResourceUrl(interventionsUIBaseURL, interventionsUIAppointmentNotAttendedLocation, event.appointment.id)
+        emailSender.sendEmail(
+          appointmentNotAttendedTemplateID,
+          userDetails.email,
+          mapOf(
+            "ppFirstName" to userDetails.firstName,
+            "referralNumber" to event.appointment.actionPlan.referral.referenceNumber!!,
```
tomsmyers, line 74: we should probably update this var name in the template so it's clear it's expecting the reference number. 'referralNumber' is ambiguous - it could be the referral id too.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/NotifyService.kt
```diff
@@ -49,6 +51,35 @@ class NotifyActionPlanService(
...
+            "deliusAbsence" to location.toString(),
```
tomsmyers, line 75: same with this variable - it's not clear that it's expecting a link

src/main/resources/application.yml
```diff
@@ -76,11 +76,13 @@ notify:
     referral-sent: "336f2158-d130-4c63-a59b-da12df15a5a3"
     referral-assigned: "f0fe2fce-df0f-46c0-839a-822037d39c28"
     action-plan-submitted: "83743b43-9abb-4ffb-913d-dc583ab22dbf"
+    appointment-not-attended: "21b14c86-30d2-4bf2-b355-8f3deb62ef69"
 
 interventions-ui:
   locations:
     sent-referral: "/service-provider/referrals/{id}"
     submit-action-plan: "/action-plan/{id}/toBeSpecified"
+    appointment-not-attended: "/appointment/{id}/toBeSpecified"
```
tomsmyers, line 85: not a blocker - is it better to have these links point to the pp dashboard, so at least they actually work, rather than to a 404 page?

tomsmyers (changes requested):
nice work

one change required (URL params) and some smaller things to talk through

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/AppointmentsService.kt
mahalrmoj: Perhaps the check for non-attendance should be in the method and the method called notifyIfNonAttendance which means the main method just orchestrates

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/component/LocationMapper.kt
tomsmyers: this has been addressed in a separate change so is no longer an issue

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/events/AppointmentEventPublisher.kt
tomsmyers: leaving for now since it matches the other event string

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/AppointmentsService.kt
tomsmyers: 👍  done

src/main/resources/application.yml
tomsmyers: done 

https://dsdmoj.atlassian.net/browse/IC-1482
https://dsdmoj.atlassian.net/browse/IC-1483

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/component/LocationMapper.kt
```diff
@@ -1,12 +1,22 @@
...
+import kotlin.reflect.KFunction
+import kotlin.reflect.full.findAnnotation
 
 @Component
 class LocationMapper {
   fun expandPathToCurrentRequestBaseUrl(path: String, vararg uriVariableValues: Any): URI {
     return ServletUriComponentsBuilder.fromCurrentRequestUri().path(path).buildAndExpand(*uriVariableValues).toUri()
   }
+  
+  fun getPathFromControllerMethod(method: KFunction<*>): String {
```
tomsmyers, line 17: this is split into its own commit. please review that in isolation to see this change.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/NotifyService.kt
```diff
@@ -49,6 +51,35 @@ class NotifyActionPlanService(
...
+    when (event.type) {
+      AppointmentEventType.NO_ATTENDANCE -> {
+        val ppDetails = hmppsAuthService.getUserDetail(event.appointment.actionPlan.referral.getResponsibleProbationPractitioner())
+        val location = generateResourceUrl(interventionsUIBaseURL, interventionsUIAppointmentNotAttendedLocation, event.appointment.id)
+        emailSender.sendEmail(
+          appointmentNotAttendedTemplateID,
+          ppDetails.email,
+          mapOf(
+            "ppFirstName" to ppDetails.firstName,
+            "referenceNumber" to event.appointment.actionPlan.referral.referenceNumber!!,
```
tomsmyers, line 74: i have updated these var names in the template

src/main/resources/application.yml
```diff
@@ -76,11 +76,13 @@ notify:
     referral-sent: "336f2158-d130-4c63-a59b-da12df15a5a3"
     referral-assigned: "f0fe2fce-df0f-46c0-839a-822037d39c28"
     action-plan-submitted: "83743b43-9abb-4ffb-913d-dc583ab22dbf"
+    appointment-not-attended: "21b14c86-30d2-4bf2-b355-8f3deb62ef69"
 
 interventions-ui:
   locations:
     sent-referral: "/service-provider/referrals/{id}/details"
```
mahalrmoj, line 83: Just a thought but should this one be progress to rather than service provider? It is used in the NSI/Referrals url in delius

mahalrmoj (approved):

src/main/resources/application.yml
```diff
@@ -76,11 +76,13 @@ notify:
...
-    submit-action-plan: "/action-plan/{id}/toBeSpecified"
+    submit-action-plan: "/service-provider/referrals/{id}/progress" // fixme
```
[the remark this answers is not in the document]
tomsmyers: yeh right now there is no pp facing screen with this info. but with the current implementation the pp can actually load this page. both ways are not ideal, but i thought perhaps this way was better than an error?

tomsmyers (approved):

PR #169, opened 2021-03-29
IC-535 Fixed email recipient - referral sender

mahalrmoj:

When an action plan is submitted the notification email should be sent to the "Sender of the Referral". 

Fix as it was being sent to the wrong recipient.


tomsmyers (approved):
👍 thanks!

PR #170, opened 2021-03-30
IC-1403 Correct referrals backlink url from delius

mahalrmoj:

Corrects the url link from Delius as a result of creating an NSI via community api

To fix a problems and be able to link back from Delius


tomsmyers (approved):
👍

PR #172, opened 2021-03-30
IC-1412automatically create empty appointments on action plan submission

rgforsyth:

When a action plan is submitted, empty appointments will now be created based on the number of sessions the action plan has.

Removed the need for the user to have to create the appointments manually

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/ActionPlanService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/ActionPlanService.kt
@@ -20,7 +20,8 @@ class ActionPlanService(
   val referralRepository: ReferralRepository,
   val actionPlanRepository: ActionPlanRepository,
   val actionPlanValidator: ActionPlanValidator,
-  val actionPlanEventPublisher: ActionPlanEventPublisher
+  val actionPlanEventPublisher: ActionPlanEventPublisher,
+  val appointmentsService: AppointmentsService,
 ) {
 
   fun createDraftActionPlan(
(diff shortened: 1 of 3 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/ActionPlanService.kt
```diff
@@ -82,6 +86,20 @@ class ActionPlanService(
       ?: throw EntityNotFoundException("action plan not found [referralId=$referralId]")
   }
 
+  private fun createEmptyAppointments(draftActionPlan: ActionPlan, submittedByUser: AuthUser) {
+    val numberOfSessions = draftActionPlan.numberOfSessions
```
tomsmyers, line 90: nit: would be better to make the not null assertion on this line

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/ActionPlanService.kt
```diff
@@ -82,6 +86,20 @@ class ActionPlanService(
       ?: throw EntityNotFoundException("action plan not found [referralId=$referralId]")
   }
 
+  private fun createEmptyAppointments(draftActionPlan: ActionPlan, submittedByUser: AuthUser) {
```
tomsmyers, line 89: nit: at the point this method is being called, the action plan is no longer a draft

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/ActionPlanServiceTest.kt
```diff
@@ -207,11 +217,18 @@ internal class ActionPlanServiceTest {
...
+      SampleData.sampleActionPlanAppointment(
+        actionPlan = SampleData.sampleActionPlan(), createdBy = SampleData.sampleAuthUser()
+      )
+    )
 
     val submittedActionPlan = actionPlanService.submitDraftActionPlan(actionPlanId, authUser)
 
     assertThat(submittedActionPlan).isNotNull
     verify(actionPlanValidator).validateSubmittedActionPlan(any())
+    verify(actionPlanAppointmentRepository, times(2)).save(any())
```
tomsmyers, line 231: it's somewhat of a shame we can't validate that the two records are actually created against the current action plan here. could we use argument captors?

mahalrmoj (commented):
Looks ok but I just had one question about createdBy on the appointment. In this new scheme, it reflects the submitter of the action plan and not the person who enters the appointment. Presumably this is ok

tomsmyers (approved):

PR #173, opened 2021-03-31
Refactor 'LocationMapper' so that all required functionality is in a single method

tomsmyers:

Refactor 'LocationMapper' so that all required functionality is in a single method

add a unit test for the new method.

improves the readabillity and maintainability of the location mapper class.


mahalrmoj (approved):
That is much cleaner and I like the way you did the test using RequestContextHolder

PR #178, opened 2021-04-07
fix the helm template ports for UR deploy

tomsmyers:

fix the helm template ports for UR deploy

unblock UR deploy

### DIFF
--- a/helm_deploy/values-research.yaml
+++ b/helm_deploy/values-research.yaml
@@ -3,7 +3,9 @@ replicaCount: 2
 image:
   repository: quay.io/hmpps/hmpps-interventions-service
   tag: latest
-  port: 8080
+  ports:
+    app: 8080
+    dataDictionary: 8080
 
 ingress:
   enabled: true


mahalrmoj (approved):

helm_deploy/values-research.yaml
```diff
@@ -3,7 +3,9 @@ replicaCount: 2
 image:
   repository: quay.io/hmpps/hmpps-interventions-service
   tag: latest
-  port: 8080
+  ports:
+    app: 8080
+    dataDictionary: 8080
```
sldblog, line 8: I'm kinda thinking this kind of configuration isn't _very_ useful. Unless we introduce a new container, I doubt these ports will change that much. And even if they do, we could introduce the branching while we roll it out, instead of always being there as configuration.

What do you say that we remove all this port config from `values-*.yaml` files?

sldblog: Same with `ingress.enabled` tbh -- we know our service has an ingress

PR #179, opened 2021-04-07
IC-1426 API for recording session behaviour and notifying PP if necessary

tomsmyers:

new endpoint and service methods for recording SU behaviour during a session.

new event type for informing the PP on unacceptable behaviour

event refactoring to push the logic about whether or not to inform the PP into the service layer.

new contract state to satify the contract tests for this on the UI.

alllow SPs to record SU behaviour for a session.


mahalrmoj (approved):
So conceptually similar to your first trial of changes.

The only thing that comes to mind is at what point should we start chopping up the the service files that contain many classes.

Also, now appointment holds sessions, attendance details and a resulting action. I suppose at some point we may need to normalise and have embedded components - but yeah let's do it when we need to. 

Great progress.

PR #186, opened 2021-04-13
DAPI-88 Improved request objects in calls to community api to use off…

mahalrmoj:

After completing the end to end integration between Interventions and Delius UI (via community-api and delius-api) a number of improvements were identified in order to have a working integration. These comprise:

- Interventions to provide offsets with dates and times so that appointments and referrals can be converted correctly for Delius which relies on Europe/London local dates and times
- Use UUIDs over long text to identify reference data (e.g. service category). The use of descriptions was deemed fragile ( business codes are not available for service categories)
- Provide integration context in all requests to identify the correct mapping configuration for a client of community-api

To achieve a workable integration between new Interventions Service and the existing Case Management System Delius for both referrals and appointments creation.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/CommunityAPIBookingService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/CommunityAPIBookingService.kt
@@ -1,14 +1,11 @@
 package uk.gov.justice.digital.hmpps.hmppsinterventionsservice.service
 
-import com.fasterxml.jackson.annotation.JsonFormat
 import mu.KLogging
 import org.springframework.beans.factory.annotation.Value
 import org.springframework.stereotype.Service
 import org.springframework.web.util.UriComponentsBuilder
 import uk.gov.justice.digital.hmpps.hmppsinterventionsservice.component.CommunityAPIClient
 import uk.gov.justice.digital.hmpps.hmppsinterventionsservice.jpa.entity.ActionPlanAppointment
-import java.time.LocalDate
-import java.time.LocalTime
 import java.time.OffsetDateTime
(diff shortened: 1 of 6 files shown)


rgforsyth (approved):

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/CommunityAPIBookingService.kt
```diff
@@ -70,21 +66,11 @@ class CommunityAPIBookingService(
 }
 
 data class AppointmentCreateRequestDTO(
-
-  @JsonFormat(pattern = "yyyy-MM-dd")
```
tomsmyers, line 74: glad to see these go in favour of application level config

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/CommunityAPIBookingService.kt
```diff
@@ -70,21 +66,11 @@ class CommunityAPIBookingService(
...
-  @NotNull val officeLocationCode: String,
-
-  @NotNull val notes: String,
-
-  @NotNull val context: String,
+  val appointmentStart: OffsetDateTime,
+  val appointmentEnd: OffsetDateTime,
+  val officeLocationCode: String,
+  val notes: String,
+  val context: String,
```
tomsmyers, line 73: is the removal of the not null annotation related to the date formatting changes?

mahalrmoj: As a request object I didn't see what value they provided. Not null is enforced by Kotlin.

PR #187, opened 2021-04-14
Only allow reset for the research DB from main

sldblog:

Constraints user research environment reset to be only possible on `main` branches: https://app.circleci.com/pipelines/github/ministryofjustice/hmpps-interventions-service?branch=main

This has two benefits as far as I can see

1. Ensures we only deploy seed files on the mainline, which reduces potential misunderstandings (trial deployment can still be made locally with `helm` CLI)
2. Removes the "never green" check on PRs, e.g.:
    image

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -236,6 +236,9 @@ workflows:
             - hmpps-interventions-service-research
       - approve_reset_research_db:
           type: approval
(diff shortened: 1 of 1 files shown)


tomsmyers (approved):
yeh much better - thanks

PR #188, opened 2021-04-14
IC-1453 add patch endpoint to update action plan activity details

rgforsyth:

Adds a new endpoint to allow action plan activities to be updated.

Allows action plan activities to be updated.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/dto/UpdateActionPlanActivityDTO.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/dto/UpdateActionPlanActivityDTO.kt
@@ -1,8 +1,0 @@
-package uk.gov.justice.digital.hmpps.hmppsinterventionsservice.dto
-
-import java.util.UUID
-
-data class UpdateActionPlanActivityDTO(
-  val description: String?,
-  val desiredOutcomeId: UUID?
-)
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/dto/UpdateActionPlanActivityDTO.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/dto/UpdateActionPlanActivityDTO.kt
@@ -1,0 +1,8 @@
+package uk.gov.justice.digital.hmpps.hmppsinterventionsservice.dto
+
+import java.util.UUID
+
+data class UpdateActionPlanActivityDTO(
+  val description: String?,
+  val desiredOutcomeId: UUID?
+)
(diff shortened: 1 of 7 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/dto/UpdateActionPlanActivityDTO.kt
```diff
@@ -0,0 +1,8 @@
+package uk.gov.justice.digital.hmpps.hmppsinterventionsservice.dto
```
tomsmyers, line 1: i feel like this would be more at home alongside `ActionPlanActivityDTO` which is actually inside `ActionPlanDTO.kt`. either that or split both out into their own classes?

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/ActionPlanService.kt
```diff
@@ -60,6 +63,17 @@ class ActionPlanService(
     return actionPlanRepository.save(draftActionPlan)
   }
 
+  fun updateActionPlanActivity(actionPlanId: UUID, activityId: UUID, description: String?, desiredOutcomeId: UUID?): ActionPlan {
+    val actionPlan = getActionPlan(actionPlanId)
+    actionPlan.activities.forEach { x ->
```
tomsmyers, line 68: i think a more descriptive var name would improve readbility here

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/ActionPlanService.kt
```diff
@@ -108,4 +122,10 @@ class ActionPlanService(
     draftActionPlan.submittedBy = authUserRepository.save(submittedByUser)
     return draftActionPlan
   }
+
+  private fun getDesiredOutcomeByIdOrElseThrowException(id: UUID): DesiredOutcome {
+    return desiredOutcomeRepository.findById(id).orElseThrow {
```
tomsmyers, line 127: nit: i think `findByIdOrNull(id) ?: throw ...` is more idiomatic kotlin

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/ActionPlanServiceTest.kt
```diff
@@ -143,10 +147,9 @@ internal class ActionPlanServiceTest {
 
     whenever(
       actionPlanRepository.save(
-        ArgumentMatchers.argThat {
-          (
-            numberOfSessionsArg, activitiesArg, _, _, _, _, _, _
-          ) ->
+        ArgumentMatchers.argThat { (
```
tomsmyers, line 150: are these nessecary? can't we just match `any()`?

tomsmyers (approved):
nice one, thanks!

PR #190, opened 2021-04-15
IC-1315 cancellation reasons

rgforsyth:

Adds a new table storing cancellation reasons, and an endpoint to retrieve them.

Allows the front end to request all the possible cancellation reasons.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/controller/ReferralController.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/controller/ReferralController.kt
@@ -162,7 +162,7 @@ class ReferralController(
     return referralService.getSentReferralsSentBy(user).map { SentReferralDTO.from(it) }
   }
 
-  @GetMapping("/cancellation-reasons")
+  @GetMapping("/referral-cancellation-reasons")
   fun getCancellationReasons(): CancellationReasonsDTO {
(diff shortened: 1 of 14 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/controller/ReferralController.kt
```diff
@@ -161,6 +162,11 @@ class ReferralController(
     return referralService.getSentReferralsSentBy(user).map { SentReferralDTO.from(it) }
   }
 
+  @GetMapping("/cancellation-reasons")
```
tomsmyers, line 165: wondering if this should be prefixed with `/referral`?

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/jpa/entity/SampleData.kt
```diff
@@ -298,6 +298,13 @@ class SampleData {
       return AuthUser(id, authSource, userName)
     }
 
+    fun sampleCancellationReason(
```
tomsmyers, line 301: let's add this to the factory classses instead since we decided to try to use those methods more consistently

tomsmyers (commented):
one outstanding question on the url

tomsmyers (approved):

PR #191, opened 2021-04-16
change /sent-referrals endpoint to accept query params for filtering referrals based on different criteria

tomsmyers:

change /sent-referrals endpoint to accept query params for filtering referrals based on different criteria

allows getting sent referrals based on who they were sent by, which SP org they were sent to, and which case worker they are assigned to.

this makes the endpoint usable by both PP and SP pages in the UI


mahalrmoj (approved):
LGTM

rgforsyth (approved):

PR #192, opened 2021-04-16
add pact state for submitting appointment attendance and behaviour

rgforsyth:

Adds pact state for submitting appointment behaviour & attendance.

Fix failing pact test.


tomsmyers (approved):
awesome - thank you!

PR #184, opened 2021-04-13
IC-1505: Upload daily snapshots to Analytical Platform

sldblog:
## What does this pull request do?

Periodically exports all tables to CSVs and then uploads those into a timestamped "folder" in an Analytical Platform S3 "landing" bucket for further analytical processing

It is scheduled for 1 am every day (UTC)

🙋 Uses the new `data-engineering-data-extractor` image: https://github.com/ministryofjustice/data-engineering-data-extractor/pull/4
🙋 The `s3://hmpps-interventions-dev-landing` bucket is provisioned by the Analytical Platform, but access to it is controlled via https://github.com/ministryofjustice/cloud-platform-environments/pull/4604

## Depends on

- [x] https://github.com/ministryofjustice/data-engineering-data-extractor/pull/4
- [x] https://github.com/ministryofjustice/cloud-platform-environments/pull/4604
- [x] https://github.com/ministryofjustice/cloud-platform-environments/pull/4607
- [x] https://github.com/ministryofjustice/data-engineering-data-extractor/pull/6
- there will be further PRs to add the policy/buckets to other environments

## What is the intent behind these changes?

To enable the analysis and processing of interventions data.

PR #195, opened 2021-04-16
Upload daily snapshots for reporting

sldblog:

Periodically exports all tables to CSVs and then uploads those into a timestamped "folder" in an NDMIS S3 "landing" bucket for a reporting ETL process.

It has exactly the same pattern as the analytics data extractor (#184), except for configuration differences.

However, the script will be slightly different over time, so there's no point in further DRYing at the moment.

## Waiting on

- [x] Depends on changes introduced in #193, that one has to be merged first, then I can change the base branch of this PR to `main`
- [x] Also depends on https://github.com/ministryofjustice/cloud-platform-environments/pull/4635

To enable the reporting on interventions data.

### DIFF
--- a/helm_deploy/hmpps-interventions-service/templates/cronjob-data-extractor-reporting.yaml
+++ b/helm_deploy/hmpps-interventions-service/templates/cronjob-data-extractor-reporting.yaml
@@ -1,7 +1,7 @@
 apiVersion: batch/v1beta1
 kind: CronJob
 metadata:
-  name: data-extractor
+  name: data-extractor-reporting
 spec:
   schedule: "0 1 * * *"
   jobTemplate:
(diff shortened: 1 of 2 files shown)


helm_deploy/hmpps-interventions-service/templates/cronjob-data-extractor-reporting.yaml
```diff
@@ -0,0 +1,59 @@
...
+  jobTemplate:
+    spec:
+      template:
+        spec:
+          restartPolicy: "Never"
+          containers:
+          - name: data-extractor-reporting
+            image: ministryofjustice/data-engineering-data-extractor:develop
+            {{ if (and .Values.env_details.contains_live_data (not .data_protection_impact_assessment_signed_off)) }}
+            command: ["sh", "-c", "psql -c 'SELECT 1' && mkdir -p 'export/csv' && touch 'export/csv/signed_off_DPIA_required.csv' && transfer_local_to_s3.sh"]
```
tomsmyers, line 16: what's the purpose of `select 1` here?

tomsmyers (approved):

PR #196, opened 2021-04-16
IC-1506: Document data dictionary in repeatable migration

sldblog:

We want to update the comments on tables and columns to update the data dictionary, which is deployed to 🕸️https://hmpps-interventions-service-dev.apps.live-1.cloud-platform.service.justice.gov.uk/meta/schema/

Repeatable migrations re-run if the contents change, so they are ideal as a way to keep the dictionary up-to-date: we can edit this file any time to add/change definitions.

📝 See also: https://flywaydb.org/documentation/getstarted/advanced/repeatable
➕ **Bonus** Fixes the accidental links in the dictionary by changing `[reference data]` to `**reference data**` (it will show up as bold)

To keep the data dictionary more up-to-date. (It already has new columns and tables 😂)


tomsmyers (approved):
nice - this repeatable migration format does seem like a good way to maintain the reference data in a more visible way than the regular migrations too.

PR #197, opened 2021-04-19
IC-1356 allow end endpoint to accept cancellation reason

rgforsyth:

Allows a cancellation reason to be passed the the 'end' endpoint.

Allows a cancellation reason to be persisted when ending a referral.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/controller/mappers/CancellationReasonMapper.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/controller/mappers/CancellationReasonMapper.kt
@@ -1,0 +1,18 @@
+package uk.gov.justice.digital.hmpps.hmppsinterventionsservice.controller.mappers
+
+import org.springframework.data.repository.findByIdOrNull
+import org.springframework.http.HttpStatus
+import org.springframework.stereotype.Component
+import org.springframework.web.server.ResponseStatusException
+import uk.gov.justice.digital.hmpps.hmppsinterventionsservice.jpa.entity.CancellationReason
+import uk.gov.justice.digital.hmpps.hmppsinterventionsservice.jpa.repository.CancellationReasonRepository
+
+@Component
+class CancellationReasonMapper(
+  private val cancellationReasonRepository: CancellationReasonRepository
+) {
+  fun mapCancellationReasonIdToCancellationReason(cancellationReasonId: String): CancellationReason {
+    return cancellationReasonRepository.findByIdOrNull(cancellationReasonId)
+      ?: throw ResponseStatusException(HttpStatus.NOT_FOUND, "Cancellation reason not found [id=$cancellationReasonId]")
+  }
+}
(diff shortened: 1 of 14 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/controller/mappers/CancellationReasonMapper.kt
```diff
@@ -0,0 +1,18 @@
...
+import uk.gov.justice.digital.hmpps.hmppsinterventionsservice.jpa.entity.CancellationReason
+import uk.gov.justice.digital.hmpps.hmppsinterventionsservice.jpa.repository.CancellationReasonRepository
+
+@Component
+class CancellationReasonMapper(
+  private val cancellationReasonRepository: CancellationReasonRepository
+) {
+  fun mapCancellationReasonIdToCancellationReason(cancellationReasonId: String): CancellationReason {
+    return cancellationReasonRepository.findByIdOrNull(cancellationReasonId)
+      ?: throw ResponseStatusException(HttpStatus.NOT_FOUND, "Cancellation reason not found [id=$cancellationReasonId]")
```
tomsmyers, line 16: i'm not sure 404 is the appropriate response here. if we raised `EntityNotFound` we would also return 404. but i think bad request might be better. i.e. the cancellation code is invalid

tomsmyers: or have this method raise entitynotfound, and have the controller catch it and return bad request

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/dto/EndedReferralDTO.kt
```diff
@@ -21,6 +22,7 @@ class EndedReferralDTO(
         referenceNumber = referral.referenceNumber!!,
         assignedTo = referral.assignedTo?.let { AuthUserDTO.from(it) },
         referral = DraftReferralDTO.from(referral),
+        cancellationReason = referral.cancellationReason!!.id
```
tomsmyers, line 25: is the id the best thing to include in this response? i feel like the string representation would be more useful.

tomsmyers (commented):
couple small things

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/dto/EndedReferralDTO.kt
rgforsyth: Just to clarify. The 'id' in this case is the 3 letter string code. ('MIS') for example. The alternative would be providing the description, or both?

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/dto/EndedReferralDTO.kt
```diff
@@ -21,6 +24,9 @@ class EndedReferralDTO(
         referenceNumber = referral.referenceNumber!!,
         assignedTo = referral.assignedTo?.let { AuthUserDTO.from(it) },
         referral = DraftReferralDTO.from(referral),
+        cancellationReason = referral.cancellationReason!!.description,
```
tomsmyers, line 27: please can we leave these as nullable for now? i'm not sure that ended referrals will always have cancellation reasons and comments

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/dto/EndReferralDTO.kt
```diff
@@ -0,0 +1,6 @@
+package uk.gov.justice.digital.hmpps.hmppsinterventionsservice.dto
+
+data class EndReferralDTO(
+  val cancellationReasonCode: String,
+  val cancellationComments: String,
```
tomsmyers, line 5: please can we make this optional?

tomsmyers (approved):
awesome- thanks

PR #200, opened 2021-04-21
send notify email on end of service report submission

rgforsyth:

Adds ability to send email notification to pp on when submitting an end of service report
Updates/adds factory classes for test data.

Allows a pp to be notified by email when an end of service report has been submitted.


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/events/EndOfServiceReportEventPublisher.kt
```diff
@@ -0,0 +1,47 @@
...
+import uk.gov.justice.digital.hmpps.hmppsinterventionsservice.jpa.entity.EndOfServiceReport
+
+enum class EndOfServiceReportEventType {
+  SUBMITTED,
+}
+
+class EndOfServiceReportEvent(
+  source: Any,
+  val type: EndOfServiceReportEventType,
+  val endOfServiceReport: EndOfServiceReport,
```
tomsmyers, line 18: the EOSR now has a reference to the referral in it, so we can remove the `referralReference` and `ppUser` fields from this even constructor. in the listener class we can pull them out of the referral.

tomsmyers (changes requested):
the comment on here about getting the referral directly from the eosr entity simplifies a lot of these changes. we can get rid of the repo method, remove the lookup for referral etc.

tomsmyers (approved):
nice - thanks! 

ideally this would be squashed into two commits, one to fix the JPA bug, and one with the implementation of the email notification. but not a blocker.

PR #202, opened 2021-04-23
DAPI-170 Change to url for community-api create appointments endpoint

mahalrmoj:
Community-api has Introduced the distinction between a well known client and a none well known client. Each has its own endpoint for create appointment which shares same service implementation. For the well known client like interventions, defaults/preprocessing are provided by community-api:
1) find a requirement for the service provided
2) use defaults as identified by the integration context

This enables community-api to support consumers with differing needs whilst sharing the same implementation.


tomsmyers (approved):
cool

mahalrmoj:
Community-api approach has now changed

PR #203, opened 2021-04-23
IC-1571 create new generic referral DTO and endpoint 

tomsmyers:

creates a new referral DTO type that can get used for a generic 'get referral by id' endpoint. 

creates a new service category DTO type that contains just the id and name.

creates the new 'get referral by id' endpoint. 

allow the UI to GET a referral without knowing if it's draft, sent, or ended


src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/dto/ReferralDTOTest.kt
```diff
@@ -0,0 +1,73 @@
...
+  val referralFactory = ReferralFactory()
+
+  @Nested
+  inner class EndedReferral {
+    @Test
+    fun `endedReferralFields are populated when referral has endedAt`() {
+      val endedReferral = referralFactory.createEnded()
+      val referralDTO = ReferralDTO.from(endedReferral)
+      assertThat(referralDTO.endedFields).isNotNull()
+      assertThat(referralDTO.endedFields!!.endedAt).isEqualTo(endedReferral.endedAt)
```
tomsmyers, line 19: here you can use `endedFields?.endedAt` which returns either the field or null instead of making the not null assertion.

tomsmyers: see https://kotlinlang.org/docs/null-safety.html#safe-calls

tomsmyers (commented):
i can't approve because i opened the PR, but i've review the tests @andrewburrowsmoj and they look really good 👍

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/dto/ServiceCategoryFullDTO.kt
```diff
@@ -21,15 +21,15 @@ data class ComplexityLevelDTO(
   val description: String,
 )
 
-data class ServiceCategoryDTO(
+data class ServiceCategoryFullDTO(
```
mahalrmoj, line 24: Could we rename 1) ServiceCategoryDTO as ServiceCategoryBaseDTO and 2) ServiceCategoryFullDTO as ServiceCategoryDTO, and also, inherit one from the other - but understand if it causes too much change

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/dto/ReferralDTO.kt
```diff
@@ -0,0 +1,104 @@
...
+
+      val referralDTO = ReferralDTO(
+        id = referral.id,
+        createdAt = referral.createdAt,
+        serviceUser = ServiceUserDTO.from(referral.serviceUserCRN, referral.serviceUserData),
+        serviceCategory = ServiceCategoryDTO.from(contract.serviceCategory),
+        serviceProvider = ServiceProviderDTO.from(contract.serviceProvider),
+
+        formFields = FormFields(
+          completionDeadline = referral.completionDeadline,
```
mahalrmoj, line 64: Would it be better to call these BasicAttributes from a domain model perspective rather than FormFields, as form implies the source?

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/util/ReferralFactory.kt
```diff
@@ -14,7 +14,7 @@ import java.util.UUID
 class ReferralFactory(em: TestEntityManager? = null) : EntityFactory(em) {
   private val authUserFactory = AuthUserFactory(em)
   private val interventionFactory = InterventionFactory(em)
-  private val calcellationReasonFactory = CancellationReasonFactory(em)
+  private val cancellationReasonFactory = CancellationReasonFactory(em)
```
mahalrmoj, line 17: Nice catch

mahalrmoj (approved):
LGTM only a couple of preferences more than anything else

PR #205, opened 2021-04-26
DAPI-170 Include Integration Context as path variable

mahalrmoj:
Comunity-api Introduced the distinction between a client that requires additional processing and and one that doesn't. Each has its own endpoint for create appointment which shares same service implementation. Where additional processing is required, an integration context name must be supplied.

@RequestMapping(value = "/offenders/crn/{crn}/sentence/{sentenceId}/appointments/context/{contextName}" ...

The additionally processing identified by contextName is in the form of

find a requirement for the sentence provided
use defaults as identified by the integration context

This change is to change the endpoint url called by interventions

This enables community-api to support consumers with differing needs whilst sharing the same implementation.


tomsmyers (approved):
👍

PR #210, opened 2021-04-29
IC-1171 Update Delius of outcome of Appointment (attendance and PP behaviour notification) 

mahalrmoj:
When attendance and session feedback are submitted, Delius must be notified of the outcome. This feature makes the request to community-api with the necessary data on the submit event. As part of the request, an appointment id is provided which is now being saved when the initial appointment is made.

This enables interventions to update Delius with the outcome of appointments with Service Users.


src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/events/AppointmentEventPublisherTest.kt
```diff
@@ -66,4 +68,28 @@ class AppointmentEventPublisherTest {
     Assertions.assertThat(event.detailUrl).isEqualTo("http://localhost/action-plan/123/appointments/1")
     Assertions.assertThat(event.notifyPP).isTrue
   }
+
+  @Test
+  fun `builds an appointment session feedback event and publishes it`() {
+    val actionPlan = SampleData.sampleActionPlan()
```
rgforsyth, line 74: Not blocking at all, but I think we were trying to move away from using SampleData and towards using the Factory helpers

rgforsyth (approved):
Looks good. Small comment about using SampleData rather than the Factories for creating test data.

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/events/AppointmentEventPublisherTest.kt
mahalrmoj: Thanks for the reminder

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/component/CommunityAPIClient.kt
```diff
@@ -25,7 +25,19 @@ class CommunityAPIClient(
       .subscribe()
   }
 
-  fun <T : Any> makeSyncPostRequest(uri: String, requestBody: Any, responseBodyClass: Class<T>): T? {
+  fun makeAsyncPatchRequest(uri: String, requestBody: Any) {
+    communityApiWebClient.patch().uri(uri)
+      .body(Mono.just(requestBody), requestBody::class.java)
+      .retrieve()
+      .bodyToMono(Unit::class.java)
```
tomsmyers, line 32: what does this line do?

tomsmyers: i see it above as well, but just want to remind myself

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/AppointmentsServiceTest.kt
```diff
@@ -139,6 +140,48 @@ internal class AppointmentsServiceTest {
 
     assertThat(updatedAppointment).isEqualTo(actionPlanAppointment)
     verify(communityAPIBookingService).book(actionPlanAppointment, appointmentTime, durationInMinutes)
+    verify(actionPlanAppointmentRepository).save(
+      ArgumentMatchers.argThat { (
```
tomsmyers, line 144: do we really need this explicit matcher? can't we just match `any()`? i find this code really hard to grep

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/CommunityAPIService.kt
```diff
@@ -48,6 +57,43 @@ class CommunityAPIService(
...
+    when (event.type) {
+      SESSION_FEEDBACK_RECORDED -> {
+        val url = UriComponentsBuilder.fromHttpUrl(interventionsUIBaseURL)
+          .path(interventionsUISessionFeedbackLocation)
+          .buildAndExpand(event.appointment.actionPlan.referral.id, event.appointment.sessionNumber)
+          .toString()
+
+        val jsonPatch = JsonPatch(
+          listOf(
+            ReplaceOperation(of("notes"), valueOf(url)),
```
tomsmyers, line 80: this is neat - is this a new thing you guys have implemented in community api?

tomsmyers: the json patch format that is

src/main/resources/application.yml
```diff
@@ -88,11 +88,13 @@ interventions-ui:
     view-appointment: "/probation-practitioner/referrals/{id}/progress"
     pp-referral-progress: "/probation-practitioner/referrals/{id}/progress"
     submit-end-of-service-report: "/probation-practitioner/end-of-service-report/{id}"
+    session-feedback: "/probation-practitioner/action-plan/{id}/appointment/{sessionNumber}/post-session-feedback"
 
 community-api:
   locations:
     sent-referral: "/secure/offenders/crn/{crn}/referral/sent"
     book-appointment: "/secure/offenders/crn/{crn}/sentence/{sentenceId}/appointments/context/{contextName}"
+    patch-appointment: "/secure/offenders/crn/{crn}/appointments/{appointmentId}/context/{contextName}"
```
tomsmyers, line 97: it's interesting that this URL doesn't have the sentence id in it. i guess now we have the appointment id we don't need it

tomsmyers (commented):
nice - just one or two questions around the need for the new event type

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/CommunityAPIService.kt
```diff
@@ -48,6 +57,43 @@ class CommunityAPIService(
...
+  @Value("\${interventions-ui.locations.session-feedback}") private val interventionsUISessionFeedbackLocation: String,
+  @Value("\${community-api.locations.patch-appointment}") private val communityAPIPatchAppointmentLocation: String,
+  @Value("\${community-api.integration-context}") private val integrationContext: String,
+  private val communityAPIClient: CommunityAPIClient,
+) : ApplicationListener<AppointmentEvent>, CommunityAPIService {
+  companion object : KLogging()
+
+  override fun onApplicationEvent(event: AppointmentEvent) {
+    when (event.type) {
+      SESSION_FEEDBACK_RECORDED -> {
```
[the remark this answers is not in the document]
mahalrmoj: Needed an event when I could be sure both the attended enum and the notifyPP flag were set. ATTENDANCE_RECORDED signifies the former and BEHAVIOUR_RECORDED signifies the latter. So I created an event that fires when I know both are done which I called SESSION_FEEDBACK_RECORDED which tallied with the action "submitSessionFeedback".

Will raise a tech debt item as suggested by Tom to collapse the former two into the last one with flags for notifyPPOfAttendance and notifyPPOfBehaviour

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/CommunityAPIService.kt
mahalrmoj: This is how the Delius-api is written to accept jsonPatch and so it was thought it would also work well for community-api too. However, Prisons teams are suggesting PUT/POST is much more informative.

src/main/resources/application.yml
mahalrmoj: Yes that's right

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/component/CommunityAPIClient.kt
mahalrmoj: My reasoning is that since the response type is Unit, then the response will not be decoded. One is provided by community-api and nothing is done with it.

makeSyncPostRequest has a bodyToMono(AppointmentCreateResponseDTO::class.java) which successfully decodes the response from which we get the delius_appointment_id)

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/AppointmentsServiceTest.kt
mahalrmoj: It was the only way I could confirm that the deliusAppointmentId was being persisted.

PR #216, opened 2021-05-05
In User Research - Disable appointment bookings in Delius

mahalrmoj:

Disable appointment booking feature in user-research 

Enable user-research to be unaffected by Delius


sldblog:
> Enable user-research to be unaffected by Delius

Because we expect the experiments to be repeated and the same appointment details to be reused -- and we don't want to stop the service working there with genuine scheduling conflicts, right?

sldblog (approved):

PR #217, opened 2021-05-05
Make application state observable from logs

sldblog:

Make more events visible on non-local environments by reducing the default log level to `INFO`.

Reducing the minimum log level to `INFO` makes us able to observe steps in the boot process, e.g. migrations, web listeners, etc.

That capability is required for us to have a sense of what is happening

### DIFF
--- a/src/main/resources/logback-spring.xml
+++ b/src/main/resources/logback-spring.xml
@@ -64,7 +64,7 @@
             </encoder>
         </appender>
 
-        <root level="warn">
+        <root level="info">
             <appender-ref ref="stdout" />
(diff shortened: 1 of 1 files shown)


src/main/resources/logback-spring.xml
```diff
@@ -64,7 +64,7 @@
             </encoder>
         </appender>
 
-        <root level="warn">
+        <root level="info">
```
sldblog, line 67: 🙋 This is in `<springProfile name="!local">`

mahalrmoj (approved):

PR #218, opened 2021-05-05
Changed command to  use the jsonl flow

green-digger:

Alters the commands run within the data extractor cronJob to use the jsonl flow.

To transfer data from the Interventions development environment in jsonl format to the Analytical Platform. This is to allow us to prepare our processes to handle jsonl instead of csv.

### DIFF
--- a/helm_deploy/hmpps-interventions-service/templates/cronjob-data-extractor-analytics.yaml
+++ b/helm_deploy/hmpps-interventions-service/templates/cronjob-data-extractor-analytics.yaml
@@ -12,6 +12,7 @@ spec:
           containers:
           - name: data-extractor-analytics
             image: ministryofjustice/data-engineering-data-extractor:develop
+            imagePullPolicy: Always
             {{ if (and .Values.env_details.contains_live_data (not .data_protection_impact_assessment_signed_off)) }}
             command: ["sh", "-c", "psql -c 'SELECT 1' && mkdir -p 'export/csv' && touch 'export/csv/signed_off_DPIA_required.csv' && transfer_local_to_s3.sh"]
             {{ else }}
--- a/helm_deploy/hmpps-interventions-service/templates/cronjob-data-extractor-analytics.yaml
+++ b/helm_deploy/hmpps-interventions-service/templates/cronjob-data-extractor-analytics.yaml
@@ -21,6 +21,8 @@ spec:
             env:
               - name: LOCAL_EXPORT_DESTINATION
                 value: export
+              - name: PGPORT
+                value: 5432
               - name: PGHOST
                 valueFrom:
                   secretKeyRef:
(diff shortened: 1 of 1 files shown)


helm_deploy/hmpps-interventions-service/templates/cronjob-data-extractor-analytics.yaml
```diff
@@ -15,7 +15,7 @@ spec:
             {{ if (and .Values.env_details.contains_live_data (not .data_protection_impact_assessment_signed_off)) }}
             command: ["sh", "-c", "psql -c 'SELECT 1' && mkdir -p 'export/csv' && touch 'export/csv/signed_off_DPIA_required.csv' && transfer_local_to_s3.sh"]
             {{ else }}
-            command: ["sh", "-c", "extract_psql_all_tables_to_csv.sh && transfer_local_to_s3.sh"]
+            command: ["sh", "-c", "python3 /data-engineering-data-extractor/modules/extract_table_names.py && python3 /data-engineering-data-extractor/modules/extract_pg_jsonl_snapshot.py && transfer_local_to_s3.sh"]
```
sldblog, line 18: _nit_
Could Python files as `/data-engineering-data-extractor/modules/extract_table_names.py` have an exec flag and a shebang?

As the `/data-engineering-data-extractor/modules` dir is already in the `$PATH`, the above would allow us to call the chain as
```
extract_table_names.py && extract_pg_jsonl_snapshot.py && transfer_local_to_s3.sh
```

sldblog (approved):

sldblog (changes requested):
🙋‍♂️ When I try this, the command errors:
```
$ k logs pod/data-extractor-analytics-test-cmt9p
Traceback (most recent call last):
  File "/data-engineering-data-extractor/modules/extract_table_names.py", line 52, in <module>
    extract_table_names()
  File "/data-engineering-data-extractor/modules/extract_table_names.py", line 12, in extract_table_names
    db_details = pg.get_database_details_from_env()
  File "/data-engineering-data-extractor/modules/postgres_functions.py", line 10, in get_database_details_from_env
    'port': os.getenv('PGPORT').strip(),
AttributeError: 'NoneType' object has no attribute 'strip'
```

In my interpretation, either the code needs to allow for default ports, or `PGPORT` has to be added to the `CronJob` config

Once that is resolved, it looks like `jsonl` files are excluded by the S3 transfer script:
```
Data export finished (jsonl)
Starting Copy to S3
COMPLETED
```

Looking at https://github.com/ministryofjustice/data-engineering-data-extractor/blob/b791ed306003669d89e8ba456df90665e2fd16d2/modules/transfer_local_to_s3.sh this shouldn't be that case, so an `imagePullPolicy: "Always"` is probably required on our jobs as Kubernetes by default will not pull an image down if the tag already exists

----

To summarise,

Could you add `PGPORT` env value and `imagePullPolicy` to the CronJob, please?

helm_deploy/hmpps-interventions-service/templates/cronjob-data-extractor-analytics.yaml
green-digger: Agree. This will be done but with the next set of changes :)

sldblog (approved):

PR #221, opened 2021-05-10
DAPI-172 Updates to the endpoint for referral sent (start)

mahalrmoj:
After a review of the interaction with Interventions, a number of changes where identified for the referral sent end point:

1) A more generic name for the endpoint is ReferralStart
2) ContractType is to be used to map to NsiType instead of ServiceCategory
3) SentAt is renamed to StartedAt
4) Context name is to be specified in URL as apposed to in the body - in line with appointment creation

This change means that interventions new contract type change will be supportable and the endpoint is using the agreed scheme for well-known clients.

**Breaking Changes?**
Yes and limited to interventions referralSent endpoint.


tomsmyers (approved):
thanks! do we have a ticket open to fix the contract type code?

PR #225, opened 2021-05-11
add subcontractors and rename prime provider

rgforsyth:

Renames service provider id to prime provider id in the db. Allows subcontractors to be stored against a contract.

Allow subcontractors to be recorded against a contract


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/jpa/entity/DynamicFrameworkContract.kt
```diff
@@ -15,7 +17,7 @@ data class DynamicFrameworkContract(
   @NotNull @ManyToOne @JoinColumn(name = "service_category_id")
   val serviceCategory: ServiceCategory,
 
-  @NotNull @ManyToOne @JoinColumn(name = "service_provider_id")
+  @NotNull @ManyToOne @JoinColumn(name = "prime_provider_id")
   val serviceProvider: ServiceProvider,
```
tomsmyers, line 21: please can we also refactor this field name to `primeProvider`? thanks

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/util/DynamicFrameworkContractFactory.kt
```diff
@@ -27,6 +27,7 @@ class DynamicFrameworkContractFactory(em: TestEntityManager? = null) : EntityFac
     npsRegion: NPSRegion? = null,
     pccRegion: PCCRegion? = null,
     contractReference: String = RandomStringUtils.randomAlphanumeric(8),
+    subcontractor: Set<ServiceProvider>? = null,
```
tomsmyers, line 30: nit: most of these classes set the default params in the signature, instead of lower down. this means the type doesn't have to be nullable. e.g. contractReference above.

tomsmyers (approved):
thanks!

PR #226, opened 2021-05-12
new modular authorization components which can be stacked to build access controls

tomsmyers:

A handful of self-contained auth components designed to be combined to create access controls for referral endpoints.

None of the endpoints actually have any of these restrictions yet, this PR just contains bits of logic to determine if access to a particular referral is allowed.

get started on having an access framework in place to build on top of


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/HMPPSAuthService.kt
```diff
@@ -45,6 +45,19 @@ class HMPPSAuthService(
...
   private val hmppsAuthApiWebClient: WebClient,
 ) {
+  fun getUserGroups(user: AuthUser): List<AuthGroupID>? {
+    val url = UriComponentsBuilder.fromPath(authUserGroupsLocation)
+      .buildAndExpand(user.userName)
+      .toString()
+
+    return hmppsAuthApiWebClient.get().uri(url)
+      .retrieve()
+      .onStatus({ HttpStatus.NOT_FOUND == it }, { Mono.just(null) })
```
tomsmyers, line 55: what do you think of this method returning a nullable list vs throwing an error?

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/authorization/UserAccessScopeMapper.kt
```diff
@@ -0,0 +1,51 @@
...
+  private val serviceProviderRepository: ServiceProviderRepository,
+  private val dynamicFrameworkContractRepository: DynamicFrameworkContractRepository,
+) {
+  private val serviceProviderGroupPrefix = "INT_SP_"
+  private val contractGroupPrefix = "INT_CG_"
+
+  // get the user access scope for a given service provider user
+  fun forServiceProviderUser(user: AuthUser): UserAccessScope {
+    val groups = hmppsAuthService.getUserGroups(user)
+      ?: throw RuntimeException("pp or non existent user")
```
tomsmyers, line 30: these exceptions are not ideal - i think we should write some new exception classes and handlers for auth errors. similarly for the `ServerWebInputExceptions` below

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/authorization/EligibleProviderMapper.kt
```diff
@@ -0,0 +1,22 @@
+package uk.gov.justice.digital.hmpps.hmppsinterventionsservice.authorization
+
+import org.springframework.stereotype.Component
+import uk.gov.justice.digital.hmpps.hmppsinterventionsservice.jpa.entity.Referral
+import uk.gov.justice.digital.hmpps.hmppsinterventionsservice.jpa.entity.ServiceProvider
+
+@Component
+class EligibleProviderMapper {
```
tomsmyers, line 8: i'm not super keen on the naming of some of these components. suggestions welcome.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/authorization/UserAccessScopeMapper.kt
```diff
@@ -0,0 +1,51 @@
...
+    if (serviceProviderGroups.size != 1) {
+      throw RuntimeException("there must only be one service provider group per user")
+    }
+
+    val contractGroups = groups
+      .filter { it.startsWith(contractGroupPrefix) }
+      .map { it.removePrefix(contractGroupPrefix) }
+
+    val serviceProvider = serviceProviderRepository.findByIdOrNull(serviceProviderGroups[0])
+      ?: throw RuntimeException("user does not exist in our database")
```
tomsmyers, line 45: this raised a few questions in my head about under what circumstances users are persisted. 

currently when:

 - an appointment is created by an SP user
 - an action plan is created or submitted by an SP user
 - an EOSR is created or submitted by an SP user
 - a referral is assigned to a caseworker by an SP user
 - a referral is assigned to an SP user
 - a referral end is requested by a PP user
 - a referral is created or sent by a PP user
 
 are there circumstances where a user may be trying to access a referral but has never been saved in the db by one of the actions listed above? should we persist all users when they log in to the service too?

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/authorization/UserMapper.kt
```diff
@@ -20,6 +17,7 @@ class JwtAuthUserMapper {
     val authSource = authentication.token.getClaimAsString("auth_source")
       ?: throw ServerWebInputException("no 'auth_source' claim in authentication token")
 
+    // should we persist the user here??
     return AuthUser(id = userID, authSource = authSource, userName = userName)
```
sldblog, line 21: Unless we want to cache everyone, I'd say no

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/authorization/ReferralAccessChecker.kt
```diff
@@ -0,0 +1,29 @@
...
+
+import uk.gov.justice.digital.hmpps.hmppsinterventionsservice.jpa.entity.AuthUser
+import uk.gov.justice.digital.hmpps.hmppsinterventionsservice.jpa.entity.Referral
+
+class ReferralAccessChecker(
+  private val userTypeChecker: UserTypeChecker,
+  private val eligibleProviderMapper: EligibleProviderMapper,
+  private val userAccessScopeMapper: UserAccessScopeMapper,
+) {
+  fun forServiceProviderUser(referral: Referral, user: AuthUser): Boolean {
```
andrewburrowsmoj, line 11: serviceProviderUserCanAccess?

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/authorization/UserMapper.kt
tomsmyers: thanks - yeh i agree. as i was writing this i went down a path of thinking we needed users to be persisted for scope mappings, but of course we don't. so this comment should be removed.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/authorization/UserAccessScopeMapper.kt
sldblog: In my view, this is data at this point, becoming an exception one level up

We could use the output of this function as a way to display the problems with the user:
```
scope = forServiceProviderUser(user)
scope.hasProblems() -> throw AuthException
```
where `scope.problems` could be
```
[
  "user is not a provider user",
  "provider with id WEIRD does not exist",
  "contract with reference WOO does not exist",
  "contract C2 belongs to a different provider"
]
```
etc

Treating problems as data would allow us to test all scenarios and not worry about the shortcut paths. (And possibly to serve content even if there's a minor misconfiguration)

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/authorization/ReferralAccessChecker.kt
sldblog, line 11: Is this class/function used somewhere yet? I couldn't find it

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/authorization/UserAccessScopeMapper.kt
```diff
@@ -0,0 +1,51 @@
...
+class UserAccessScopeMapper(
+  private val hmppsAuthService: HMPPSAuthService,
+  private val serviceProviderRepository: ServiceProviderRepository,
+  private val dynamicFrameworkContractRepository: DynamicFrameworkContractRepository,
+) {
+  private val serviceProviderGroupPrefix = "INT_SP_"
+  private val contractGroupPrefix = "INT_CG_"
+
+  // get the user access scope for a given service provider user
+  fun forServiceProviderUser(user: AuthUser): UserAccessScope {
```
sldblog, line 28: What do you think about `validatedScopeFor(user)`?

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/authorization/UserMapper.kt
```diff
@@ -1,16 +1,13 @@
-package uk.gov.justice.digital.hmpps.hmppsinterventionsservice.controller.mappers
+package uk.gov.justice.digital.hmpps.hmppsinterventionsservice.authorization
```
sldblog, line 1: 🇬🇧 / 🇺🇸 ? :trollface:

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/authorization/EligibleProviderMapper.kt
```diff
@@ -0,0 +1,22 @@
...
+  // get the service providers who are eligible to access this referral
+  fun fromReferral(referral: Referral): List<ServiceProvider> {
+    return listOf(getPrimeProvider(referral)) + getSubcontractorProviders(referral)
+  }
+
+  private fun getPrimeProvider(referral: Referral): ServiceProvider {
+    return referral.intervention.dynamicFrameworkContract.primeProvider
+  }
+
+  private fun getSubcontractorProviders(referral: Referral): List<ServiceProvider> {
```
sldblog, line 18: Is there a need to differentiate between prime and subcontractor providers at this level?

Can we have a `providersWithAccessTo(referral): List<ServiceProvider>` call that hides this complexity here? After all, the main check is "can the user's provider have access to this" -- the caller shouldn't have to check for prime, then subs, etc.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/authorization/ReferralAccessChecker.kt
```diff
@@ -0,0 +1,29 @@
...
+    if (!userTypeChecker.isServiceProviderUser(user)) {
+      return false
+    }
+
+    val userScope = userAccessScopeMapper.forServiceProviderUser(user)
+    val eligibleServiceProviders = eligibleProviderMapper.fromReferral(referral)
+
+    if (!eligibleServiceProviders.contains(userScope.serviceProvider)) {
```
sldblog, line 19: Kotlin ignorance question: is object equality here well defined enough for the `contains` check?

sldblog (commented):
Looks nice, couple of questions :)

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/authorization/ReferralAccessChecker.kt
tomsmyers: no it's not - yet. i mention this in the PR description. andrew and ryan are working on integration tests to allow us to roll this out across the controllers.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/authorization/UserMapper.kt
tomsmyers: yeh this is my bad. however i've not trained myself to really get out of this habit since i find most libs use american spellings anyway. e.g. https://docs.spring.io/spring-security/site/docs/5.2.x/reference/html/authorization.html

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/authorization/UserAccessScopeMapper.kt
tomsmyers: would you keep the name of the class as it?

`userAccessScopeMapper.validatedScopeFor` ?

i think it's important here to distinguish between sp and pp users ( since pp users have no concept of an access scope).

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/authorization/EligibleProviderMapper.kt
tomsmyers: this is a private method. intentionally private to hide this complexity.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/authorization/ReferralAccessChecker.kt
tomsmyers: i made a comment about this later. it does work, but i want to dig into exactly how.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/authorization/ReferralAccessChecker.kt
```diff
@@ -0,0 +1,29 @@
+package uk.gov.justice.digital.hmpps.hmppsinterventionsservice.authorization
+
+import uk.gov.justice.digital.hmpps.hmppsinterventionsservice.jpa.entity.AuthUser
+import uk.gov.justice.digital.hmpps.hmppsinterventionsservice.jpa.entity.Referral
+
+class ReferralAccessChecker(
```
sldblog, line 6: 🤔 I'm wondering if this has to be an application layer concern.

Does it make sense to say: we only get back objects from the database that are valid for the user's validated scope?

I.e. should we have `findById(id)` and `findByIdAndScope(id, userScope)`; and similar for the list lookups? We then could construct a query that effectively implements `eligibleProviderMapper` as filter conditions.

The net benefit of this approach could be: application code could rely that it always gets properly scoped data, unless I'm wrong in my reasoning here

tomsmyers: i'd be happy to go down this route. it needs more thinking though. i don't think we should be blocked on getting some kind of implementation rolled out in the short term.

e.g. with this idea, how do we differentiate between referral ids that don't exist and access being restricted in the application layer?

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/authorization/UserAccessScopeMapper.kt
sldblog: 👍 `providerScopeMapper` maybe?

Should we have a different `AuthUser` class for SPs and PPs to make this explicit?

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/authorization/UserMapper.kt
sldblog: 😁 totally! I was just trolling

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/authorization/EligibleProviderMapper.kt
sldblog: I completely missed that :) 👍

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/authorization/UserAccessScopeMapper.kt
tomsmyers: ok i've put together something along these lines. 

to stop the caller of this from explicitly having to check for config errors (easily forgotten), i make this class raise an exception, but with all possible configuration errors included.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/authorization/UserAccessScopeMapper.kt
tomsmyers: after our discussion we agreed to rename the whole component to be SP specific. i've made this change now.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/authorization/ReferralAccessChecker.kt
tomsmyers: after a lengthy discussion about this we determined that utilizing the components here will well as long as the implementation is 'hidden' in the service layer. 

i.e. controllers still call 'user scoped' service layer methods like "get sent referral for user", which handle all this complexity.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/authorization/ReferralAccessChecker.kt
tomsmyers: made a ticket for this

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/authorization/ServiceProviderAccessScopeMapper.kt
```diff
@@ -0,0 +1,75 @@
...
+    }
+
+    val contracts = dynamicFrameworkContractRepository.findAllByContractReferenceIn(contractGroups)
+
+    if (contracts.size != contractGroups.size) {
+      configErrors.add("at least one contract group does not exist in the interventions database")
+    }
+
+    if (configErrors.isNotEmpty()) {
+      throw AccessError(errorMessage, configErrors)
```
sldblog, line 70: _not a blocker_
For config problems, we could still allow providers to get their "validated scope" results

This might become crucial if contract references get out of sync for some accidental reason

sldblog (approved):

PR #227, opened 2021-05-13
IC-1690 implement access controls for referral endpoints that return lists

tomsmyers:

a major change to the `GET /sent-referrals` endpoint which calculates the list of referrals to return based on contract scopes for SP users.

this will required contract changes, and UI changes to function. 

for now it's in draft until we iron out any issues.

restrict SP referral access based on contract scopes


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/authorization/ReferralAccessFilter.kt
```diff
@@ -6,15 +6,9 @@ import uk.gov.justice.digital.hmpps.hmppsinterventionsservice.jpa.entity.Referra
 
 @Component
 class ReferralAccessFilter(
-  private val userTypeChecker: UserTypeChecker,
   private val userAccessScopeMapper: UserAccessScopeMapper,
 ) {
   fun serviceProviderReferrals(referrals: List<Referral>, user: AuthUser): List<Referral> {
-    if (!userTypeChecker.isServiceProviderUser(user)) {
```
tomsmyers: i think it's best to remove this from here. this method is filtering referrals for SP users, so we don't need to check if this is an SP user or not. it's up to the caller to handle that side of things.

andrewburrowsmoj (approved):

PR #228, opened 2021-05-13
IC-1699 add access controls to all 'single referral' endpoints 

andrewburrowsmoj:

Adds referral access restrictions for all endpoints which relate to a single referral resource.

**Please review these changes commit by commit to see the functional changes seperated from the test changes.**

Ensure that only permitted users are able to access referral details.


mahalrmoj (commented):
From my reading of this, the endpoints that work on action plans and appointment (via their ids) are not secured from a referral viewpoint, i.e. we don't check their owning referral. Sorry if I have missed it.

Do we need to do anything special to run locally now?

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/authorization/ServiceProviderAccessScopeMapper.kt
```diff
@@ -40,36 +40,51 @@ class ServiceProviderAccessScopeMapper(
...
   }
+
+  private fun getServiceProvider(serviceProviderGroups: List<String>, configErrors: MutableList<String>): ServiceProvider? {
+    return when {
+      serviceProviderGroups.isEmpty() -> {
+        configErrors.add("no service provider groups associated with user")
+        null
+      }
+      serviceProviderGroups.size > 1 -> {
+        configErrors.add("more than one service provider group associated with user")
```
andrewburrowsmoj, line 65: More of a question for me... I suppose it's not known how hmpps auth is set up in prod but do we know for certain that users will only ever be set up for a single service provider auth group? 
Is it possible that hmpps auth is not entirely clean - maybe on renaming of an SP they don't delete the old? I know it's not our responsibility - but is there some way to know about the cleanliness of hmpps auth before prod live to reduce issues arising?

andrewburrowsmoj (commented):
Looks great. Love the changes that's been done for the Integration tests.
(I can't approve my own PR)

Would it be useful to add some logging for auth failures? 
This may be useful for support to diagnose issues.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/authorization/ServiceProviderAccessScopeMapper.kt
tomsmyers: great question - this is a constraint that we are in charge of. since we have no idea about which is the correct provider in the case where there is more than one, we have to enforce this rule.

i'm sure we could run queries against auth to validate that there is only one INT_SP_ group per user, but really this is more of a business rule around 'onboarding' SP users into auth. we can raise this with sarah smith to discuss further.

tomsmyers:
> Looks great. Love the changes that's been done for the Integration tests.
> 
> Would it be useful to add some logging for auth failures?
> This may be useful for support to diagnose issues.
> Otherwise they would only have the reason for the failure but not the specifics: i.e. the name of the group in the database vs the name of the group in hmpps auth
> 
> (I can't approve my own PR)

@andrewburrowsmoj like the idea of adding additional logging. i'll open a ticket to get this done.

this would include details about the required vs actual group memberships etc i presume?

tomsmyers:
> LGTM
> 
> From my reading of this, the endpoints that work on action plans and appointment (via their ids) are not secured from a referral viewpoint, i.e. we don't check their owning referral. Sorry if I have missed it. One option may be to always have the referralId present on all urls and therefore it can always be checked.
> 
> Do we need to do anything special to run locally now?

@rajivmahal1 you are correct about that. let's discuss this in a morning meeting and determine the priority/need to get it done.

tomsmyers (approved):

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/integration/authorization/SingleReferralEndpoints.kt
```diff
@@ -0,0 +1,349 @@
...
+import org.springframework.boot.test.mock.mockito.MockBean
+import uk.gov.justice.digital.hmpps.hmppsinterventionsservice.integration.IntegrationTestBase
+import uk.gov.justice.digital.hmpps.hmppsinterventionsservice.jpa.entity.AuthUser
+import uk.gov.justice.digital.hmpps.hmppsinterventionsservice.jpa.entity.DynamicFrameworkContract
+import uk.gov.justice.digital.hmpps.hmppsinterventionsservice.jpa.entity.Referral
+import uk.gov.justice.digital.hmpps.hmppsinterventionsservice.service.HMPPSAuthService
+import uk.gov.justice.digital.hmpps.hmppsinterventionsservice.util.JwtTokenFactory
+import java.util.UUID
+
+class SingleReferralEndpoints : IntegrationTestBase() {
```
rgforsyth, line 16: Nice job on all these scenarios!

rgforsyth (approved):

PR #229, opened 2021-05-13
Create nightly policy scan reports with Veracode

sldblog:

Adds daily veracode scanning ([link to reports](https://analysiscenter.veracode.com/auth/index.jsp#ViewReportsResultSummary:77328:1140482:12058273:12032662:12048325:3225960))

We are using the "policy scan" product, which doesn't support concurrent reports, so it's not feasible to run on every commit

As this is purely informational at this point, a daily report is useful. We may switch to blocking "pipeline scans" soon

Using the continuous security tooling currently being tested/rolled out in Ministry of Justice

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -164,7 +164,7 @@ jobs:
           keys:
             - gradle-{{ checksum "build.gradle.kts" }}
             - gradle-
-      - run: ./gradlew assemble --console=plain
+      - run: ./gradlew assemble
       - persist_to_workspace:
           root: build/libs
           paths:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -13,6 +13,7 @@ orbs:
   kubernetes: circleci/kubernetes@0.11.2
   mem: circleci/rememborb@0.0.1
   snyk: snyk/snyk@0.0.12
+  veracode: orb-01scan/sast-orb@0.0.9
 
 _snyk_options: &snyk_options
   snyk-scan: true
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -13,6 +13,7 @@ orbs:
   kubernetes: circleci/kubernetes@0.11.2
   mem: circleci/rememborb@0.0.1
   snyk: snyk/snyk@0.0.12
+  veracode: orb-01scan/sast-orb@0.0.9
```
tomsmyers, line 16: weird import path eh?

.circleci/config.yml
```diff
@@ -155,6 +156,30 @@ jobs:
...
 
+  assemble:
+    executor: hmpps/java
+    steps:
+      - checkout
+      - restore_cache:
+          keys:
+            - gradle-{{ checksum "build.gradle.kts" }}
+            - gradle-
+      - run: ./gradlew assemble --console=plain
```
tomsmyers, line 167: what does --console=plain do?

tomsmyers (approved):

.circleci/config.yml
sldblog: Yeah. Its support for CircleCI started recently, so it's a spike I think

.circleci/config.yml
sldblog: Ah, it's a leftover (I got into the habit of using it instead of the fancy outputs)

It "disables all color and other rich output in the console output. This is the default when Gradle is not attached to a terminal."

As it's the default on CircleCI, I'll remove it

sldblog: Removed in 9c97a137bae96a32e9f71e51d86810087fcb877b

PR #236, opened 2021-05-18
IC-1736 change contract group prefix from 'INT_CG_' to 'INT_CR_'

tomsmyers:

change contract group prefix from 'INT_CG_' to 'INT_CR_'

make it more explicit that the contract group in auth should contain the contract reference number


mahalrmoj (approved):
LGTM

PR #238, opened 2021-05-18
IC-1709 do not allow an action plan to be created for a referral that already has one

tomsmyers:

 do not allow an action plan to be created for a referral that already has one

it's possible to do this when you open the referral progress page in two tabs at once. if you get into this state it causes sql consistency issues and breaks the referral entirely.


mahalrmoj (commented):
As a business rule should it not be implemented in the service tier and should it be guaranteed via a db unique index?

mahalrmoj (approved):

PR #239, opened 2021-05-18
Fix UR env by temporarily removing the V100.X migrations 

tomsmyers:

- Fix UR env by temporarily removing the V100.X migrations 
- fixes the existing migration generator script to use the latest schema. 

deploy to the UR env without cleaning the database

### DIFF
--- a/src/main/resources/application-research.yml
+++ b/src/main/resources/application-research.yml
@@ -1,3 +1,3 @@
 spring:
   flyway:
-    locations: classpath:db/migration,classpath:db/research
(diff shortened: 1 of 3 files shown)


sldblog (approved):
Good to have these in separate commits for later
LGTM 👍

sldblog:
We could maybe also explore alternatives to schema migrations for database "seeding", if you'd like

sldblog:
> deploy to the UR env without cleaning the database

I think it would be useful for posterity to explain why this is needed

PR #244, opened 2021-05-20
use contract type to generate reference number

tomsmyers:

use contract type to generate reference number

the reference number generator _used to use_ two letter abbreviations of the service category previously e.g. accommodation = AC. now that contracts are defined against contract types, these should be used instead e.g. women's services = WO.


rgforsyth (approved):
Thanks

PR #246, opened 2021-05-21
WIP allow selecting complexity level  by service category

tomsmyers:

_Required._

_Required._


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/ReferralService.kt
```diff
@@ -273,6 +264,30 @@ class ReferralService(
...
+      errors.add(FieldError(field = "complexityLevel", error = Code.INVALID_COMPLEXITY_LEVEL_FOR_SERVICE_CATEGORY))
+    }
+
+    if (errors.isNotEmpty()) {
+      throw ValidationError("draft referral update invalid", errors)
+    }
+
+    referral.complexityLevelIds[serviceCategoryId] = complexityLevelId
+    return referralRepository.save(referral)
+  }
```
andrewburrowsmoj, line 289: should also check that the service category is valid for the contract type too?

tomsmyers: hm. i think this validation lives in the method that allows setting of the service category. i think here we can be happy that the service categories which have been selected are valid.

PR #247, opened 2021-05-21
IC-1718 Notify Action Plan Submitted and tidy up Notes associated to Events

mahalrmoj:
This PR covers the changes to notify Delius of an Action Plan being submitted for approval. Also, to provide actual Contract Types for mapping to NSI Types in community-api.

To provide a trace of significant events for a Service User in Delius.


src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/CommunityAPIActionPlanServiceTest.kt
```diff
@@ -0,0 +1,66 @@
...
+    )
+  }
+
+  private fun getEvent(
+    ActionPlanEventType: ActionPlanEventType
+  ): ActionPlanEvent =
+    ActionPlanEvent(
+      "source",
+      ActionPlanEventType,
+      SampleData.sampleActionPlan(
```
rgforsyth, line 53: Not a blocker, but do we have an actionPlan Factory that we could use instead?

rgforsyth (approved):
LGTM 👍

PR #248, opened 2021-05-21
IC-1747 allow setting of complexity levels on a per service category basis

tomsmyers:

implements the endpoint to allow setting complexity level on a referral for a particular service category

make cohort referrals


src/main/resources/db/migration/V1_52__cohort_complexity_levels.sql
```diff
@@ -0,0 +1,7 @@
+create table referral_complexity_level_ids (
+    referral_id uuid not null,
+    complexity_level_ids uuid,
```
andrewburrowsmoj, line 3: should it not be `complexity_level_id` as there would only ever be one selected?

tomsmyers: this is the default jpa mapping that means we don't have to add any additional annotations on the entity class. i didn't think too hard about the column naming here.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/dto/SetComplexityLevelRequestDTO.kt
```diff
@@ -0,0 +1,8 @@
+package uk.gov.justice.digital.hmpps.hmppsinterventionsservice.dto
+
+import java.util.UUID
+
+data class SetComplexityLevelRequestDTO(
```
andrewburrowsmoj, line 5: Can perhaps combine ReferralComplexityLevel and SetComplexityLevelRequestDTO into a single DTO if making the assumption that DraftReferralDTO is the REST resource that is being modified.

src/main/resources/db/migration/V1_52__cohort_complexity_levels.sql
```diff
@@ -0,0 +1,16 @@
...
+    complexity_level_ids_key uuid not null,
+    constraint fk_referral_complexity_level_ids_referral_id foreign key (referral_id) references referral,
+    constraint fk_referral_complexity_level_ids_complexity_level_ids_key foreign key (referral_id, complexity_level_ids_key) references referral_selected_service_category,
+    primary key (referral_id, complexity_level_ids_key)
+);
+
+insert into referral_complexity_level_ids
+    select referral.id, complexity_level.id, service_category.id from referral, complexity_level, service_category
+    where referral.complexity_levelid = complexity_level.id
+    and complexity_level.service_category_id = service_category.id;
```
andrewburrowsmoj, line 13: I think the sql scripts in /src/main/resources/db/local/ will need changing to as the column is now dropped

src/main/resources/db/migration/V1_52__cohort_complexity_levels.sql
andrewburrowsmoj: Ah ok. I can foresee that it might be confusing if anyone ever needs to access the database (more with the ids_key than having _ids).

Not sure if support would have readonly access to database to perform any queries?

src/main/resources/db/migration/V1_52__cohort_complexity_levels.sql
tomsmyers: oh yeh, thanks!

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/dto/SetComplexityLevelRequestDTO.kt
tomsmyers: the dto would end up with both fields being nullable though i think? how would you structure it?

andrewburrowsmoj: Oh I'm not sure what you mean...

It's a really minor point...

I was just thinking you could rename ReferralComplexityLevel to ReferralComplexityLevelDTO, and use this as the REST resource that is being PATCHed rather than a separate DTO object.

andrewburrowsmoj (approved):

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/dto/SetComplexityLevelRequestDTO.kt
andrewburrowsmoj: Sorry, you're right, as it's a list rather than a direct object.

PR #249, opened 2021-05-21
IC-1682 patch service categories

rgforsyth:

Allows service categories to be patched on the referral. This prevents service categories being patched that are not linked to the contract type of the referral.

To allow service categories to be added to the referral


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/ReferralService.kt
```diff
@@ -204,6 +207,13 @@ class ReferralService(
       }
     }
 
+    update.serviceCategoryIds?.let {
+      val serviceCategories = serviceCategoryRepository.findByIdIn(it)
+      if (!referral.intervention.dynamicFrameworkContract.contractType.serviceCategories.containsAll(serviceCategories)) {
+        errors.add(FieldError(field = "serviceCategories", error = Code.INVALID_SERVICE_CATEGORY_FOR_CONTRACT))
+      }
+    }
+
```
andrewburrowsmoj, line 216: I don't think its necessary to fetch the serviceCategories since you can just to the comparison on the ids directly:

referral.intervention.dynamicFrameworkContract.contractType.serviceCategories.map{it.id}.containsAll(serviceCategoryIds)

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/ReferralService.kt
```diff
@@ -204,6 +207,13 @@ class ReferralService(
       }
     }
 
+    update.serviceCategoryIds?.let {
+      val serviceCategories = serviceCategoryRepository.findByIdIn(it)
+      if (!referral.intervention.dynamicFrameworkContract.contractType.serviceCategories.containsAll(serviceCategories)) {
+        errors.add(FieldError(field = "serviceCategories", error = Code.INVALID_SERVICE_CATEGORY_FOR_CONTRACT))
```
tomsmyers, line 213: i think the field name is incorrect in this error. it should be `serviceCategoryIds`?

tomsmyers (approved):
nice work

PR #250, opened 2021-05-21
IC-951: Turn on web application firewall on live envs

sldblog:

More details: https://user-guide.cloud-platform.service.justice.gov.uk/documentation/other-topics/modsecurity.html

Our agreements with the security team at the beginning was that we can go live without IP allow listing if we turn on Web Application Firewall. See the link above on why `mod_security` setting on the ingress achieves this.

### DIFF
--- a/helm_deploy/hmpps-interventions-service/templates/ingress.yaml
+++ b/helm_deploy/hmpps-interventions-service/templates/ingress.yaml
@@ -7,8 +7,16 @@ metadata:
   labels:
     {{- include "app.labels" . | nindent 4 }}
   annotations:
-    kubernetes.io/ingress.class: "nginx"
     nginx.ingress.kubernetes.io/custom-http-errors: "418"
+    {{- if .Values.env_details.contains_live_data }}
+    kubernetes.io/ingress.class: "modsec01"
+    nginx.ingress.kubernetes.io/enable-modsecurity: "true"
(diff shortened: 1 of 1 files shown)


tomsmyers (approved):
approving, but i don't really fully understand this change. would be good to get 👀 from someone more in tune with ops.

helm_deploy/hmpps-interventions-service/templates/ingress.yaml
```diff
@@ -7,8 +7,16 @@ metadata:
   labels:
     {{- include "app.labels" . | nindent 4 }}
   annotations:
-    kubernetes.io/ingress.class: "nginx"
     nginx.ingress.kubernetes.io/custom-http-errors: "418"
+    {{- if .Values.env_details.contains_live_data }}
```
mattops, line 11: Perhaps this should be enabled on all environments so you can deal with any potential false positives sooner.

mattops (commented):
LGTM

helm_deploy/hmpps-interventions-service/templates/ingress.yaml
sldblog: 👍 Thanks! I think I will explore that next week -- it would be a bit weird to accidentally block development and user research at this time

PR #253, opened 2021-05-24
IC-1746: Transformed Desired Outcomes to accept selected service cate…

andrewburrowsmoj:
…gory

Provide ability to set desired outcome on a draft referral for a given service category

To provide SP users the ability to provide desired outcomes for cohort referrals.


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/controller/ReferralController.kt
```diff
@@ -156,6 +157,13 @@ class ReferralController(
     return DraftReferralDTO.from(updatedReferral)
   }
 
+  @PatchMapping("/draft-referral/{id}/desired-outcomes")
+  fun setDraftReferralDesiredOutcomes(authentication: JwtAuthenticationToken, @PathVariable id: UUID, @RequestBody request: SetDesiredOutcomesRequestDTO): DraftReferralDTO{
+    val referral = getDraftReferralForAuthenticatedUser(authentication, id)
+    val updatedReferral = referralService.updateDraftReferralDesiredOutcomes(referral, request.serviceCategoryId, request.desiredOutcomesIds)
+    return DraftReferralDTO.from(updatedReferral)
+  }
+
```
andrewburrowsmoj, line 166: I'm assuming here that we can allow the caller to remove all desired outcomes for a service category. 

I assume this will be required if the user deselects the service category from the make a referral - we will also need to remove any desired outcomes as well.

However the behaviour before was to prevent it being unset... there could be a check in the ReferralService to ensure that at least one desired outcome for all service categories is left. Do you think I should add this check?

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/dto/DraftReferralDTO.kt
```diff
@@ -27,7 +27,7 @@ data class DraftReferralDTO(
   val additionalRiskInformation: String? = null,
   val usingRarDays: Boolean? = null,
   val maximumRarDays: Int? = null,
-  val desiredOutcomesIds: List<UUID>? = null,
+  val desiredOutcomes: Map<UUID, List<UUID>>? = null,
```
andrewburrowsmoj, line 30: Realise that I should be using a new DTO object rather than Map based on the code in the UI.
Going to make that change quickly just now...

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/controller/ReferralController.kt
tomsmyers: we need to explicitly stop unsetting of desiredoutcomes. previously this check handled this in `validateDraftReferralUpdate`

```
      if (it.isEmpty()) {
        errors.add(FieldError(field = "desiredOutcomesIds", error = Code.CANNOT_BE_EMPTY))
      }
      ```

File: src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/dto/DraftReferralDTO.kt [hunk 3]
```diff
@@ -61,3 +61,8 @@ data class DraftReferralDTO(
     }
   }
 }
+
+data class SelectedDesiredOutcomesDTO(
```
tomsmyers, line 65: please can this go in it's own file

tomsmyers: it would be fine here, except that it's also used on its own outside of the context of DraftReferralDTO, which ReferralComplexityLevel is not

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/controller/ReferralController.kt
andrewburrowsmoj: (talking to myself here)

.. we should make sure that we remove desiredOutcomes and complexityLevel for a given serviceCategoryId when it is removed via the updateDraftReferral route.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/ReferralService.kt
```diff
@@ -297,6 +286,31 @@ class ReferralService(
...
+
+    if (!validComplexityLevelIds.containsAll(desiredOutcomeIds)) {
+      throw ServerWebInputException("desired outcomes cannot be updated: at least one desired outcome is not valid for this service category")
+    }
+
+    if (referral.selectedDesiredOutcomes == null) {
+      referral.selectedDesiredOutcomes = mutableListOf()
+    }
+
+    referral.selectedDesiredOutcomes!!.removeIf { desiredOutcome -> desiredOutcome.serviceCategoryId === serviceCategoryId }
```
tomsmyers, line 309: nit: you shouldn't need the not null assertion here. the compiler will pick up that it cannot be null due to the above check ( i think)

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/ReferralService.kt
```diff
@@ -297,6 +286,31 @@ class ReferralService(
...
+  fun updateDraftReferralDesiredOutcomes(referral: Referral, serviceCategoryId: UUID, desiredOutcomeIds: List<UUID>): Referral {
+    if (referral.selectedServiceCategories.isNullOrEmpty()) {
+      throw ServerWebInputException("desired outcomes cannot be updated: no service categories selected for this referral")
+    }
+
+    if (!referral.selectedServiceCategories!!.map { it.id }.contains(serviceCategoryId)) {
+      throw ServerWebInputException("desired outcomes cannot be updated: specified service category not selected for this referral")
+    }
+
+    val validComplexityLevelIds = serviceCategoryRepository.findByIdOrNull(serviceCategoryId)?.desiredOutcomes?.map { it.id }
```
tomsmyers, line 298: variable name copy paste error

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/dto/DraftReferralDTO.kt
andrewburrowsmoj: Ok, I did originally, but noticed that InterventionDTO had a DTO in the same file.
I'll do that change.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/ReferralService.kt
tomsmyers, line 309: i'm not sure using the referential equality operator is correct here (`===`). (too much javascript ?? 😛 )

andrewburrowsmoj: Oh... my IDE makes it red if I don't have ? or !!

andrewburrowsmoj: hmm somehow it appears to work on the unit test, but I'll update to '=='

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/controller/ReferralController.kt
tomsmyers: not following the comment about removing them from updateDraftReferral.. aren't they already removed in this PR here https://github.com/ministryofjustice/hmpps-interventions-service/pull/253/files#diff-f7a6efe2090c4b294417026b860873e6c0fbc3f1b603412b45a294f1e44150c9L250 ?

i think we just need to check if the list of desiredoutcomes is empty in the new referral service method for updating the outcomes

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/dto/DraftReferralDTO.kt
andrewburrowsmoj: fixed

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/ReferralService.kt
andrewburrowsmoj: fixed

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/ReferralService.kt
andrewburrowsmoj: fixed

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/controller/ReferralController.kt
andrewburrowsmoj: Added back in the check for empty list. 

Setting to empty list is still possible via UpdateDraftReferral when unselecting a service Category. This will be a new PR with a new ticket (still to be created).

andrewburrowsmoj:
Rebased from main and resolve conflicts

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/ReferralService.kt
```diff
@@ -312,6 +301,40 @@ class ReferralService(
     return referralRepository.save(referral)
   }
 
+  fun updateDraftReferralDesiredOutcomes(referral: Referral, serviceCategoryId: UUID, desiredOutcomeIds: List<UUID>): Referral {
+    val errors = mutableListOf<FieldError>()
+
+    if (desiredOutcomeIds.isEmpty()) {
+      errors.add(FieldError(field = "desiredOutcomesIds", error = Code.CANNOT_BE_EMPTY))
```
tomsmyers, line 308: small thing: these `FieldError` objects are intended to be used when there are multiple fields being updating in a request. in this case, because we are only ever updating desired outcomes, i think just a plain 400 error is preferable here. it is more consistent with the responses generated for other error conditions below.

src/main/resources/db/migration/V1_53__referral_selected_desired_outcomes.sql
```diff
@@ -0,0 +1,14 @@
+begin;
```
tomsmyers, line 1: just noticed this lonely `begin`, what does it do?

tomsmyers: i guess it goes with the `commit` at the end?

tomsmyers (approved):
approving now but would be good to get the error code thing changed before merge. thanks!

src/main/resources/db/migration/V1_53__referral_selected_desired_outcomes.sql
andrewburrowsmoj: It wraps up the whole set of statements into a single transaction so rolls back if for instance the update fails.

It was more for when I was running directly on the database for testing - here it's not really needed but shouldn't hurt to have it

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/ReferralService.kt
andrewburrowsmoj: Ok. I assumed because it's something that is directly related to the user's parameters (rather than indirectly) that we needed to use this type of exception.

I have changed the exception to ServerWebInputException

src/main/resources/db/migration/V1_53__referral_selected_desired_outcomes.sql
tomsmyers: nice

PR #255, opened 2021-05-24
IC-1772: Delete selected desiredOutcomes and complexityLevel when selected ServiceCategory is unselected

andrewburrowsmoj:

Ensure that desired outcomes and complexity level are removed from referral for any that don't match to the selected service categories

Prevent unlinked data from being persisted to the database

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/ReferralService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/ReferralService.kt
@@ -270,10 +270,18 @@ class ReferralService(
       )
     }
 
-    update.serviceCategoryIds?.let {
-      referral.selectedServiceCategories = serviceCategoryRepository.findByIdIn(it)
+    // The selected complexity levels and desired outcomes need to be removed if the user has unselected service categories that are linked to them
+    update.serviceCategoryIds?.let { serviceCategoryIds ->
+      referral.complexityLevelIds = referral.complexityLevelIds?.filterKeys { serviceCategoryId -> serviceCategoryIds.contains(serviceCategoryId) }?.toMutableMap()
+      referral.selectedDesiredOutcomes = referral.selectedDesiredOutcomes?.filter { desiredOutcome -> serviceCategoryIds.contains(desiredOutcome.serviceCategoryId) }?.toMutableList()
     }
 
+    // Need to save and flush the entity before removing selected service categories due to constraint violations being thrown from selected complexity levels and desired outcomes
+    referralRepository.saveAndFlush(referral)
+
+    update.serviceCategoryIds?.let { serviceCategoryIds ->
+      referral.selectedServiceCategories = serviceCategoryRepository.findByIdIn(serviceCategoryIds)
+    }
     return referralRepository.save(referral)
   }
 
--- a/src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/ReferralServiceTest.kt
+++ b/src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/ReferralServiceTest.kt
@@ -22,6 +22,7 @@ import uk.gov.justice.digital.hmpps.hmppsinterventionsservice.config.ValidationE
 import uk.gov.justice.digital.hmpps.hmppsinterventionsservice.dto.DraftReferralDTO
 import uk.gov.justice.digital.hmpps.hmppsinterventionsservice.events.ReferralEventPublisher
 import uk.gov.justice.digital.hmpps.hmppsinterventionsservice.jpa.entity.AuthUser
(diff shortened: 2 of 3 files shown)


andrewburrowsmoj:
Please view the 2nd commit as I've had to branch off IC-1746 which is still awaiting approval

andrewburrowsmoj:
I'm writing one more test with real entity manager (will be pushed soon)

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/ReferralService.kt
```diff
@@ -281,10 +270,18 @@ class ReferralService(
...
+      referral.selectedDesiredOutcomes = referral.selectedDesiredOutcomes?.filter { desiredOutcome -> serviceCategoryIds.contains(desiredOutcome.serviceCategoryId) }?.toMutableList()
     }
 
+    // Need to save and flush the entity before removing selected service categories due to constraint violations being thrown
+    referralRepository.save(referral)
+    referralRepository.flush()
+
+    update.serviceCategoryIds?.let { serviceCategoryIds ->
+      referral.selectedServiceCategories = serviceCategoryRepository.findByIdIn(serviceCategoryIds)
+    }
```
andrewburrowsmoj, line 284: Need to save, flush and save to prevent the entitymanager attempting to delete the selected service categories first before the selected complexity level and desired outcomes. This would otherwise throw a constraint violation.

(This occurs even if you order sequence of statements in the correct order)

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/ReferralService.kt
```diff
@@ -281,10 +270,18 @@ class ReferralService(
...
     return referralRepository.save(referral)
   }
```
andrewburrowsmoj, line 286: Need to save, flush and save to prevent the entitymanager attempting to delete the selected service categories first before the selected complexity level and desired outcomes. This would otherwise throw a constraint violation.

(This occurs even if you order sequence of statements in the correct order)

andrewburrowsmoj:
rebased from main

sldblog:
At the risk of being annoying, could the pull request title also have a summary of the changes, please? It'd help me make sense of my notifications without having to look into JIRA or the contents of the PR

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/ReferralService.kt
sldblog: Would this be simpler if both our data model and the JSON payload would support this on a separate endpoints?

`PUT /draft-referral/{id}/service-category` &rarr; creates or updates categories, outcomes, complexities
`DELETE /draft-referral/{id}/service-category` &rarr; deletes

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/ReferralService.kt
```diff
@@ -270,10 +270,18 @@ class ReferralService(
       )
     }
 
-    update.serviceCategoryIds?.let {
-      referral.selectedServiceCategories = serviceCategoryRepository.findByIdIn(it)
+    update.serviceCategoryIds?.let { serviceCategoryIds ->
```
tomsmyers, line 273: happy with these changes. however i would like a comment in this block to describe what's going on here. i don't think it's immediately obvious when reading the code.

also a couple of really tiny things:

1. `referral.complexityLevelIds?.filter { complexityLevel -> serviceCategoryIds.contains(complexityLevel.key) }` could be simplified to `referral.complexityLevelIds?.filterKeys { serviceCategoryId -> serviceCategoryIds.contains(serviceCategoryId) }`

2. you can combine `save` and `flush` into `saveAndFlush(referral)`

tomsmyers (approved):

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/ReferralService.kt
andrewburrowsmoj: Some good ideas..
However I think the way that the DraftReferral has been designed is that complexity level and desired outcomes are not directly linked to the service category and are instead linked to the referral so it would be a bit less confusing if the update occurred on the DraftReferral resources directly. The user will never want to DELETE a selected ServiceCategory instead they would only ever want to replace it with a new selection.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/ReferralService.kt
andrewburrowsmoj: I've done as you've suggested

PR #256, opened 2021-05-24
Bubble communty-api exceptions up to Spring and to Sentry

sldblog:

Removes custom error handling on the community-api client handler side.

Also does two minor things:
- `makeAsyncPatchRequest` is unused, so removes it
- `makeAsyncPostRequest` &rarr; `makeFireAndForgetRequest` as we seem to differentiate between async/sync in that we don't actually care about its return value

The hypothesis is that this will allow us to have better context and stacktraces for errors (and perhaps even tracing in AppInsights)

If this experiment doesn't work, 948855d8bf915832842459b71f1f61852f63d9c0 can be reverted


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/component/CommunityAPIClient.kt
```diff
@@ -1,65 +1,26 @@
...
-    communityApiWebClient.patch().uri(uri)
-      .body(Mono.just(requestBody), requestBody::class.java)
-      .retrieve()
-      .bodyToMono(Unit::class.java)
-      .onErrorResume { e ->
-        handleResponse(e, requestBody)
-        Mono.empty()
-      }
-      .subscribe()
+      .block()
```
sldblog, line 16: go back to `subscribe()` to make it clearer that it's fire and forget

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/component/CommunityAPIClient.kt
```diff
@@ -1,65 +1,26 @@
...
-      else -> e.localizedMessage
-    }
-    logger.error(
-      "Call to community api failed",
-      e,
-      StructuredArguments.kv("req.body", requestBody),
-      StructuredArguments.kv("res.body", responseBodyAsString)
-    )
-    return responseBodyAsString.let { it } ?: "No response body message"
+      .block()!!
```
sldblog, line 24: Try verify the assumption that we can rely on Spring+Sentry to provide enough and good context for these kind of situations, run this locally with Sentry to have supporting data

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/component/CommunityAPIClient.kt
```diff
@@ -1,65 +1,26 @@
...
-
-  fun handleResponse(e: Throwable, requestBody: Any): String {
-    val responseBodyAsString = when (e) {
-      is BadRequest -> e.responseBodyAsString
-      else -> e.localizedMessage
-    }
-    logger.error(
-      "Call to community api failed",
-      e,
-      StructuredArguments.kv("req.body", requestBody),
```
sldblog, line 60: double check that Sentry can receive potential offender data

sldblog (commented):
Notes from collaborative review session

sldblog:
I'll split the rename + bubble up (and test it)

PR #257, opened 2021-05-25
Switch from cmd arg

green-digger:

Switches the reporting and analytical jobs to supply arguments rather than commands to the docker container. Plus, removes the env variables 'LOCAL_EXPORT_DESTINATION' as that is now set as part of the entry point wrapper.

For the analytical job, it switches to calling the python scripts directly.

To ensure the entry point wrapper can set env variables before running the supplied arguments.


sldblog (approved):
👍 Might conflict with #259

PR #258, opened 2021-05-25
Add observable IDs to the messages on group configuration issues

sldblog:

Essentially, add IDs to these `AccessError` reasons:

```diff
- configErrors.add("service provider id does not exist in the interventions database")
+ configErrors.add("service provider id '$providerGroupCode' does not exist in the interventions database")
```

```diff
- configErrors.add("user has no valid contract groups configured")
+ configErrors.add("contract '$removedContract' does not exist in the interventions database")
```

It also adds a **FIXME** to `ServiceProviderAccessScopeMapper.kt` about excluding contracts that belong to other providers 

So when someone has a config error, we don't have to

1. find out the user id
2. look up the groups of the user id from hmpps-auth
3. figure out what it was

but have this information at hand (it's not sensitive)

It will also be useful for displaying on a "what can I have access to" page

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/authorization/ServiceProviderAccessScopeMapper.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/authorization/ServiceProviderAccessScopeMapper.kt
@@ -48,6 +48,8 @@ class ServiceProviderAccessScopeMapper(
 
     val contracts = getContracts(contractGroups, configErrors)
 
+    // FIXME we also need to remove contracts which do not belong to the user's provider
+
     if (configErrors.isNotEmpty()) {
       throw AccessError(errorMessage, configErrors)
     }
@@ -84,8 +86,9 @@ class ServiceProviderAccessScopeMapper(
       emptyList()
     } else {
       val contracts = dynamicFrameworkContractRepository.findAllByContractReferenceIn(contractGroups)
-      if (contracts.isEmpty()) {
-        configErrors.add("user has no valid contract groups configured")
+      val removedContracts = contractGroups.subtract(contracts.map(DynamicFrameworkContract::contractReference))
+      for (removedContract in removedContracts) {
+        configErrors.add("contract '$removedContract' does not exist in the interventions database")
       }
       contracts
     }
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/authorization/ServiceProviderAccessScopeMapper.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/authorization/ServiceProviderAccessScopeMapper.kt
@@ -68,9 +68,10 @@ class ServiceProviderAccessScopeMapper(
         null
       }
       else -> {
-        val provider = serviceProviderRepository.findByIdOrNull(serviceProviderGroups[0])
+        val providerGroupCode = serviceProviderGroups[0]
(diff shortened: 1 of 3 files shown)


src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/integration/authorization/SingleReferralEndpoints.kt
```diff
@@ -230,7 +230,7 @@ class SingleReferralEndpoints : IntegrationTestBase() {
       """
       {"accessErrors": [
       "no service provider groups associated with user",
-      "no contract groups associated with user",
+      "no contract groups associated with user"
```
sldblog, line 233: This was complaining to me that it's not valid JSON to have a line-trailing `,`

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/authorization/ServiceProviderAccessScopeMapper.kt
```diff
@@ -83,8 +86,9 @@ class ServiceProviderAccessScopeMapper(
       emptyList()
     } else {
       val contracts = dynamicFrameworkContractRepository.findAllByContractReferenceIn(contractGroups)
-      if (contracts.isEmpty()) {
-        configErrors.add("user has no valid contract groups configured")
+      val removedContracts = contractGroups.subtract(contracts.map(DynamicFrameworkContract::contractReference))
+      for (removedContract in removedContracts) {
```
tomsmyers, line 90: tiny nit: using `forEach` is more idiomatic here

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/authorization/ServiceProviderAccessScopeMapper.kt
```diff
@@ -83,8 +86,9 @@ class ServiceProviderAccessScopeMapper(
       emptyList()
     } else {
       val contracts = dynamicFrameworkContractRepository.findAllByContractReferenceIn(contractGroups)
-      if (contracts.isEmpty()) {
-        configErrors.add("user has no valid contract groups configured")
+      val removedContracts = contractGroups.subtract(contracts.map(DynamicFrameworkContract::contractReference))
```
tomsmyers, line 89: i've not seen this type of argument to 'map' before - does it do the same as 

`val removedContracts = contractGroups.subtract(contracts.map { it.contractReference })`?

tomsmyers (approved):
all good - i wish this information could be presented in a structured way - but it's better than nothing

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/authorization/ServiceProviderAccessScopeMapper.kt
sldblog: yes, it's the method reference syntax: https://docs.oracle.com/javase/tutorial/java/javaOO/methodreferences.html

> 

>     Arrays.sort(rosterAsArray,
>         (a, b) -> Person.compareByAge(a, b)
>     );
>
> Because this lambda expression invokes an existing method, you can use a method reference instead of a lambda expression:
>
>     Arrays.sort(rosterAsArray, Person::compareByAge);

sldblog: In hindsight, it's probably not worth the extra characters...

PR #261, opened 2021-05-26
DAPI-177 Provide Referral Id as part of requests to community-api so ensure a unique NSI can be located

mahalrmoj:
There is an edge case when more that one Referral may be created in R&M that appear to cover the same Intervention, i.e. are for the same CRN and Sentence, cover the same Contract Type (e.g. Personal Wellbeing) and start on the same date. Prior to this change they would both be tied to the same NSI in Delius. After this change, each can result in its own NSI using the ReferralId (in the form of an URN) to distinguish them in community-api. 

To ensure that should multiple referrals be created which appear to be the same, they can be accounted for correctly by the calculations carried out in Delius.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/CommunityAPIService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/CommunityAPIService.kt
@@ -16,6 +16,7 @@ import uk.gov.justice.digital.hmpps.hmppsinterventionsservice.events.ReferralEve
 import uk.gov.justice.digital.hmpps.hmppsinterventionsservice.events.ReferralEventType
 import uk.gov.justice.digital.hmpps.hmppsinterventionsservice.jpa.entity.Referral
 import java.time.OffsetDateTime
+import java.util.UUID
 
 interface CommunityAPIService {
   fun getNotes(referral: Referral, url: String, description: String): String {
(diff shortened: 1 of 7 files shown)


sldblog:
To the intent, I'd also add that this is how we want the overall integration to work -- we own the activity, so it's not our job to _request_ a global ID, but the _collector_ service to track the context

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/CommunityAPIService.kt
```diff
@@ -228,6 +233,7 @@ data class ReferRequest(
   val contractType: String,
   val startedAt: OffsetDateTime,
   val sentenceId: Long,
+  val referralId: UUID,
```
sldblog, line 236: Does it make sense to remove everything _except_ referral IDs?

sldblog (approved):
👍

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/CommunityAPIService.kt
mahalrmoj: ReferralId is only used as a last resort if the existing criteria don't result in a unique match.

sldblog: But it can be used as known 1:1 mapping, right?

mahalrmoj: Principally yes but I think searching via a notes field in the database is not a preferred approach. Currently, CRN Sentence and NSI Type (Contract Type) is the criteria for querying the database in community-api.

mahalrmoj: I will merge this change for now and perhaps we can speak to @mxco86

PR #263, opened 2021-05-27
IC-1801: Amend CRS NDMIS report to support v1 schema

sldblog:

Amend CRS NDMIS report to support v1 schema:

- fills out `contract_type` as we have it now
- removes `caseworker_id`
- adds `intervention_end_reason`

To satisfy the report schema, which defines these fields:
```
referral_ref
referral_id
crs_contract_reference
crs_contract_type
crs_provider_id
referring_officer_id
relevant_sentence_id
service_user_crn
date_referral_received
date_saa_booked
date_saa_attended
date_first_action_plan_submitted
date_of_first_action_plan_approval
date_of_first_session
outcomes_to_be_achieved_count
outcomes_progress
count_of_sessions_expected
count_of_sessions_attended
date_intervention_ended
intervention_end_reason
date_end_of_service_report_submitted
```

### DIFF
--- a/helm_deploy/hmpps-interventions-service/reports/crs_performance_report.sql
+++ b/helm_deploy/hmpps-interventions-service/reports/crs_performance_report.sql
@@ -3,7 +3,7 @@ COPY (
     r.reference_number      AS referral_ref,
     r.id                    AS referral_id,
     c.contract_reference    AS crs_contract_reference,
-    'coming-later'          AS crs_contract_type,
+    ct.code                 AS crs_contract_type,
     prime.id                AS crs_provider_id,
     r.sent_by_id            AS referring_officer_id,
     r.relevant_sentence_id  AS relevant_sentence_id,
@@ -37,6 +37,7 @@ COPY (
     referral r
     JOIN intervention i ON (r.intervention_id = i.id)
     JOIN dynamic_framework_contract c ON (i.dynamic_framework_contract_id = c.id)
+    JOIN contract_type ct ON (c.contract_type_id = ct.id)
     JOIN service_provider prime ON (c.prime_provider_id = prime.id)
     LEFT JOIN action_plan ap ON (ap.referral_id = r.id) --❗️assumes a SINGLE action plan
     LEFT JOIN end_of_service_report eosr ON (eosr.referral_id = r.id)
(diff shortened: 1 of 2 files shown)


helm_deploy/hmpps-interventions-service/reports/crs_performance_report.sql
```diff
@@ -25,12 +31,16 @@ COPY (
...
-    (
-      select count(app.id)
-      from action_plan_appointment app
-      where app.action_plan_id = ap.id and attended in ('YES', 'LATE')
-    )                       AS count_of_sessions_attended,
+    s.attended              AS count_of_sessions_attended,
     r.concluded_at          AS date_intervention_ended,
+    (
+      CASE
+        WHEN r.concluded_at IS NOT NULL AND eosr.id IS NULL THEN 'cancelled'
```
tomsmyers, line 38: i'm not sure this is the correct check. 

a referral could have been completed, but an EOSR may not yet have been created. 

in code we do this check as follows:

```
    if (!hasActionPlan)
      return CANCELLED

    if (hasAttendedNoAppointments)
      return CANCELLED

    if (hasAttendedSomeAppointments && hasSubmittedEndOfServiceReport)
      return PREMATURELY_ENDED

    if (hasAttendedAllAppointments && hasSubmittedEndOfServiceReport)
      return COMPLETED
```

tomsmyers (commented):
only one major thing around computing end reason

helm_deploy/hmpps-interventions-service/reports/crs_performance_report.sql
sldblog: Thanks @tomsmyers!

Looking at the surrounding code, in https://github.com/ministryofjustice/hmpps-interventions-service/blob/efc79fd4934c0d783ef37bf5ed059676f7cb47a5/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/ReferralConcluder.kt#L31-L32

we set the `concluded_at` time if we "done" with the referral, right?

So I'm assuming if we check that `concluded_at` exists, that means the state around end of service report can no longer change. Is that correct?

sldblog:
♻️ rebased

sldblog:
♻️ Rebased again to pass the tests

helm_deploy/hmpps-interventions-service/reports/crs_performance_report.sql
tomsmyers: yes you're right - thanks!

tomsmyers (approved):

PR #264, opened 2021-05-28
Get pact tests closer to being green again

tomsmyers:

Adds required states, fixes hibernate errors, reduces number of failing contracts

before this PR there are 22 failing pacts. after there are 13 - which are all legit AFAIK.

get pact tests to reflect that actual state of the API consumer expectations

### DIFF
--- a/src/test/resources/application-test.yml
+++ b/src/test/resources/application-test.yml
@@ -10,3 +10,8 @@ pactbroker:
   scheme: https
   consumerversionselectors:
     tags: main
+
+spring:
+  jpa:
+    properties:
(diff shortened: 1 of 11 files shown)


src/test/resources/application-test.yml
```diff
@@ -10,3 +10,8 @@ pactbroker:
   scheme: https
   consumerversionselectors:
     tags: main
+
+spring:
+  jpa:
+    properties:
+      hibernate.enable_lazy_load_no_trans: true
```
mahalrmoj, line 17: Never knew this existed

mahalrmoj (approved):

PR #279, opened 2021-06-02
DAPI-1824 Call to community-api for referral send is synchronous and made services transactions

mahalrmoj:
This PR moves the "Referral Sent" notification to Delius from being an async process to being synchronous, such that, if the call is not successful then Interventions will rollback the transaction.

To ensure the rollback functions correctly, services are annotated as Transactional. 

The reason for this change is to make the link between Interventions and Delius more reliable and thus keep Delius in sync. There are plans to make the notification async and until this is done, this PR will remain in place.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/AppointmentsService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/AppointmentsService.kt
@@ -56,7 +56,6 @@ class AppointmentsService(
     }
   }
 
-  @Transactional
   fun updateAppointment(
     actionPlanId: UUID,
     sessionNumber: Int,
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/AppointmentsService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/AppointmentsService.kt
@@ -15,8 +15,10 @@ import java.time.OffsetDateTime
 import java.util.UUID
 import javax.persistence.EntityExistsException
 import javax.persistence.EntityNotFoundException
+import javax.transaction.Transactional
 
 @Service
+@Transactional
 class AppointmentsService(
   val actionPlanAppointmentRepository: ActionPlanAppointmentRepository,
   val actionPlanRepository: ActionPlanRepository,
@@ -54,6 +56,7 @@ class AppointmentsService(
     }
   }
 
+  @Transactional
   fun updateAppointment(
     actionPlanId: UUID,
     sessionNumber: Int,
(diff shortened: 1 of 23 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/AppointmentsService.kt
```diff
@@ -54,6 +56,7 @@ class AppointmentsService(
     }
   }
 
+  @Transactional
```
tomsmyers, line 59: is this required here as well as on the class?

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/ReferralService.kt
```diff
@@ -130,8 +133,15 @@ class ReferralService(
 
   fun sendDraftReferral(referral: Referral, user: AuthUser): Referral {
     referral.sentAt = OffsetDateTime.now()
-    referral.sentBy = authUserRepository.save(user)
     referral.referenceNumber = generateReferenceNumber(referral)
+    referral.sentBy = authUserRepository.save(user)
```
mahalrmoj, line 137: move sentBy before referenceNumber step

mahalrmoj: Done

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/CommunityAPIReferralService.kt
```diff
@@ -0,0 +1,61 @@
...
+  fun send(referral: Referral) {
+    val url = UriComponentsBuilder.fromHttpUrl(interventionsUIBaseURL)
+      .path(interventionsUISentReferralLocation)
+      .buildAndExpand(referral.id)
+      .toString()
+
+    postReferralStartRequest(referral, url)
+  }
+
+  private fun postReferralStartRequest(referral: Referral, url: String) {
```
tomsmyers, line 34: nit: method name slightly confused me with start vs send

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/CommunityAPIReferralService.kt
```diff
@@ -0,0 +1,61 @@
...
+data class ReferralSentRequest(
+  val contractType: String,
+  val startedAt: OffsetDateTime,
+  val sentenceId: Long,
+  val referralId: UUID,
+  val notes: String,
+)
+
+data class ReferralSentResponseDTO(
+  @NotNull val nsiId: Long
```
tomsmyers, line 60: nit: i don't think the annotation is required. not nullable fields in the response class will error if missing.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/ReferralService.kt
tomsmyers: what's the reason for this?

tomsmyers (approved):
nice. how does the error handling story fit into this work? will it all be contained within the community api service layer?

mahalrmoj:
> nice. how does the error handling story fit into this work? will it all be contained within the community api service layer?

That's my next challenge - will consult u on it

PR #286, opened 2021-06-03
IC-1625 remove 'using_rar_days' from referral and rename 'rar' to 'enforceable' elsewhere

tomsmyers:

IC-1625 remove 'using_rar_days' from referral and rename 'rar' to 'enforceable' elsewhere

make the naming consistent with the UI and the biz terms


src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/ReferralServiceTest.kt
```diff
@@ -297,22 +297,6 @@ class ReferralServiceTest @Autowired constructor(
     referralService.updateDraftReferral(sampleReferral, DraftReferralDTO(hasAdditionalResponsibilities = true, whenUnavailable = "wednesdays"))
   }
 
-  @Test
-  fun `when usingRarDays is true, maximumRarDays must be set`() {
```
sldblog, line 301: Could we add a test please about what happens if the `maximumEnforceableDays` is null when we try to send the referral?

tomsmyers: sure thing. having said that i'm not sure we explicitly have tests for any other required fields having been set... 

perhaps something to add in another changeset to validate all the required fields?

sldblog (approved):
👍 Let's specify preconditions to sending a referral in a separate PR

PR #287, opened 2021-06-03
Remove DPIA check as it is now approved

sldblog:

Remove DPIA check from data extractors as the Data Protection Impact Assessment (DPIA) is now approved

To allow data to reach where it has to be


tomsmyers (approved):
LGTM

PR #288, opened 2021-06-03
Make service caterogies ordered in draft referral

andrewburrowsmoj:

Makes service category ordered in draft referral DTO

REST resources must be **idempotent**. To achieve this we need to make the service categories always return in the same order for each REST call.

## Further Info
The selected service categories is a Set on the database object so there is no guaranteed order. 

I've ordered complexityLevel and desiredOutcomes based as well to keep them consistent with the service categories even though they should always return in the same order.

This change ultimately comes about because the order in which the UI gets the service categories is always different in each call, which causes weirdness on make a referral form.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/dto/DraftReferralDTO.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/dto/DraftReferralDTO.kt
@@ -39,7 +39,7 @@ data class DraftReferralDTO(
         id = referral.id,
         createdAt = referral.createdAt,
         completionDeadline = referral.completionDeadline,
-        complexityLevels = referral.complexityLevelIds?.map { ReferralComplexityLevel(it.key, it.value) },
+        complexityLevels = referral.complexityLevelIds?.map { ReferralComplexityLevel(it.key, it.value) }?.sortedBy { it.serviceCategoryId },
         furtherInformation = referral.furtherInformation,
         additionalNeedsInformation = referral.additionalNeedsInformation,
         accessibilityNeeds = referral.accessibilityNeeds,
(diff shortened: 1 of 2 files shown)


sldblog (commented):
Agree on the intent of making the order consistent

Does this also apply to things like service categories on the interventions?

Could it be done on the lowest level -- i.e. we always retrieve service categories by default name ordering, in any query?

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/dto/DraftReferralDTO.kt
```diff
@@ -51,13 +51,14 @@ data class DraftReferralDTO(
...
-        desiredOutcomes = referral.selectedDesiredOutcomes?.groupBy { it.serviceCategoryId }?.map { (serviceCategoryId, desiredoutcomes) -> SelectedDesiredOutcomesDTO(serviceCategoryId, desiredoutcomes.map { it.desiredOutcomeId }) },
+        desiredOutcomes = referral.selectedDesiredOutcomes?.groupBy { it.serviceCategoryId }?.toSortedMap()
+          ?.map { (serviceCategoryId, desiredoutcomes) -> SelectedDesiredOutcomesDTO(serviceCategoryId, desiredoutcomes.map { it.desiredOutcomeId }.sorted()) },
         serviceUser = ServiceUserDTO.from(referral.serviceUserCRN, referral.serviceUserData),
         serviceProvider = ServiceProviderDTO.from(contract.primeProvider),
         relevantSentenceId = referral.relevantSentenceId,
         // TODO: remove this once cohort referrals changes are complete
         serviceCategoryId = if (referral.selectedServiceCategories?.isNotEmpty() == true) referral.selectedServiceCategories!!.elementAt(0).id else null,
-        serviceCategoryIds = referral.selectedServiceCategories?.map { it.id },
+        serviceCategoryIds = referral.selectedServiceCategories?.map { it.id }?.sorted(),
```
sldblog, line 61: Can we sort these by name, please?

andrewburrowsmoj: I've resorted based on name

tomsmyers (approved):
i think this is a fine fix for the time being - lots of discussion on data types has happened on slack - it would be good to capture these somewhere for the future.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/dto/DraftReferralDTO.kt
```diff
@@ -39,7 +39,7 @@ data class DraftReferralDTO(
         id = referral.id,
         createdAt = referral.createdAt,
         completionDeadline = referral.completionDeadline,
-        complexityLevels = referral.complexityLevelIds?.map { ReferralComplexityLevel(it.key, it.value) },
+        complexityLevels = referral.complexityLevelIds?.map { ReferralComplexityLevel(it.key, it.value) }?.sortedBy { it.serviceCategoryId },
```
sldblog, line 42: This ordering makes me nervous -- does it matter?

sldblog (approved):
It looks like _all_ of this ordering could've been omitted if the service categories and related settings were a distinct data structure or endpoint

```json
{
  "selectedServiceCategories": {
    "service-category-id": {
      "complexityLevelId": "...",
      "desiredOutcomeIds": "..."
    }
  }
}
```

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/dto/DraftReferralDTO.kt
andrewburrowsmoj: I will reorder based on selectedServiceCategory.name

sldblog:
Just want to flag that at this point it looks like this "things that belong together not represented together" approach is causing more and more problems.

From a relative outsider perspective, a _lot_ of effort has been put into making orchestration for these fields. Is it worth the extra effort?

andrewburrowsmoj:
I also prefer the data structure you propose - but that will require database changes which is quite a bit of work (given priorities)

sldblog:
I don't understand that point, sorry -- the JSON representation for the endpoints is independent from the database representation

andrewburrowsmoj:
Got you... yes, that will also be quite a bit of work as well!

PR #289, opened 2021-06-03
IC-1827 send risk information to ARN on referral send 

tomsmyers:

- Adds new client config for the assess risks and needs service API (ARN)
- Sends the risk information from the draft referral form to ARN on referral send
- Deletes the risk information from the referral once sent to ARN 
- Stores a risk id on the referral from ARN 
- Sends the risk id in the sent referral DTO

allow PP's to enter risk information and have it stored in ARN


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/ReferralService.kt
```diff
@@ -141,16 +142,33 @@ class ReferralService(
 
     /*
      * This is a temporary solution until a robust asynchronous link is created between Interventions
-     * and Delius. Once the asynchronous link is implemented, this feature can be turned off, and instead
-     * Delius will be notified via the normal asynchronous route
+     * and Delius/ARN. Once the asynchronous link is implemented, this feature can be turned off, and instead
+     * Delius/ARN will be notified via the normal asynchronous route
      */
+    submitAdditionalRiskInformation(referral, user)
```
tomsmyers, line 148: there is a big issue here is the subsequent call to community API fails... needs some thought.

helm_deploy/values-prod.yaml
```diff
@@ -19,4 +19,5 @@ env:
   HMPPSAUTH_BASEURL: "https://sign-in.hmpps.service.justice.gov.uk/auth"
   INTERVENTIONSUI_BASEURL: "https://refer-monitor-intervention.service.justice.gov.uk"
   COMMUNITYAPI_BASEURL: "https://community-api-secure.probation.service.justice.gov.uk"
+  ASSESSRISKSANDNEEDS_BASEURL: "https://assess-risks-and-needs-prod.hmpps.service.justice.gov.uk"
```
SWhitlam, line 22: we may change this to https://assess-risks-and-needs.hmpps.service.justice.gov.uk but I'll confirm ASAP!

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/ReferralService.kt
SWhitlam: Consider switching the calls around. Our API doesn't have a legacy service to write to so it's more likely that the Community API call will fail (famous last words I know!)

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/ReferralService.kt
```diff
@@ -141,16 +142,33 @@ class ReferralService(
...
     return sentReferral
   }
 
+  private fun submitAdditionalRiskInformation(referral: Referral, user: AuthUser) {
+    val riskId = referral.additionalNeedsInformation?.let { riskInformation ->
+      assessRisksAndNeedsService.createSupplementaryRisk(
+        referral.id,
+        referral.serviceUserCRN,
+        user.id,
+        OffsetDateTime.now(), // fixme: this should be the timestamp which we store when additionalRiskInformation is set
```
SWhitlam, line 162: You don't need to send us a timestamp here, if you don't we default it to LocalDateTime.now()

tomsmyers: we have a concept of this information potentially being somewhat outdated between the user inputing the risk data and submitting the form. sending the timestamp they actually entered the risk data seems like a good idea here. (as per the comment, that's what will eventually happen)

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/ReferralService.kt
tomsmyers: done - this should be sufficient for right now.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/ReferralService.kt
```diff
@@ -141,16 +142,34 @@ class ReferralService(
 
     /*
      * This is a temporary solution until a robust asynchronous link is created between Interventions
-     * and Delius. Once the asynchronous link is implemented, this feature can be turned off, and instead
-     * Delius will be notified via the normal asynchronous route
+     * and Delius/ARN. Once the asynchronous link is implemented, this feature can be turned off, and instead
+     * Delius/ARN will be notified via the normal asynchronous route
      */
     communityAPIReferralService.send(referral)
+    submitAdditionalRiskInformation(referral, user)
```
sldblog, line 149: Sorry, asking the transactional question 😅 

What should our behaviour be when the latter (risk info) does not succeed and we cannot persist the sent referral?

tomsmyers: yeh i have thought about this, and right now we are pretty stuck if that happens. 

answer about community api idempotency is 

"Then we will end up getting two NSIs and then when you progress the referral it will complain as it finds multiple"

sldblog: Can we perhaps defer it after persistence? Make it a two-stage process?

1. community-api
2. save
3. event
4. risk
5. save

?

sldblog: Based on: observing state will tell us if the persistence was successful or not, so we can always retry this on sent referrals that still have the risk info

src/main/resources/db/migration/R__data_dictionary.sql
```diff
@@ -55,6 +55,7 @@ COMMENT ON COLUMN referral.interpreter_language IS 'what language the interprete
 COMMENT ON COLUMN referral.has_additional_responsibilities IS 'if the service user has attitional responsibilities in their life';
 COMMENT ON COLUMN referral.when_unavailable IS 'when the service user is unavailable due to existing commitments';
 COMMENT ON COLUMN referral.additional_risk_information IS 'further risk information about the service user';
+COMMENT ON COLUMN referral.supplementary_risk_id IS 'ID of the supplementary risk information stored in ''assess risks and needs'' service for this referral';
```
sldblog, line 58: ❤️

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/ReferralService.kt
sldblog: Ok I muddled this a bit. I meant making the transaction successful just after points 1-3.

But it's ok -- our recovery runbook for this is: delete the NSI and let the user try again

sldblog: Added it under https://dsdmoj.atlassian.net/wiki/spaces/IC/pages/3386179786/Recovery+scenarios#ARN-call-fails-and-the-referral-cannot-be-sent

sldblog (approved):

PR #293, opened 2021-06-04
Disable all emails, temporarily

sldblog:

Disable all emails on production

1. We need to change the referral notification email from "one address per provider" to "one address per intervention"
2. We do not yet have all email addresses on file

So in interest of keeping things rolling, we're turning of all emails until we have the above and/or start onboarding providers

### DIFF
--- a/helm_deploy/values-prod.yaml
+++ b/helm_deploy/values-prod.yaml
@@ -20,3 +20,4 @@ env:
   INTERVENTIONSUI_BASEURL: "https://refer-monitor-intervention.service.justice.gov.uk"
   COMMUNITYAPI_BASEURL: "https://community-api-secure.probation.service.justice.gov.uk"
   SENTRY_ENVIRONMENT: "prod"
+  NOTIFY_ENABLED: false


helm_deploy/values-prod.yaml
```diff
@@ -20,3 +20,4 @@ env:
   INTERVENTIONSUI_BASEURL: "https://refer-monitor-intervention.service.justice.gov.uk"
   COMMUNITYAPI_BASEURL: "https://community-api-secure.probation.service.justice.gov.uk"
   SENTRY_ENVIRONMENT: "prod"
+  NOTIFY_ENABLED: false
```
tomsmyers, line 23: how about preprod?

tomsmyers (approved):

helm_deploy/values-prod.yaml
sldblog: I assumed we wouldn't want to send emails to real providers on pre-prod, so that wouldn't be an issue. Is that correct?

PR #298, opened 2021-06-04
Use HMPPS auth 'authSource' for ARN 'createdByUserType'

tomsmyers:

Use HMPPS auth 'authSource' for ARN 'createdByUserType'

latest ARN API uses these strings for the user type field


sldblog (approved):
Nice 👍 

I wonder if hmpps-auth could create a single field as a unique person identifier...

PR #299, opened 2021-06-04
IC-1846 / IC-1847 store timestamp when updating draft referral risk information / update column names

tomsmyers:

IC-1846 store timestamp when updating draft referral risk information
IC-1847 update database column names with more descriptive field names (but don't change the API)

for auditing purposes we know exactly when the PP entered the risk information - even if it's quite different from when they sent off the referral

make the data dictionary clear in terms of what the risk fields represent.


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/ReferralService.kt
```diff
@@ -161,7 +161,7 @@ class ReferralService(
       referral.id,
       referral.serviceUserCRN,
       user,
-      OffsetDateTime.now(), // fixme: this should be the timestamp which we store when additionalRiskInformation is set
+      referral.additionalRiskInformationUpdatedAt,
       riskInformation,
     )
```
tomsmyers, line 167: should we clear the timestamp below too??

sldblog: would be good to retain it until we have a way to retrieve it from ARN

tomsmyers: 👍🏼

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/RisksAndNeedsService.kt
```diff
@@ -22,7 +22,7 @@ data class CreateSupplementaryRiskRequest(
   val crn: String,
   val createdByUser: String,
   val createdByUserType: String,
-  val createdDate: LocalDateTime,
+  val createdDate: LocalDateTime?,
```
sldblog, line 25: It sounds like this field shouldn't be optional -- intent-wise there must always be a time where the risk is 'signalled'

What do you think about setting `additionalRiskInformationUpdatedAt` to current time in `ReferralService.submitAdditionalRiskInformation` if it's missing? (Together with a warning?)

sldblog (commented):
👍 nice -- just a question about `riskCreatedAt` field being optional

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/RisksAndNeedsService.kt
tomsmyers: yes happy with that idea.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/jpa/entity/Referral.kt
```diff
@@ -54,7 +54,8 @@ class Referral(
 
   // draft referral fields
   @OneToOne(mappedBy = "referral", cascade = arrayOf(CascadeType.ALL)) @PrimaryKeyJoinColumn var serviceUserData: ServiceUserData? = null,
-  var additionalRiskInformation: String? = null,
+  @Column(name = "draft_supplementary_risk") var additionalRiskInformation: String? = null,
```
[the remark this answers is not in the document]
tomsmyers: i tried doing this. and i found it confusing having the difference there between the entity field and the DTO field - especially in the test code.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/ReferralService.kt
```diff
@@ -157,11 +157,17 @@ class ReferralService(
     val riskInformation = referral.additionalRiskInformation
       ?: throw ServerWebInputException("can't submit a referral without risk information")
 
+    val riskSubmittedAt = referral.additionalRiskInformationUpdatedAt
```
tomsmyers, line 160: @sldblog changes as suggested here

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/ReferralServiceUnitTest.kt
```diff
@@ -578,41 +589,69 @@ class ReferralServiceUnitTest {
...
 
-    @Test
-    fun `cant send draft referral without risk information`() {
-      val referral = referralFactory.createDraft()
-      val authUser = authUserFactory.create()
-
-      val e = assertThrows<ServerWebInputException> {
-        referralService.sendDraftReferral(referral, authUser)
-      }
+  @Test
```
sldblog, line 594: _nit_ Wondering if we should have a ktlint job 🤔

sldblog (approved):
🎉

PR #300, opened 2021-06-05
Final changes to get the contract tests green

tomsmyers:
please review each commit in isolation.

This PR does include some functional changes! 

- collection type referral fields (desired outcomes, complexity levels, service categories) are explicitly set to null if empty. the reason for this is to ensure the responses are consistent between freshly created referrals (these fields are `null`) and retrieved referrals (where they are empty lists - due to the way JPA loads them). since these things can never be emptied via our api, an empty list isn't a valid state for the fields to be in. so if the collection is empty in all three cases, we return a null value in the response. this also ties in with a failing contract on the UI which was expecting complexity levels to be null, but it was returning an empty list.

- sent referral DTOs no longer include additionalRiskInformation. this is to avoid API clients trying to use it. once the referral is sent we do not store or purport to store _any_ risk information in the service. this API changes makes that clear.

get the pacts green before prod usage commences.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/controller/ReferralController.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/controller/ReferralController.kt
@@ -1,5 +1,6 @@
 package uk.gov.justice.digital.hmpps.hmppsinterventionsservice.controller
 
+import com.fasterxml.jackson.annotation.JsonView
 import mu.KLogging
 import org.springframework.http.HttpStatus
 import org.springframework.http.ResponseEntity
@@ -23,6 +24,7 @@ import uk.gov.justice.digital.hmpps.hmppsinterventionsservice.dto.SelectedDesire
 import uk.gov.justice.digital.hmpps.hmppsinterventionsservice.dto.SentReferralDTO
 import uk.gov.justice.digital.hmpps.hmppsinterventionsservice.dto.ServiceCategoryFullDTO
 import uk.gov.justice.digital.hmpps.hmppsinterventionsservice.dto.SetComplexityLevelRequestDTO
+import uk.gov.justice.digital.hmpps.hmppsinterventionsservice.dto.Views
 import uk.gov.justice.digital.hmpps.hmppsinterventionsservice.jpa.entity.AuthUser
 import uk.gov.justice.digital.hmpps.hmppsinterventionsservice.jpa.entity.CancellationReason
 import uk.gov.justice.digital.hmpps.hmppsinterventionsservice.jpa.entity.Referral
@@ -40,6 +42,7 @@ class ReferralController(
 ) {
   companion object : KLogging()
 
+  @JsonView(Views.SentReferral::class)
   @PostMapping("/sent-referral/{id}/assign")
   fun assignSentReferral(
     @PathVariable id: UUID,
@@ -59,6 +62,7 @@ class ReferralController(
     )
   }
 
+  @JsonView(Views.SentReferral::class)
   @PostMapping("/draft-referral/{id}/send")
   fun sendDraftReferral(@PathVariable id: UUID, authentication: JwtAuthenticationToken): ResponseEntity<SentReferralDTO> {
     val user = userMapper.fromToken(authentication)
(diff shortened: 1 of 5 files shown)


sldblog:
There was a tiny change in https://github.com/ministryofjustice/hmpps-interventions-ui/pull/380 that introduced checking for empty that now can be rolled back I suppose

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/dto/DraftReferralDTOTest.kt
```diff
@@ -94,6 +94,29 @@ class DraftReferralDTOTest(@Autowired private val json: JacksonTester<DraftRefer
...
+  @Nested inner class RiskInformation {
+    // this field has some funky JsonView annotations going on,
+    // so double check it's working as expected
+
+    @Test
+    fun `test serialization of additionalRiskInformation in draft referrals`() {
+      val referral = referralFactory.createDraft()
+
+      val out = json.write(DraftReferralDTO.from(referral))
+      assertThat(out).hasJsonPath("$.additionalRiskInformation")
```
sldblog, line 106: Is it worth testing non-null values too (to assert it serialises as expected)?

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/controller/ReferralController.kt
```diff
@@ -59,6 +62,7 @@ class ReferralController(
     )
   }
 
+  @JsonView(Views.SentReferral::class)
   @PostMapping("/draft-referral/{id}/send")
   fun sendDraftReferral(@PathVariable id: UUID, authentication: JwtAuthenticationToken): ResponseEntity<SentReferralDTO> {
```
sldblog, line 67: 🤔 Would be nice if the `SentReferralDTO` could be annotated with `JsonView`, adding default behaviour to it on serialisation

sldblog (approved):
🎉 👍 LGTM -- I suppose we need to modelling review session once things settle down

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/controller/ReferralController.kt
tomsmyers: yes i tried this but with no joy...

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/dto/DraftReferralDTO.kt
```diff
@@ -49,14 +66,23 @@ data class DraftReferralDTO(
...
+            )
+          },
         serviceUser = ServiceUserDTO.from(referral.serviceUserCRN, referral.serviceUserData),
         serviceProvider = ServiceProviderDTO.from(contract.primeProvider),
         relevantSentenceId = referral.relevantSentenceId,
         // TODO: remove this once cohort referrals changes are complete
-        serviceCategoryId = if (referral.selectedServiceCategories?.isNotEmpty() == true) referral.selectedServiceCategories!!.elementAt(0).id else null,
-        serviceCategoryIds = referral.selectedServiceCategories?.map { it.id }?.sorted(),
+        serviceCategoryId = referral.selectedServiceCategories?.firstOrNull()?.id,
+        serviceCategoryIds = referral.selectedServiceCategories?.ifEmpty { null }
```
sldblog, line 83: should this be unit tested? or is it ok to rely on the contracts to do it?

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/controller/ReferralController.kt
sldblog: ah well. annotations 🙃

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/dto/DraftReferralDTO.kt
tomsmyers: i can add a quick unit test

tomsmyers: done

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/dto/DraftReferralDTOTest.kt
tomsmyers: sure thing - done

PR #302, opened 2021-06-06
IC-1593 allow http client timeouts to be set in config / comment out non functional LAO checks

tomsmyers:

- remove unneeded webclient config
- allow http client timeouts to be set in config
- comment out non functional LAO checks

incremental improvements to webclient code. see commit message / fixme comments for motivation on the other stuff.


sldblog (approved):
👍 I like the explicit timeouts + removal of `ExchangeStrategies.builder`

We could also remove said code vs commenting it out, but that's not a blocker

PR #305, opened 2021-06-07
IC-1593 renable LAO checks

tomsmyers:

- adds a new helper class for calling rest methods with webclients in a more consistent way
- changes the LAO check to use a new community API endpoint where we can specify the username in the url
- slightly changes the way we handle the response from the community api endpoint - instead of manually checking the exclusions and restrictions, we just check the status code, this is to avoid duplicating this logic which exists in community api:

```
    private ResponseEntity<AccessLimitation> accessLimitationResponseEntityOf(final OffenderDetail offender, final String username) {
        final var accessLimitation = userService.accessLimitationOf(username, offender);
        return new ResponseEntity<>(accessLimitation, (accessLimitation.isUserExcluded() || accessLimitation.isUserRestricted()) ? FORBIDDEN : OK);
    }
```
- changes the risks and needs service to use the new rest client helper class

implement LAO checks for PP referral access.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/CommunityAPIOffenderService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/CommunityAPIOffenderService.kt
@@ -1,0 +1,42 @@
+package uk.gov.justice.digital.hmpps.hmppsinterventionsservice.service
+
+import org.springframework.beans.factory.annotation.Value
+import org.springframework.http.HttpStatus
+import org.springframework.stereotype.Service
+import org.springframework.web.util.UriComponentsBuilder
+import reactor.core.publisher.Mono
+import uk.gov.justice.digital.hmpps.hmppsinterventionsservice.component.RestClient
+import uk.gov.justice.digital.hmpps.hmppsinterventionsservice.jpa.entity.AuthUser
+
+data class ServiceUserAccessResult(
+  val canAccess: Boolean,
+  val messages: List<String>,
+)
+
+private data class UserAccessResponse(
+  val exclusionMessage: String?,
+  val restrictionMessage: String?,
+)
+
+@Service
+class CommunityAPIOffenderService(
+  @Value("\${community-api.locations.offender-access}") private val offenderAccessLocation: String,
+  private val communityApiRestClient: RestClient,
+) {
+  fun checkIfAuthenticatedDeliusUserHasAccessToServiceUser(user: AuthUser, crn: String): ServiceUserAccessResult {
+    val userAccessPath = UriComponentsBuilder.fromPath(offenderAccessLocation)
+      .buildAndExpand(crn, user.userName)
+      .toString()
+
+    val response = communityApiRestClient.get(userAccessPath)
+      .retrieve()
+      .onStatus({ it.equals(HttpStatus.FORBIDDEN) }, { Mono.empty() })
+      .toEntity(UserAccessResponse::class.java)
+      .block()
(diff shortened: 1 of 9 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/CommunityAPIOffenderService.kt
```diff
@@ -0,0 +1,42 @@
...
+  private val communityApiRestClient: RestClient,
+) {
+  fun checkIfAuthenticatedDeliusUserHasAccessToServiceUser(user: AuthUser, crn: String): ServiceUserAccessResult {
+    val userAccessPath = UriComponentsBuilder.fromPath(offenderAccessLocation)
+      .buildAndExpand(crn, user.userName)
+      .toString()
+
+    val response = communityApiRestClient.get(userAccessPath)
+      .retrieve()
+      .onStatus({ it.equals(HttpStatus.FORBIDDEN) }, { Mono.empty() })
```
tomsmyers, line 33: returning empty mono here effectively ignores this status instead of throwing webclientexception

sldblog: could we add unit tests for this class, please? (to explicitly document the expected `ServiceUserAccessResult` data in cases)

tomsmyers: yes for sure.

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/CommunityAPIOffenderServiceTest.kt
```diff
@@ -0,0 +1,57 @@
...
+    val client = createMockedRestClient(HttpStatus.FORBIDDEN, "{\"exclusionMessage\": \"family exclusion\", \"restrictionMessage\": null}")
+    val offenderService = CommunityAPIOffenderService("/", client)
+    val result = offenderService.checkIfAuthenticatedDeliusUserHasAccessToServiceUser(authUserFactory.create(), "X123456")
+    assertThat(result.canAccess).isFalse
+    assertThat(result.messages).isEqualTo(listOf("family exclusion"))
+  }
+
+  @Test
+  fun `checkIfAuthenticatedDeliusUserHasAccessToServiceUser unhandled error`() {
+    val client = createMockedRestClient(HttpStatus.INTERNAL_SERVER_ERROR, "{\"exclusionMessage\": null, \"restrictionMessage\": null}")
```
sldblog, line 51: Interesting, do we get the same payload with errors?

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/CommunityAPIOffenderService.kt
```diff
@@ -0,0 +1,42 @@
...
+      .toEntity(UserAccessResponse::class.java)
+      .block()
+
+    return ServiceUserAccessResult(
+      !response.statusCode.equals(HttpStatus.FORBIDDEN),
```
sldblog, line 38: If this endpoint ever redirects or switches to another 4xx code, we might have a problem. Is there a test that says the response is the 2xx range ~~ successful?

sldblog (approved):
👍 no blockers (still in draft though)

PR #311, opened 2021-06-08
IC-1885: Added new entity for appointment delivery

andrewburrowsmoj:

Creates an entity that will allow to tie a meeting method and location details for an appointment

Provide ability to add location details

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/jpa/entity/AppointmentDelivery.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/jpa/entity/AppointmentDelivery.kt
@@ -1,0 +1,34 @@
+package uk.gov.justice.digital.hmpps.hmppsinterventionsservice.jpa.entity
+
+import com.vladmihalcea.hibernate.type.basic.PostgreSQLEnumType
+import org.hibernate.annotations.Type
+import org.hibernate.annotations.TypeDef
+import java.util.UUID
+import javax.persistence.Entity
+import javax.persistence.EnumType
+import javax.persistence.Enumerated
+import javax.persistence.Id
+import javax.persistence.OneToOne
+import javax.persistence.PrimaryKeyJoinColumn
+import javax.validation.constraints.NotNull
+
+@Entity
+@TypeDef(name = "appointment_delivery_type", typeClass = PostgreSQLEnumType::class)
+data class AppointmentDelivery(
+  @Id
+  var appointmentId: UUID,
+  @Type(type = "appointment_delivery_type")
+  @Enumerated(EnumType.STRING)
+  @NotNull var appointmentDeliveryType: AppointmentDeliveryType,
+  var npsOfficeCode: String? = null,
+  @OneToOne
+  @PrimaryKeyJoinColumn
+  var appointmentDeliveryAddress: AppointmentDeliveryAddress? = null,
+)
+
+enum class AppointmentDeliveryType {
+  PHONE_CALL,
+  VIDEO_CALL,
+  IN_PERSON_MEETING_PROBATION_OFFICE,
+  IN_PERSON_MEETING_OTHER;
+}
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/jpa/entity/Appointment.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/jpa/entity/Appointment.kt
@@ -13,6 +13,8 @@ import javax.persistence.EnumType
 import javax.persistence.Enumerated
 import javax.persistence.Id
 import javax.persistence.ManyToOne
+import javax.persistence.OneToOne
+import javax.persistence.PrimaryKeyJoinColumn
 import javax.validation.constraints.NotNull
 
 @Entity
@@ -35,6 +37,10 @@ data class Appointment(
   @NotNull val createdAt: OffsetDateTime,
   var appointmentTime: OffsetDateTime,
   var durationInMinutes: Int,
+
+  @OneToOne
+  @PrimaryKeyJoinColumn
+  var appointmentDelivery: AppointmentDelivery? = null,
 
   @Id val id: UUID,
 )
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/jpa/entity/AppointmentDeliveryAddress.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/jpa/entity/AppointmentDeliveryAddress.kt
@@ -1,0 +1,18 @@
+package uk.gov.justice.digital.hmpps.hmppsinterventionsservice.jpa.entity
+
+import java.util.UUID
+import javax.persistence.Entity
+import javax.persistence.Id
+import javax.persistence.Table
+
+@Entity
+@Table(name = "appointment_delivery_address")
+data class AppointmentDeliveryAddress(
+  @Id
+  var appointmentDeliveryId: UUID,
+  var firstAddressLine: String,
+  var secondAddressLine: String? = null,
+  var townCity: String,
+  var county: String,
+  var postCode: String,
+)
--- a/src/main/resources/db/migration/R__data_dictionary.sql
+++ b/src/main/resources/db/migration/R__data_dictionary.sql
@@ -97,3 +97,16 @@ COMMENT ON COLUMN action_plan_activity.id IS 'service user''s action plan activi
 COMMENT ON COLUMN action_plan_activity.action_plan_id IS 'service user''s action plan unique identifier';
 COMMENT ON COLUMN action_plan_activity.description IS 'description of the activity';
 COMMENT ON COLUMN action_plan_activity.created_at IS 'when the service user''s action plan was created';
+
+COMMENT ON TABLE appointment_delivery IS 'appointment delivery details';
+COMMENT ON COLUMN appointment_delivery.appointment_id IS 'appointment''s unique identifier';
+COMMENT ON COLUMN appointment_delivery.appointment_delivery_type IS 'type to denote how the appointment is delivered';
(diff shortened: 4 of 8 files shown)


andrewburrowsmoj:
tests are failing because there was a conflict with sql file versioning (due to rebase) - after correcting circleci is not triggering

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/jpa/entity/ActionPlanAppointmentMeetingMethod.kt
```diff
@@ -0,0 +1,36 @@
...
+
+@Entity
+@Table(name = "action_plan_appointment_meeting_method")
+@TypeDef(name = "meeting_method_code", typeClass = PostgreSQLEnumType::class)
+data class ActionPlanAppointmentMeetingMethod(
+  @Id val id: UUID,
+  @Type(type = "meeting_method_code")
+  @Enumerated(EnumType.STRING)
+  @NotNull var meetingMethodCode: MeetingMethodCode,
+  var videoCallAddress: String? = null,
```
tomsmyers, line 25: nit: is this an address or url?

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/jpa/entity/ActionPlanAppointmentMeetingMethod.kt
```diff
@@ -0,0 +1,36 @@
...
+  var npsOfficeCode: String? = null,
+  @ManyToOne @Fetch(FetchMode.JOIN)
+  var otherMeetingLocation: MeetingLocation? = null,
```
tomsmyers, line 28: addresses are notoriously difficult to store correctly when split into constituent parts (house name / number, street address, city, county, postal code, country etc etc).

my vote to keep this is simple as possible would be a single text field.

happy to be overruled on this though. 

cc @sldblog

tomsmyers: furthermore, if we used a plain text field for this, we could trim down this entity to simply be 

`{ id: UUID, type: MeetingMethodCode, details: String }`

no optional fields, and the details corresponds to the meeting method - i.e. it's a url for video calls, a phone number for phone calls, an address for in person meetings etc 

what do you think?

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/jpa/entity/ActionPlanAppointmentMeetingMethod.kt
```diff
@@ -0,0 +1,36 @@
...
+import javax.persistence.Entity
+import javax.persistence.EnumType
+import javax.persistence.Enumerated
+import javax.persistence.Id
+import javax.persistence.ManyToOne
+import javax.persistence.Table
+import javax.validation.constraints.NotNull
+
+@Entity
+@Table(name = "action_plan_appointment_meeting_method")
```
tomsmyers, line 18: please can we make this table a more generic thing? e.g. `appointment_delivery`

tomsmyers (commented):
thanks for getting the ball rolling on this andrew!

src/main/resources/db/migration/V1_56__action_plan_meeting_method.sql
```diff
@@ -0,0 +1,24 @@
+CREATE TYPE meeting_method_code AS ENUM ('PHONE_CALL','VIDEO_CALL', 'IN_PERSON_MEETING_OFFICE', 'IN_PERSON_MEETING_OTHER');
+
+create table meeting_location(
+    id uuid not null,
+--     TODO: Waiting for ming to confirm if we should use address field here
+    address text not null,
+    city VARCHAR(35) not null,
+    post_code VARCHAR(8) not null,
```
sldblog, line 8: there's a place which is 58 characters in long form, not counting typos: https://en.wikipedia.org/wiki/Llanfairpwllgwyngyll ;)

we can technically fit postcode into 8 characters, after sanitisation and stripping spaces even 7

however, we can not sanitise address/city as efficiently and we _could_ allow for user error, weird formatting, copy/paste errors

however, since `text` and `varchar` in PostgreSQL have the same overhead under 126 bytes:
> The storage requirement for a short string (up to 126 bytes) is 1 byte plus the actual string

... what do you think about using `text` everywhere?

src/main/resources/db/migration/V1_56__action_plan_meeting_method.sql
```diff
@@ -0,0 +1,24 @@
+CREATE TYPE meeting_method_code AS ENUM ('PHONE_CALL','VIDEO_CALL', 'IN_PERSON_MEETING_OFFICE', 'IN_PERSON_MEETING_OTHER');
```
sldblog, line 1:
```suggestion
CREATE TYPE meeting_method_code AS ENUM ('PHONE_CALL','VIDEO_CALL', 'IN_PERSON_MEETING_PROBATION_OFFICE', 'IN_PERSON_MEETING_OTHER');
```

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/jpa/entity/ActionPlanAppointmentMeetingMethod.kt
```diff
@@ -0,0 +1,36 @@
...
+  var videoCallAddress: String? = null,
+  var npsOfficeCode: String? = null,
+  @ManyToOne @Fetch(FetchMode.JOIN)
+  var otherMeetingLocation: MeetingLocation? = null,
+)
+
+enum class MeetingMethodCode {
+  PHONE_CALL,
+  VIDEO_CALL,
+  IN_PERSON_MEETING_OFFICE,
```
sldblog, line 34:
```suggestion
  IN_PERSON_MEETING_PROBATION_OFFICE,
```

src/main/resources/db/migration/V1_56__action_plan_meeting_method.sql
andrewburrowsmoj: To confess I copied the sql fields that were used in community-api as I thought it would be a good reference

tomsmyers: 🙀

src/main/resources/db/migration/V1_56__action_plan_meeting_method.sql
andrewburrowsmoj: changed to this

src/main/resources/db/migration/V1_56__action_plan_meeting_method.sql
andrewburrowsmoj: no longer relevant because we are using address field for everything

andrewburrowsmoj:
test

andrewburrowsmoj:
test

src/main/resources/db/migration/V1_57__appointment_delivery.sql
```diff
@@ -0,0 +1,21 @@
+CREATE TYPE appointment_delivery_type AS ENUM ('PHONE_CALL','VIDEO_CALL', 'IN_PERSON_MEETING_PROBATION_OFFICE', 'IN_PERSON_MEETING_OTHER');
+
+create table appointment_delivery(
+    appointment_id uuid not null,
+    appointment_delivery_type appointment_delivery_type not null,
+    nps_office_code VARCHAR(7) null,
```
sldblog, line 6: can I be annoying and ask to update https://github.com/ministryofjustice/hmpps-interventions-service/blob/main/src/main/resources/db/migration/R__data_dictionary.sql as well, please? Without URNs, it would be important with cross-system IDs to explain where the source of truth is

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/jpa/entity/AppointmentDelivery.kt
```diff
@@ -0,0 +1,35 @@
...
+import javax.validation.constraints.NotNull
+
+@Entity
+@TypeDef(name = "appointment_delivery_type", typeClass = PostgreSQLEnumType::class)
+data class AppointmentDelivery(
+  @Id
+  var appointmentId: UUID,
+  @Type(type = "appointment_delivery_type")
+  @Enumerated(EnumType.STRING)
+  @NotNull var appointmentDeliveryType: AppointmentDeliveryType,
```
sldblog, line 22: Can we introduce referential integrity validation, where we test the accompanying option against the detail field?

I.e. make it invalid to set it `VIDEO_CALL` without `videoCallUrl` etc etc.

src/main/resources/db/migration/V1_57__appointment_delivery.sql
andrewburrowsmoj: yes I will do that

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/jpa/entity/AppointmentDelivery.kt
andrewburrowsmoj: video call url is no longer required - we just set the type

I'd rather have that logic within the code since introducing these constraints makes it more complicated when it comes to updating/deleting (due to hibernate not understanding which order to perform operations) 

but if you feel strongly I will do it.

src/main/resources/db/migration/V1_57__appointment_delivery.sql
andrewburrowsmoj: I have now included that in the PR

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/jpa/entity/AppointmentDelivery.kt
sldblog: yeah, service code is fine -- I couldn't find it though, is that in the other PR?

PR #312, opened 2021-06-09
Publish domain events as business events to AppInsights

sldblog:

For all events that go through `SNSPublisher`, also send them to AppInsights as business events.

I considered pulling the IDs out of the `additionalInformation` field but didn't want to add too much extra complexity

image

Domain events are our pivotal events, so it would be good to be able to explicitly count/aggregate them without resolving them from other indicators


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/component/SNSPublisher.kt
```diff
@@ -1,25 +1,38 @@
...
   private val client: SnsClient,
   private val objectMapper: ObjectMapper,
+  private val telemetryClient: TelemetryClient,
   @Value("\${aws.sns.enabled}") private val enabled: Boolean,
   @Value("\${aws.sns.topic.arn}") private val arn: String,
 ) {
   companion object : KLogging()
 
-  fun publish(event: EventDTO) {
+  fun publish(referralId: UUID, actor: AuthUser?, event: EventDTO) {
```
tomsmyers, line 24: nit: non nullable actor would be nice

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/component/SNSPublisher.kt
```diff
@@ -1,25 +1,38 @@
...
     if (enabled) {
       buildRequestAndPublish(event)
+      telemetryClient.trackEvent(
+        "InterventionsDomainEvent",
```
tomsmyers, line 28: out of interest - using this type of naming scheme, what other event types could you envisage?

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/component/SNSPublisher.kt
```diff
@@ -1,25 +1,38 @@
...
   private val objectMapper: ObjectMapper,
+  private val telemetryClient: TelemetryClient,
   @Value("\${aws.sns.enabled}") private val enabled: Boolean,
   @Value("\${aws.sns.topic.arn}") private val arn: String,
 ) {
   companion object : KLogging()
 
-  fun publish(event: EventDTO) {
+  fun publish(referralId: UUID, actor: AuthUser?, event: EventDTO) {
     if (enabled) {
```
tomsmyers, line 25: nit: AI event should probably trigger even if SNS is disabled?

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/SNSService.kt
```diff
@@ -91,7 +91,8 @@ class SNSAppointmentService(
           event.appointment.attendanceSubmittedAt!!,
           mapOf("serviceUserCRN" to referral.serviceUserCRN, "referralId" to referral.id)
         )
-        snsPublisher.publish(snsEvent)
+        // FIXME we don't know who submits these -- needs explicit actor
```
tomsmyers, line 94: agree - we do not have a field for `attendanceSubmittedby` - we should add one

tomsmyers (approved):
nice !

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/component/SNSPublisher.kt
sldblog: yeah, we can do that as soon as we add the `appointmentAttendedRecordedBy` and similar "by" fields on appointment

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/component/SNSPublisher.kt
sldblog: hmm. yes

sldblog: will do that in a follow-up PR

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/component/SNSPublisher.kt
sldblog: One useful telemetry would be `InterventionsAuthorizationError` I think

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/component/SNSPublisher.kt
sldblog: done in #322

PR #315, opened 2021-06-09
Explicitly log auth source for mismatched user types

sldblog:

Explicitly log auth source for mismatched user types

To help with understanding how people end up in these branches


tomsmyers (approved):
nice one

PR #316, opened 2021-06-09
disable community api referral integration for running locally

tomsmyers:

disable community api referral integration for running locally

unblock local dev

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/CommunityAPIBookingService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/CommunityAPIBookingService.kt
@@ -14,7 +14,7 @@ import javax.validation.constraints.NotNull
 @Service
 @Transactional
 class CommunityAPIBookingService(
-  @Value("\${appointments.bookings.enabled}") private val bookingsEnabled: Boolean,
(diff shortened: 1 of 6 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/CommunityAPIBookingService.kt
```diff
@@ -14,7 +14,7 @@ import javax.validation.constraints.NotNull
 @Service
 @Transactional
 class CommunityAPIBookingService(
-  @Value("\${appointments.bookings.enabled}") private val bookingsEnabled: Boolean,
+  @Value("\${community-api.appointments.bookings.enabled}") private val bookingsEnabled: Boolean,
```
sldblog, line 17: also has mention in helm values

sldblog (approved):

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/CommunityAPIBookingService.kt
tomsmyers: thanks! value removed from helm since it was set to true (the default) anyway

mahalrmoj (approved):
LGTM

PR #317, opened 2021-06-09
Send custom events about our authorisation problems

sldblog:

Send custom events about our authorisation problems

`UserMapper` now raises `AccessDeniedException` instead of `AccessError` as the latter is now is always in the user's context

So we can monitor the percentage of our user base having issues (and what they are)


tomsmyers (approved):
lgtm

sldblog:
thank you!

PR #320, opened 2021-06-09
IC-1459 initial assessment

rgforsyth:

Separates appointment into own table so it can be reused between action plan and initial assessment.
Creates initial assessments when a referral is sent.
Adds endpoint for initial assessment to be updated with time and duration.

Allows appointments to be generic.
Adds initial assessment/supplier assessment.


helm_deploy/hmpps-interventions-service/reports/crs_performance_report.sql
```diff
@@ -26,9 +26,10 @@ COPY (
     ap.submitted_at         AS date_first_action_plan_submitted,
     'coming-later'          AS date_of_first_action_plan_approval,
     (
-      select min(app.appointment_time)
-      from action_plan_appointment app
-      where app.action_plan_id = ap.id and attended in ('YES', 'LATE')
+        select min(app.appointment_time)
```
rgforsyth, line 29: This sql is currently failing and needs to be re-looked at. It was updated on main since my intitial change

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/AppointmentsService.kt
```diff
@@ -158,20 +173,54 @@ class AppointmentsService(
...
+    )
+    return referralRepository.save(referral)
+  }
+
+  fun updateInitialAssessment(
+    referral: Referral,
+    durationInMinutes: Int?,
+    appointmentTime: OffsetDateTime?
+  ): SupplierAssessment {
+    durationInMinutes?.let { referral.supplierAssessment!!.appointment.durationInMinutes = durationInMinutes }
```
rgforsyth, line 221: Not sure if using '!!' here is the correct way to handle this. There should always be a supplier assessment at this point due to it being created upon sending a referral

PR #321, opened 2021-06-09
IC-1882: Sentence case complexity level and desired outcome description

andrewburrowsmoj:

Changes occurrence of `Service User` to `Service user`

Ensures correct casing is used.

## Notes
Bit of a strange story this. It can be done by: 
1. Transforming on the UI side before presenting to the user
2. Transforming the description in the ComplexityLevelDTO and DesiredOutcomeDTO
3. Making changes directly to the database (via new migration sql script)

Option 3 was chosen because there is less code changes required and we have complete control of the data.

However I think any of these options are valid.


sldblog (commented):
I would vote for 3️⃣ the DB migration path

This is data we control, it doesn't seem to justify additional code + tests

sldblog (approved):

PR #322, opened 2021-06-09
Send domain events even if SNS is disabled

sldblog:

Send domain events as telemetry even if SNS integration is disabled

As per feedback in https://github.com/ministryofjustice/hmpps-interventions-service/pull/312#discussion_r648087392

To make sure telemetry is independent from toggling features

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/component/SNSPublisher.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/component/SNSPublisher.kt
@@ -24,18 +24,22 @@ class SNSPublisher(
   fun publish(referralId: UUID, actor: AuthUser, event: EventDTO) {
     if (enabled) {
       buildRequestAndPublish(event)
-      telemetryClient.trackEvent(
-        "InterventionsDomainEvent",
-        mapOf(
-          "event" to event.eventType,
-          "referralId" to referralId.toString(),
-          "actorUserId" to actor.id,
-        ),
-        null
-      )
     } else {
       logger.debug("Event notification was not published due to sns being disabled.")
     }
+    sendCustomEvent(referralId, actor, event.eventType)
+  }
+
+  private fun sendCustomEvent(referralId: UUID, actor: AuthUser, eventType: String) {
+    telemetryClient.trackEvent(
+      "InterventionsDomainEvent",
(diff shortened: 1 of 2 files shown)


tomsmyers (approved):
nice - is this class still the SNS service or is it now some kinda 'business domain events service'?

sldblog:
yeah, I guess the latter. Should I rename it here?

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/component/SNSPublisher.kt
```diff
@@ -24,18 +24,22 @@ class SNSPublisher(
...
+  private fun sendCustomEvent(referralId: UUID, actor: AuthUser?, eventType: String) {
+    telemetryClient.trackEvent(
+      "InterventionsDomainEvent",
+      mapOf(
+        "event" to eventType,
+        "referralId" to referralId.toString(),
+        "actorUserId" to actor?.id,
+      ),
+      null
+    )
```
andrewburrowsmoj, line 42: Some considerations:
1. Would it be useful to provide metrics on SNS send success/fail?
2. Maybe put it in a finally block incase of any exceptions?

sldblog: Thanks, Andrew

1. I'm assuming we could assess failures from number of exceptions, which should always be 0
2. As this telemetry is only for awareness (real data comes from our database), it doesn't sound too important to guard from exceptional cases

What do you think?

andrewburrowsmoj: These are just considerations without knowing in detail how these metrics are used...

I see that it's more tied to overall 'event' itself rather than metrics of an SNS event.

It may be hard to tie errors to specific eventTypes if wanting to have alerts/metrics of exactly where it's failing. But presumably this can be derived from the exceptions.

sldblog: Yes, the call stack on each exception should tell us exactly where it's failing. Thanks for the think about this, my feeling is that it's good enough for our purposes now

PR #323, opened 2021-06-09
IC-1894: Generate team KPIs

sldblog:

Codify the reporting for team's initial set of (internal) KPIs.

I am _sure_ I am doing this wrong, but we have a semi-convenient way to run these reports every day already (CronJobs + ConfigMaps)

As a bonus, these SQLs also get validated at build time, so they can't fail silently

These will show up in Slack every day:
image

To make sure the entire team can be updated on any KPI we define


helm_deploy/hmpps-interventions-service/templates/cronjob-team-kpis.yaml
```diff
@@ -0,0 +1,72 @@
...
+            args: ["psql -v ON_ERROR_STOP=1 --tuples-only --expanded --file=/reports/team_kpi_report.sql > 'export/team_kpi_report' && /scripts/send_to_slack.sh 'export/team_kpi_report'"]
+            volumeMounts:
+            - name: data-extractor-reports
+              mountPath: /reports/team_kpi_report.sql
+              readOnly: true
+              subPath: team_kpi_report.sql
+            - name: kpi-scripts
+              mountPath: /scripts/send_to_slack.sh
+              readOnly: true
+              subPath: send_to_slack.sh
```
sldblog, line 43: I used these patterns only recently. I learned that:

- During runtime, we make a `/scripts/send_to_slack.sh` file available
- through volume mounts &rarr; volumes
- from the `kpi-scripts` ConfigMap, which contains a key `send_to_slack.sh` and the contents of `helm_deploy/hmpps-interventions-service/reports/send_to_slack.sh` as a value

The end result is... make the file in github available in the runtime env. This is how we get both the report in the PR and the script to the env without packaging it up in Docker

helm_deploy/hmpps-interventions-service/reports/team_kpi_report.sql
```diff
@@ -0,0 +1,43 @@
+WITH counts AS (
+    SELECT count(r.id) FILTER ( WHERE r.sent_at IS NULL )                   AS count_of_draft_referrals,
+           count(r.id) FILTER ( WHERE r.sent_at IS NOT NULL )               AS count_of_sent_referrals,
+           count(r.id) FILTER ( WHERE r.concluded_at IS NOT NULL )          AS count_of_concluded_referrals,
+           count(r.id) FILTER ( WHERE r.concluded_at IS NOT NULL AND
+                                      r.end_requested_at IS NOT NULL AND
+                                      eosr.submitted_at IS NULL )           AS count_of_cancelled_referrals,
```
sldblog, line 7: I suspect a `referral_state` materialised view is ripe now

mahalrmoj (approved):
Like FILTER WHERE too!

helm_deploy/hmpps-interventions-service/reports/team_kpi_report.sql
```diff
@@ -0,0 +1,46 @@
...
+       counts.count_of_sent_referrals,
+       counts.count_of_concluded_referrals,
+       counts.count_of_live_referrals,
+       counts.count_of_cancelled_referrals,
+       counts.count_of_mistaken_referrals,
+       counts.count_of_early_end_referrals,
+       counts.count_of_pp_users_sending_referrals,
+       case
+           when counts.count_of_draft_referrals = 0 then 0.0
+           else 100.0 * counts.count_of_sent_referrals / counts.count_of_draft_referrals
```
sldblog, line 37: :doh: this should be `counts.count_of_sent_referrals / counts.count_all_referrals`

PR #325, opened 2021-06-10
updated 'generic appointment' changes

tomsmyers:

warning - giant pr. please review each commit in isolation. there are some simple renaming refactoring commits early that can be quickly skimmed.

the test changes are almost unreviewable - sorry, i didn't know how to make this any better.

there is a single pact test failure which is legit and needs changing on the UI. 

Preparation work for a shared appointments model across for supplier initial assessments and delivery assessments.


src/main/resources/db/migration/V1_56_1__remove_appointment_fields_from_action_plan.sql
```diff
@@ -0,0 +1,21 @@
...
+    action_plan_session_id uuid not null,
+    appointments_id uuid not null,
+    constraint fk_action_plan_session_appointments_action_plan_session foreign key (action_plan_session_id) references action_plan_session,
+    constraint fk_action_plan_session_appointments_appointment_id foreign key (appointments_id) references appointment,
+    constraint uk_action_plan_session_appointments_appointments_id unique (appointments_id),
+    primary key (action_plan_session_id, appointments_id)
+);
+
+insert into action_plan_session
+    select id, session_number, action_plan_id from action_plan_appointment;
```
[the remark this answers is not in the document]
tomsmyers: we did look into that but the change is really messy and we decided to drop all existing appointments as part of this change.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/ActionPlanSessionsService.kt
```diff
@@ -49,19 +51,41 @@ class ActionPlanSessionsService(
...
+    val existingAppointment = session.currentAppointment
+
+    val deliusAppointmentId = communityAPIBookingService.book(
+      session.actionPlan.referral,
+      existingAppointment,
+      appointmentTime,
+      durationInMinutes
+    )
+
+    if (existingAppointment == null) {
```
rgforsyth, line 72: What's the benefit of doing this here rather than creating an empty appointment when the session was created? Not an issue, just curious

tomsmyers: the main benefit is that it allows the appointment entity to contain non nullable fields for duration and time. so a session can exist in an unscheduled state, but it doesn't really make sense for an appointment exist in that way (since at appoint at it's core is just currently time and duration). 

so conceptually the sessions are just containers for appointments. and to create a session without an appointment is exactly that, a session, with an empty appointment set.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/ActionPlanSessionsService.kt
```diff
@@ -0,0 +1,192 @@
...
+    appointment.attendanceBehaviour = behaviour
+    appointment.attendanceBehaviourSubmittedAt = OffsetDateTime.now()
+    appointment.notifyPPOfAttendanceBehaviour = notifyProbationPractitioner
+  }
+
+  private fun checkSessionIsNotDuplicate(actionPlanId: UUID, sessionNumber: Int) {
+    getActionPlanSession(actionPlanId, sessionNumber)?.let {
+      throw EntityExistsException("Action plan session already exists for [id=$actionPlanId, sessionNumber=$sessionNumber]")
+    }
+  }
```
andrewburrowsmoj, line 184: This is not guaranteed to work as it is not atomic - for instance when the use click on the button twice in quick succession.

Perhaps have a constraint on the database where actionPlanId and sessionNumber has to be unique

tomsmyers: good point - this bug is inherited. shall we fix outside the scope of this PR? would you like to put in a quick PR?

tomsmyers: or do you think i should just include in these changes if we're just talking about a sql constraint?

andrewburrowsmoj: sorry, I just noticed. I personally think create a separate bug ticket is best - as maybe might want to write some unit tests for it which won't be as quick

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/CommunityAPIBookingService.kt
```diff
@@ -66,51 +60,41 @@ class CommunityAPIBookingService(
...
-      isDifferentTimings(existingAppointment, appointmentTime!!, durationInMinutes!!)
-
-  fun isTimingSpecified(appointmentTime: OffsetDateTime?, durationInMinutes: Int?): Boolean =
-    appointmentTime != null && durationInMinutes != null
+  fun isRescheduleBooking(existingAppointment: Appointment, appointmentTime: OffsetDateTime, durationInMinutes: Int): Boolean =
+    isDifferentTimings(existingAppointment, appointmentTime, durationInMinutes)
 
-  fun isDifferentTimings(existingAppointment: ActionPlanAppointment, appointmentTime: OffsetDateTime, durationInMinutes: Int): Boolean =
-    !existingAppointment.appointmentTime!!.isEqual(appointmentTime) || existingAppointment.durationInMinutes != durationInMinutes
+  fun isDifferentTimings(existingAppointment: Appointment, appointmentTime: OffsetDateTime, durationInMinutes: Int): Boolean =
```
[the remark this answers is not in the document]
tomsmyers: yeh i do agree - i left it this way just to make it extra clear what the not so immediately obvious boolean logic is doing

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/ActionPlanSessionsService.kt
tomsmyers: https://dsdmoj.atlassian.net/browse/IC-1897

src/main/resources/db/migration/V1_56_1__remove_appointment_fields_from_action_plan.sql
```diff
@@ -0,0 +1,21 @@
...
+    id uuid not null,
+    session_number int not null,
+    action_plan_id uuid,
+    constraint fk_action_plan_session_action_plan foreign key (action_plan_id) references action_plan,
+    primary key (id)
+);
+
+create table action_plan_session_appointments (
+    action_plan_session_id uuid not null,
+    appointments_id uuid not null,
```
sldblog, line 11: 2 observations
- it sounds like this relationship could be represented in a simpler way with an `action_plan_appointment_v2 { action_plan_id, session_number, appointment_id }` table
- given we won't ever drop sessions later, just add more, we have the option to link these to `referral` instead of the action plan, essentially giving us: `referral_session_appointment { referral_id, session_number, appointment_id }`

sldblog (commented):
random comments, no blockers

src/main/resources/db/migration/V1_56_1__remove_appointment_fields_from_action_plan.sql
tomsmyers: so we would have 

```
appointment with pk id: UUID

action_plan with pk id: UUID

action_plan_session with pk (action_plan_id: UUID, session_number: Int)

action_plan_appointment with pk (appointment_id: UUID)
```

?

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/jpa/entity/ActionPlanSession.kt
```diff
@@ -19,7 +19,7 @@ import javax.validation.constraints.NotNull
 @Entity
 @Table(name = "action_plan_appointment")
 @TypeDef(name = "attended", typeClass = PostgreSQLEnumType::class)
-data class ActionPlanAppointment(
+data class ActionPlanSession(
```
sldblog, line 13: SessionAppointment?

src/main/resources/db/migration/V1_56_1__remove_appointment_fields_from_action_plan.sql
sldblog: was thinking maybe
```
appointment with pk id: UUID

action_plan with pk id: UUID

delivery_appointment with pk (action_plan_id: UUID, session_number: Int)
  field FK (appointment_id: UUID)
```
?

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/jpa/entity/ActionPlanSession.kt
tomsmyers: we talked about this quite a bit and after enumerating all the options decided on this.

we wanted to reservce 'appointment' for only the appointment entity itself. this way reducing confusion about which objects are appointments and which aren't.

src/main/resources/db/migration/V1_56_1__remove_appointment_fields_from_action_plan.sql
tomsmyers: if you flesh this out to support multiple appointments per session i think you kind of end up with what we have here? an extra join table which ends up being an appointment 'container' - i.e. the session?

tomsmyers: i suppose you can just create multiple of the thing you're calling 'delivery_appointment' with the same session number and action plan id?

sldblog: talked offline -- all good 👍 I forgot we have to put delivery method etc somewhere too, and they'll go on the session-related bits

tomsmyers: discussed offline. leaving as is for now.

src/main/resources/db/migration/V1_56_0__new_generic_appointment.sql
```diff
@@ -0,0 +1,23 @@
...
+    created_at timestamp with time zone not null,
+    created_by_id text not null,
+
+    attended attended,
+    additional_attendance_information text,
+    attendance_submitted_at timestamp with time zone,
+
+    attendance_behaviour text,
+    notifyppof_attendance_behaviour boolean,
+    attendance_behaviour_submitted_at timestamp with time zone,
```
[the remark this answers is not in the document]
tomsmyers: i'm going to do this outside this PR. it's just too big to be making any more functional changes.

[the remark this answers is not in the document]
tomsmyers: @sldblog please can you open a ticket and remind me at stand up. thanks

helm_deploy/hmpps-interventions-service/reports/crs_performance_report.sql
```diff
@@ -1,15 +1,19 @@
 COPY (
   WITH attended_sessions AS (
-      select count(action_plan_id) AS attended, action_plan_id
-      from action_plan_appointment
+      select count(aps.action_plan_id) AS attempted, aps.action_plan_id
```
sldblog, line 3:
```suggestion
      select count(app.id) AS attended, aps.action_plan_id
```

helm_deploy/hmpps-interventions-service/reports/crs_performance_report.sql
```diff
@@ -37,7 +43,7 @@ COPY (
     )                       AS outcomes_to_be_achieved_count,
     'coming-later'          AS outcomes_progress,
     ap.number_of_sessions   AS count_of_sessions_expected,
-    shows.attended          AS count_of_sessions_attended,
+    shows                   AS count_of_sessions_attended,
```
sldblog, line 46:
```suggestion
    shows.attended          AS count_of_sessions_attended,
```

sldblog (commented):
One correctness issue in the report, otherwise LGTM
(pacts seem to be broken tho)

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/CommunityAPIBookingService.kt
```diff
@@ -25,37 +26,30 @@ class CommunityAPIBookingService(
 ) : CommunityAPIService {
   companion object : KLogging()
 
-  fun book(existingAppointment: ActionPlanAppointment, appointmentTime: OffsetDateTime?, durationInMinutes: Int?): Long? {
+  fun book(referral: Referral, existingAppointment: Appointment?, appointmentTime: OffsetDateTime, durationInMinutes: Int): Long? {
```
sldblog, line 29: this sounds odd, could it be an appointment that is `book`ed?

sldblog: never mind, I had it muddled up

sldblog:
after talking with @tomsmyers I've rebased + pushed up the proposed SQL report changes -- over to pact fix before :shipit:

rgforsyth (approved):

sldblog (approved):
:shipit: 

(But let's do a smoke test when it's on dev -- before it goes out)

sldblog:
what do we do with the old appointments table? can we rename it to `deprecated_...`?

src/main/resources/db/migration/V1_56_1__remove_appointment_fields_from_action_plan.sql
```diff
@@ -18,4 +18,5 @@ create table action_plan_session_appointments (
 insert into action_plan_session
     select id, session_number, action_plan_id from action_plan_appointment;
 
-drop table action_plan_appointment;
\ No newline at end of file
+alter table action_plan_appointment
+    rename to deprecated_action_plan_appointment;
```
sldblog, line 22: 👍 thanks

PR #327, opened 2021-06-10
IC-1912: 🪲 Count correct completion rate

sldblog:

Fix completion rate calculation introduced in #323 

If we have 14 referrals, 10 drafts and 4 sents

Currently, the completion rate says 40% (4 completed out of 10)
It should say 28.57% (4 completed out of 14)

I'm not going to unit test this -- but once we have breathing space,
this is one of the best candidates to be moved to a reporting component,
possibly a section of our API, HTTP or otherwise


helm_deploy/hmpps-interventions-service/reports/team_kpi_report.sql
```diff
@@ -1,5 +1,6 @@
 WITH counts AS (
-    SELECT count(r.id) FILTER ( WHERE r.sent_at IS NULL )                   AS count_of_draft_referrals,
+    SELECT count(r.id)                                                      AS count_of_started_referrals,
```
sldblog, line 2: `count_of_started_referrals` instead of "all referrals" because when introduce date filtering, we'll do a `started_at` between earliest, latest

PR #330, opened 2021-06-11
IC-1891 Fix for exception when non-attendance and there is no session feedback

mahalrmoj:
When non-attendance and there is no session feedback an exception is thrown. This is a fix to deal with this scenario.

Bug fix.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/AppointmentsService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/AppointmentsService.kt
@@ -119,8 +119,12 @@ class AppointmentsService(
     actionPlanAppointmentRepository.save(appointment)
 
     appointmentEventPublisher.attendanceRecordedEvent(appointment, appointment.attended == Attended.NO)
-    appointmentEventPublisher.behaviourRecordedEvent(appointment, appointment.notifyPPOfAttendanceBehaviour!!)
-    appointmentEventPublisher.sessionFeedbackRecordedEvent(appointment, appointment.notifyPPOfAttendanceBehaviour!!)
+
+    if (appointment.notifyPPOfAttendanceBehaviour != null) {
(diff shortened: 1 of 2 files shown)


src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/AppointmentsServiceTest.kt
```diff
@@ -375,6 +375,42 @@ internal class AppointmentsServiceTest {
     verify(appointmentEventPublisher).sessionFeedbackRecordedEvent(appointment, true)
   }
 
+  @Test
+  fun `attendance is submitted (without feedback)`() {
```
sldblog, line 379: can we include the expected outcome in the test description too, please?

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/AppointmentsService.kt
```diff
@@ -119,8 +119,12 @@ class AppointmentsService(
     actionPlanAppointmentRepository.save(appointment)
 
     appointmentEventPublisher.attendanceRecordedEvent(appointment, appointment.attended == Attended.NO)
-    appointmentEventPublisher.behaviourRecordedEvent(appointment, appointment.notifyPPOfAttendanceBehaviour!!)
-    appointmentEventPublisher.sessionFeedbackRecordedEvent(appointment, appointment.notifyPPOfAttendanceBehaviour!!)
+
+    if (appointment.notifyPPOfAttendanceBehaviour != null) {
+      appointmentEventPublisher.behaviourRecordedEvent(appointment, appointment.notifyPPOfAttendanceBehaviour!!)
```
sldblog, line 124: this code seems to say: there is no behaviour at all if we don't want to notify about it -- is that correct?

mahalrmoj:
Invalidated by big merge

PR #331, opened 2021-06-11
add the concept of approvals to action plans

tomsmyers:

add action plan 'approvedBy' and 'approvedAt' fields
adds controller / service methods for approving action plans 
stops session creation on action plan submission and moves it to action plan approval
adds sns and notify events for action plan approval

initially, the intent is just to stop action plans being implicitly approved on submission. this is the start of more work to come.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/controller/ActionPlanController.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/controller/ActionPlanController.kt
@@ -10,7 +10,9 @@ import org.springframework.web.bind.annotation.RequestBody
 import org.springframework.web.bind.annotation.RequestParam
 import org.springframework.web.bind.annotation.RestController
 import uk.gov.justice.digital.hmpps.hmppsinterventionsservice.authorization.UserMapper
+import uk.gov.justice.digital.hmpps.hmppsinterventionsservice.authorization.UserTypeChecker
 import uk.gov.justice.digital.hmpps.hmppsinterventionsservice.component.LocationMapper
+import uk.gov.justice.digital.hmpps.hmppsinterventionsservice.config.AccessError
 import uk.gov.justice.digital.hmpps.hmppsinterventionsservice.controller.mappers.ActionPlanMapper
 import uk.gov.justice.digital.hmpps.hmppsinterventionsservice.dto.ActionPlanDTO
 import uk.gov.justice.digital.hmpps.hmppsinterventionsservice.dto.CreateActionPlanDTO
@@ -25,7 +27,8 @@ class ActionPlanController(
   val actionPlanMapper: ActionPlanMapper,
   val userMapper: UserMapper,
   val actionPlanService: ActionPlanService,
-  val locationMapper: LocationMapper
+  val locationMapper: LocationMapper,
+  val userTypeChecker: UserTypeChecker,
 ) {
 
   @PostMapping("/draft-action-plan")
@@ -100,5 +103,16 @@ class ActionPlanController(
   ): ActionPlanDTO {
     val actionPlan = actionPlanService.getActionPlanByReferral(referralId)
     return ActionPlanDTO.from(actionPlan)
+  }
+
+  @PostMapping("/action-plan/{id}/approve")
+  fun approveActionPlan(@PathVariable id: UUID, authentication: JwtAuthenticationToken): ActionPlanDTO {
+    val user = userMapper.fromToken(authentication)
+
+    if (!userTypeChecker.isProbationPractitionerUser(user)) {
+      throw AccessError(user, "could not approve action plan", listOf("only probation practitioners can approve action plans"))
+    }
+
+    return ActionPlanDTO.from(actionPlanService.approveActionPlan(id, user))
   }
 }
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/ActionPlanService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/ActionPlanService.kt
@@ -1,5 +1,6 @@
 package uk.gov.justice.digital.hmpps.hmppsinterventionsservice.service
 
+import org.springframework.data.repository.findByIdOrNull
 import org.springframework.stereotype.Service
 import uk.gov.justice.digital.hmpps.hmppsinterventionsservice.component.ActionPlanValidator
 import uk.gov.justice.digital.hmpps.hmppsinterventionsservice.events.ActionPlanEventPublisher
@@ -95,9 +96,8 @@ class ActionPlanService(
   }
 
   fun getActionPlan(id: UUID): ActionPlan {
-    return actionPlanRepository.findById(id).orElseThrow {
-      throw EntityNotFoundException("action plan not found [id=$id]")
-    }
+    return actionPlanRepository.findByIdOrNull(id)
+      ?: throw EntityNotFoundException("action plan not found [id=$id]")
   }
 
   fun getActionPlanByReferral(referralId: UUID): ActionPlan {
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/ActionPlanService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/ActionPlanService.kt
@@ -77,8 +77,6 @@ class ActionPlanService(
     val submittedActionPlan = updateDraftActionPlanAsSubmitted(draftActionPlan, submittedByUser)
     actionPlanValidator.validateSubmittedActionPlan(submittedActionPlan)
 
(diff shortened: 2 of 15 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/ActionPlanService.kt
```diff
@@ -77,14 +77,23 @@ class ActionPlanService(
     val submittedActionPlan = updateDraftActionPlanAsSubmitted(draftActionPlan, submittedByUser)
     actionPlanValidator.validateSubmittedActionPlan(submittedActionPlan)
 
-    actionPlanSessionsService.createUnscheduledSessionsForActionPlan(submittedActionPlan)
```
sldblog, line 80: if our condition remains "allow bookings when there's a submitted plan", we have to keep doing this after submission -- it doesn't sound like we can change this now

sldblog (commented):
LGTM -- it's essentially finished, isn't it?

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/ActionPlanService.kt
tomsmyers: ah it was my understanding that we couldn't book anything until approval. i must have misheard. can we get confirmation on this? if the whole action plan approval process is motivated by a billing concern, i wouldn't expect we should allow SPs to book appointments for any proposed sessions until approval by PP.

sldblog: confirmed: you're right -- given it's more complex for us to reduce booked sessions

**decision**: enable bookings of appointments once action plan is approved

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/controller/ActionPlanController.kt
```diff
@@ -101,4 +104,15 @@ class ActionPlanController(
     val actionPlan = actionPlanService.getActionPlanByReferral(referralId)
     return ActionPlanDTO.from(actionPlan)
   }
+
+  @PostMapping("/action-plan/{id}/approve")
+  fun approveActionPlan(@PathVariable id: UUID, authentication: JwtAuthenticationToken): ActionPlanDTO {
+    val user = userMapper.fromToken(authentication)
+
+    if (!userTypeChecker.isProbationPractitionerUser(user)) {
+      throw AccessError(user, "could not approve action plan", listOf("only probation practitioners can approve action plans"))
```
tomsmyers, line 113: we don't have these kinda checks everywhere, should we?

mahalrmoj: I think it is necessary otherwise an SP could do it if they knew the link?

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/ActionPlanService.kt
```diff
@@ -77,18 +78,26 @@ class ActionPlanService(
...
     actionPlanEventPublisher.actionPlanSubmitEvent(savedSubmittedActionPlan)
 
     return savedSubmittedActionPlan
   }
 
+  fun approveActionPlan(id: UUID, user: AuthUser): ActionPlan {
+    val actionPlan = getActionPlan(id)
+    actionPlan.approvedAt = OffsetDateTime.now()
+    actionPlan.approvedBy = authUserRepository.save(user)
+
```
mahalrmoj, line 91: Should we check somewhere that the plan has been submitted first or do we continue to rely on the methods being called by the UI are in the right sequence?

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/controller/ActionPlanController.kt
mahalrmoj: Can it be approved by any PP?

mahalrmoj (commented):
My take from this is we still have the delius notification to do.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/controller/ActionPlanController.kt
tomsmyers: we'll have to follow up on that... for now let's say yes.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/ActionPlanService.kt
tomsmyers: yeh i don't think it's nessecary - althought to enforce this we could change `getActionPlan` to only return submitted action plans, kind of like the way we look up sent referrals vs draft referrals.

tomsmyers:
@rajivmahal1 good shout - i didn't realize we have to notify delius on this. we can do it in a different PR.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/controller/ActionPlanController.kt
andrewburrowsmoj: nit: We are using userTypeChecker in ReferralService but not ReferralController. I think it would be best to be consistent (I think controller is more suitable)

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/ActionPlanSessionsService.kt
```diff
@@ -29,10 +29,10 @@ class ActionPlanSessionsService(
   val communityAPIBookingService: CommunityAPIBookingService,
   val appointmentRepository: AppointmentRepository,
 ) {
-  fun createUnscheduledSessionsForActionPlan(submittedActionPlan: ActionPlan) {
-    val numberOfSessions = submittedActionPlan.numberOfSessions!!
+  fun createUnscheduledSessionsForActionPlan(approvedActionPlan: ActionPlan) {
+    val numberOfSessions = approvedActionPlan.numberOfSessions!!
     for (i in 1..numberOfSessions) {
-      createSession(submittedActionPlan, i)
+      createSession(approvedActionPlan, i)
```
andrewburrowsmoj, line 35: Do you think we should have protection against duplicated sessions? I think this can only be solved by applying a uniqueness constraint on action_plan_id and session_number

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/NotifyActionPlanServiceTest.kt
```diff
@@ -43,6 +43,7 @@ class NotifyActionPlanServiceTest {
 
   private fun notifyService(): NotifyActionPlanService {
     return NotifyActionPlanService(
+      "template",
```
andrewburrowsmoj, line 46: I think we should have a new test here for approved events, currently only assigned events are being tested

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/controller/ActionPlanController.kt
tomsmyers: sure thing - i'll move it

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/NotifyActionPlanServiceTest.kt
tomsmyers: good spot

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/ActionPlanSessionsService.kt
tomsmyers: maybe - although for now it's not a concern because this is the only place action plan sessions are created, and they are done sequentially in this loop

andrewburrowsmoj: What if the user double clicks on the approval button?

andrewburrowsmoj (approved):

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/ActionPlanSessionsService.kt
tomsmyers: good point ! i've added a unique key in at the db level to protect against this scenario.

PR #332, opened 2021-06-11
IC-1891 Prevent exception by checking attendance behaviour provided b…

mahalrmoj:
…efore sending event

When non-attendance, there is no session feedback, and the code throws an exception as it assumes it does exist. This is a fix to deal with this scenario.

Bug fix

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/ActionPlanSessionsService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/ActionPlanSessionsService.kt
@@ -144,8 +144,12 @@ class ActionPlanSessionsService(
     actionPlanSessionRepository.save(session)
 
     appointmentEventPublisher.attendanceRecordedEvent(session, appointment.attended!! == Attended.NO)
-    appointmentEventPublisher.behaviourRecordedEvent(session, appointment.notifyPPOfAttendanceBehaviour!!)
-    appointmentEventPublisher.sessionFeedbackRecordedEvent(session, appointment.notifyPPOfAttendanceBehaviour!!)
+
+    if (appointment.attendanceBehaviourSubmittedAt != null) { // excluding the case of non attendance
+      appointmentEventPublisher.behaviourRecordedEvent(session, appointment.notifyPPOfAttendanceBehaviour!!)
+    }
+
+    appointmentEventPublisher.sessionFeedbackRecordedEvent(session, appointment.notifyPPOfAttendanceBehaviour ?: false)
     return session
   }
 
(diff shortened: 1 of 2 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/ActionPlanSessionsService.kt
```diff
@@ -144,8 +144,12 @@ class ActionPlanSessionsService(
     actionPlanSessionRepository.save(session)
 
     appointmentEventPublisher.attendanceRecordedEvent(session, appointment.attended!! == Attended.NO)
-    appointmentEventPublisher.behaviourRecordedEvent(session, appointment.notifyPPOfAttendanceBehaviour!!)
-    appointmentEventPublisher.sessionFeedbackRecordedEvent(session, appointment.notifyPPOfAttendanceBehaviour!!)
+
+    if (appointment.attendanceBehaviourSubmittedAt != null) { // excluding the case of non attendance
+      appointmentEventPublisher.behaviourRecordedEvent(session, appointment.notifyPPOfAttendanceBehaviour!!)
```
sldblog, line 149: following the call chain:

given we have https://github.com/ministryofjustice/hmpps-interventions-service/blob/65ebe25d9e8648d464c3a64646b309314143da43/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/NotifyService.kt#L92-L95

https://github.com/ministryofjustice/hmpps-interventions-service/blob/65ebe25d9e8648d464c3a64646b309314143da43/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/CommunityAPIService.kt#L221-L224

https://github.com/ministryofjustice/hmpps-interventions-service/blob/65ebe25d9e8648d464c3a64646b309314143da43/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/SNSService.kt#L21-L25

It seems here we are cutting off _all_ event notifications. Could we do `appointment.notifyPPOfAttendanceBehaviour ?: false` here too, instead?

mahalrmoj: I see what you are saying, however in reality there is no Behaviour Feedback event because the guy didn't attend? Should the code not reflect that and not raise one at all.

sldblog (approved):
😅 just noticed the condition says `attendanceBehaviourSubmittedAt != null`
that makes sense

👍

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/ActionPlanSessionsService.kt
sldblog: yep, I misread the condition. my bad!

PR #334, opened 2021-06-11
IC-1886: Functionality to retrieve and update appointment details for location

andrewburrowsmoj:

Provide functionality to retrieve and update appointment details for location

Allows users of API to retrieve and update details about the location for their current appointment for the session.

## Note
In order to ensure contracts don't break the appointment update can accept an nullable appointment delivery type this should be removed after changes are merged on the front end


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/config/ErrorConfiguration.kt
```diff
@@ -22,7 +22,11 @@ enum class Code {
   CONDITIONAL_FIELD_MUST_BE_SET,
   CANNOT_BE_EMPTY,
   CANNOT_BE_NEGATIVE_OR_ZERO,
-  INVALID_SERVICE_CATEGORY_FOR_CONTRACT
+  INVALID_SERVICE_CATEGORY_FOR_CONTRACT,
+  ADDRESS_MUST_BE_PROVIDED,
```
sldblog, line 26: feels like this is covered by `CANNOT_BE_EMPTY`

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/config/ErrorConfiguration.kt
```diff
@@ -22,7 +22,11 @@ enum class Code {
...
+  INVALID_NPS_OFFICE_CODE,
+  INVALID_ADDRESS_FIELDS_COUNT,
+  INVALID_POST_CODE,
```
sldblog, line 29: It would be good to keep this `Code` ... i.e. type of error list tiny -- the errors are already scoped to the field, so we don't need new codes, right?

What do you think about using `INVALID_FORMAT` instead of `INVALID_<specific_field>`?

sldblog (commented):
👏 good stuff

pending on discussions

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/config/ErrorConfiguration.kt
andrewburrowsmoj: I thought the caller would want to know why it's invalid. For IN_PERSON_MEETING_OTHER it could be for the following reasons:

1. The delivery address is empty or not provided
2. The delivery address has incorrect number of lines
3. The delivery address has an invalid post code

We can perhaps use the same code as 'INVALID_FORMAT' but it would help to also include a message along with that; but that would mean changing the FieldError object to accept an optional message field. 
How do you feel about that?

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/config/ErrorConfiguration.kt
andrewburrowsmoj: I agree. I'll change to this.

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/ActionPlanSessionsServiceRespositoryTest.kt
```diff
@@ -0,0 +1,174 @@
...
+        testNPSOffice(actionPlanSession)
+      }
+
+      @Test
+      fun `can create new appointment with as nps office appointment`() {
+        val actionPlanSession = actionPlanSessionFactory.createUnscheduled()
+        testNPSOffice(actionPlanSession)
+      }
+
+      private fun testNPSOffice(actionPlanSession: ActionPlanSession) {
```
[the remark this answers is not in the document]
andrewburrowsmoj: Sure I've changed to this wording in the tests

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/config/ErrorConfiguration.kt
sldblog: Sounds good! Allowing for the field, the kind of problem, the message feels 👍 to me

Could still be serialised to `{"field": "...", "error": "$code: $message"}` but doesn't increase coupling to specific fields

sldblog (approved):
👍 nice

Number of error codes could be reduced in a following PR so it's not blocking

PR #336, opened 2021-06-11
IC-1895: Move incoming referral email addresses to interventions

sldblog:

🐾 Moves incoming referrals email address to be defined on individual interventions.
💾 No current emails exist, but we data migrate to retain the testing environments' integrity
💅 The last two commits deal with random warnings I came across.

IC-1895: When service providers fill out their intervention detail forms it has an entry for a receiver e-mail address. This change reflects this model of input.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/jpa/entity/Intervention.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/jpa/entity/Intervention.kt
@@ -15,6 +15,7 @@ data class Intervention(
 
   @NotNull val title: String,
   @NotNull val description: String,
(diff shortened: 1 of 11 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/jpa/entity/Intervention.kt
```diff
@@ -15,6 +15,7 @@ data class Intervention(
 
   @NotNull val title: String,
   @NotNull val description: String,
+  @NotNull val incomingReferralDistributionEmail: String,
```
sldblog, line 18: Once we have all emails, we can add `@javax.validation.constraints.Email` here

tomsmyers (approved):

PR #346, opened 2021-06-15
ensure 'selected service categories' are updated, not replaced on draft referral update

tomsmyers:

ensure 'selected service categories' are updated, not replaced on draft referral update

fix sql constraint issues - see commit message for more details


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/ReferralService.kt
```diff
@@ -328,7 +328,20 @@ class ReferralService(
...
 
     update.serviceCategoryIds?.let { serviceCategoryIds ->
-      referral.selectedServiceCategories = serviceCategoryRepository.findByIdIn(serviceCategoryIds)
+      val updatedServiceCategories = serviceCategoryRepository.findByIdIn(serviceCategoryIds)
+      // the following code looks like a strange way to do this ...
+      // however we have to be very intentional about updating the service categories,
+      // rather than replacing the collection entirely. in the former case, hibernate
+      // runs an update as you would expect. in the latter case, hibernate deletes the
+      // entity entirely and creates a new one. this causes SQL constraint violations
+      // because these service category ids are referenced in the selected complexity
```
mahalrmoj, line 337: Not sure whether it helps but have you tried cascade delete across the relationship which violates the constraint?

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/jpa/entity/Referral.kt
```diff
@@ -79,7 +79,7 @@ class Referral(
     joinColumns = [JoinColumn(name = "referral_id")],
     inverseJoinColumns = [JoinColumn(name = "service_category_id")]
   )
-  var selectedServiceCategories: Set<ServiceCategory>? = null,
+  var selectedServiceCategories: MutableSet<ServiceCategory>? = null,
```
mahalrmoj, line 82: As a mutable list can that be set to non nullable and initialised with an empty set?

sldblog:
Is it possible to add a test that reproduces the problem this solves? Ie stuff already selected with complexities etc

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/ReferralService.kt
sldblog: In this case we didn't want to delete those, afaik, it was semantically a no-op here but was running as a delete/insert

andrewburrowsmoj:
Hi David... I have just pushed in a test.

On main branch the test fails with:


```
Caused by: org.postgresql.util.PSQLException: ERROR: update or delete on table "referral_selected_service_category" violates foreign key constraint "fk_referral_desired_outcome_service_category_id" on table "referral_desired_outcome"
  Detail: Key (referral_id, service_category_id)=(ffab23c5-3dc1-4895-afa7-f7f9949f62ca, 0ea11fa7-d70a-4983-8ba4-a6c22bb3ed72) is still referenced from table "referral_desired_outcome".
```

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/jpa/entity/Referral.kt
andrewburrowsmoj: I believe we still want to maintain the concept of a null. This indicates that the user has not selected anything yet, which is different than selecting nothing.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/ReferralService.kt
tomsmyers: yes david's correct

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/jpa/entity/Referral.kt
tomsmyers: for collection types spring returns empty collections for lookups anyway, so we actually have some code in the DTO layer to translate empty collections into null values where they don't make sense to be empty collections. it's a bit of a mess tbh. but either way it doesn't really matter here. let's make the changes as minimal as possible and leave the init to null.

mahalrmoj (approved):

sldblog:
🎉

PR #351, opened 2021-06-17
Include event logs analytics extract

green-digger:

Sets `SAVE_EVENTS_LOG` in the job to extract data to the analytical platform

Removes redundant `PGPORT` variables as process now takes the default value if none is provided.

To transfer the logs that capture contextual information about the extraction to aid with monitoring its usage.

### DIFF
--- a/helm_deploy/hmpps-interventions-service/templates/cronjob-data-extractor-analytics.yaml
+++ b/helm_deploy/hmpps-interventions-service/templates/cronjob-data-extractor-analytics.yaml
@@ -15,8 +15,6 @@ spec:
             imagePullPolicy: Always
             args: ["extract_table_names.py && extract_pg_jsonl_snapshot.py && transfer_local_to_s3.sh"]
             env:
-              - name: PGPORT
-                value: "5432"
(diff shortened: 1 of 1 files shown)


tomsmyers (approved):
lgtm

helm_deploy/hmpps-interventions-service/templates/cronjob-data-extractor-analytics.yaml
```diff
@@ -54,3 +52,5 @@ spec:
                     key: secret_access_key
               - name: AWS_DEFAULT_REGION
                 value: eu-west-2
+          - name: SAVE_EVENTS_LOG
+            value: "true"
```
sldblog, line 56: malformed YAML

```
Error: UPGRADE FAILED: error validating "": error validating data: ValidationError(CronJob.spec.jobTemplate.spec.template.spec.containers[1]): unknown field "value" in io.k8s.api.core.v1.Container
```

sldblog:
why wouldn't `helm-lint` capture this? it's literally its job 😂

PR #352, opened 2021-06-17
IC-1942: multiple provider groups

sldblog:

Enables authenticated service provider users to access the service with multiple provider groups

Providers users are being setup with multiple `INT_SP_` provider groups:

We have a user that belongs to **all** of these groups:
image

This is a result of insufficient shared understanding, sorry about that.
A review of the authorization model will be booked next month

So -- multiple provider groups have to provide access to:

- referrals that have any of those providers as prime
- referrals that have any of those providers as subcontractor

Misconfiguration issues to be raised:

- because we now allow a combination of things, be explicit when a contract group is found which has no relation to any of the provider groups:

For example, error if someone has the groups represented by the purple set:
image

Trade-offs:

- with multi-provider memberships, it will be impossible to figure which provider someone is acting on behalf of

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/authorization/ServiceProviderAccessScopeMapper.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/authorization/ServiceProviderAccessScopeMapper.kt
@@ -10,14 +10,14 @@ import uk.gov.justice.digital.hmpps.hmppsinterventionsservice.jpa.repository.Ser
 import uk.gov.justice.digital.hmpps.hmppsinterventionsservice.service.HMPPSAuthService
 
 data class ServiceProviderAccessScope(
-  val serviceProviders: List<ServiceProvider>,
-  val contracts: List<DynamicFrameworkContract>,
+  val serviceProviders: Set<ServiceProvider>,
+  val contracts: Set<DynamicFrameworkContract>,
 )
 
 private data class WorkingScope(
   val authGroups: List<String>,
-  val providers: MutableList<ServiceProvider> = mutableListOf(),
-  val contracts: MutableList<DynamicFrameworkContract> = mutableListOf(),
+  val providers: MutableSet<ServiceProvider> = mutableSetOf(),
+  val contracts: MutableSet<DynamicFrameworkContract> = mutableSetOf(),
   val errors: MutableList<String> = mutableListOf(),
 )
 
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/authorization/ServiceProviderAccessScopeMapper.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/authorization/ServiceProviderAccessScopeMapper.kt
@@ -1,6 +1,5 @@
 package uk.gov.justice.digital.hmpps.hmppsinterventionsservice.authorization
 
-import org.springframework.data.repository.findByIdOrNull
 import org.springframework.stereotype.Component
 import uk.gov.justice.digital.hmpps.hmppsinterventionsservice.config.AccessError
 import uk.gov.justice.digital.hmpps.hmppsinterventionsservice.jpa.entity.AuthUser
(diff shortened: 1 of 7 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/authorization/ServiceProviderAccessScopeMapper.kt
```diff
@@ -11,8 +10,15 @@ import uk.gov.justice.digital.hmpps.hmppsinterventionsservice.jpa.repository.Ser
 import uk.gov.justice.digital.hmpps.hmppsinterventionsservice.service.HMPPSAuthService
 
 data class ServiceProviderAccessScope(
-  val serviceProvider: ServiceProvider,
-  val contracts: List<DynamicFrameworkContract>,
+  val serviceProviders: Set<ServiceProvider>,
+  val contracts: Set<DynamicFrameworkContract>,
+)
+
+data class WorkingScope(
```
tomsmyers, line 17: small nit: should these be sets too?

tomsmyers: can we make this class private?

sldblog: done (fixed up, so no individual commit

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/authorization/ServiceProviderAccessScopeMapper.kt
```diff
@@ -34,63 +40,81 @@ class ServiceProviderAccessScopeMapper(
...
-        val providerGroupCode = serviceProviderGroups[0]
-        val provider = serviceProviderRepository.findByIdOrNull(providerGroupCode)
-        if (provider == null) {
-          configErrors.add("service provider id '$providerGroupCode' does not exist in the interventions database")
-        }
-        provider
-      }
+  private fun getProviders(providerGroups: List<String>, configErrors: MutableList<String>): List<ServiceProvider> {
+    val providers = serviceProviderRepository.findAllById(providerGroups)
+    val removedProviders = providerGroups.subtract(providers.map { it.id })
```
tomsmyers, line 105: i think here and in `getContracts` the term 'removed' is a little misleading. nothing is getting removed, these are just sets of providers and contracts that exist in the user's auth config, but not in the interventions database. so these are more like 'unknownProviders' or 'unresolvedProviders' or something. then the error message could be something like "provider group {} was not added to access scope because it does not exist in the refer and monitor database"

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/authorization/ServiceProviderAccessScopeMapper.kt
```diff
@@ -34,63 +40,81 @@ class ServiceProviderAccessScopeMapper(
...
 
-    val contracts = getContracts(contractGroups, configErrors)
+    val contracts = getContracts(contractGroups, scope.errors)
+    scope.contracts.addAll(contracts)
+  }
 
-    // FIXME we also need to remove contracts which do not belong to the user's provider
+  private fun removeInaccessibleContracts(scope: WorkingScope) {
+    val orphanContracts = scope.contracts.filterNot {
+      scope.providers.contains(it.primeProvider) || it.subcontractorProviders.intersect(scope.providers).isNotEmpty()
```
tomsmyers, line 83: i think would be easier to read with `it` being named and without the extra `not` to grep

```
val contractsNotAssosciatedWithUserProviders = contracts.filter { contract ->
      val validProviders = setOf(contract.primeProvider) + contract.subcontractorProviders
      scope.providers.intersect(validProviders).isNotEmpty()
}
```

tomsmyers (approved):
a few comments, happy to get this merged and iterate on improving user facing comms

sldblog:
Thank you, will amend as per comments

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/authorization/ServiceProviderAccessScopeMapper.kt
sldblog: done in 94cb6204470d325c967d5340b6914d0c1a8a1c0a -- I opted for "unidentified"

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/authorization/ServiceProviderAccessScopeMapper.kt
sldblog: done in a897e8b7f41e739e9787cb9262742d68e990ee12

PR #353, opened 2021-06-17
IC-1943: Make appointment delivery mandatory + Pact tests

andrewburrowsmoj:

Removes the ability to specify an optional appointment delivery type when updating an appointment.

Ensure that user can not edit an appointment without also supplying where it will be delivered

## Note

This is will also require an update on UI side, this will be done in PR https://github.com/ministryofjustice/hmpps-interventions-ui/pull/436

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/dto/ActionPlanSessionsDTO.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/dto/ActionPlanSessionsDTO.kt
@@ -1,5 +1,6 @@
 package uk.gov.justice.digital.hmpps.hmppsinterventionsservice.dto
 
+import com.fasterxml.jackson.annotation.JsonCreator
 import com.fasterxml.jackson.annotation.JsonProperty
 import uk.gov.justice.digital.hmpps.hmppsinterventionsservice.jpa.entity.ActionPlanSession
 import uk.gov.justice.digital.hmpps.hmppsinterventionsservice.jpa.entity.AppointmentDeliveryType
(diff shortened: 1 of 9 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/dto/ActionPlanSessionsDTO.kt
```diff
@@ -9,8 +9,7 @@ import java.util.UUID
 data class UpdateAppointmentDTO(
   val appointmentTime: OffsetDateTime,
   val durationInMinutes: Int,
-  // TODO: remove optional when front-end changes are complete
```
tomsmyers, line 12: nice to see these todos getting removed 👍 😄

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/integration/SetupAssistant.kt
```diff
@@ -259,10 +270,13 @@ class SetupAssistant(
     attended: Attended? = null,
     attendanceInfo: String? = null,
     behaviour: String? = null,
-    notifyPPOfBehaviour: Boolean? = null
+    notifyPPOfBehaviour: Boolean? = null,
+    appointmentDeliveryType: AppointmentDeliveryType? = null,
+    appointmentDeliveryAddressLines: List<String>? = null,
```
tomsmyers, line 275: is this list of strings just used in the tests? the data we get in the API call is more structured than this right?

tomsmyers (approved):
✅

PR #355, opened 2021-06-18
slight tweak to ensure submitted action plan has number of sessions set

tomsmyers:

slight tweak to ensure submitted action plan has number of sessions set

contract tests


sldblog (approved):
LGTM, for posterity, would be useful to know which `@State`s this affects

PR #357, opened 2021-06-18
IC-1951 Added End Of Service Report Flag to single sent referral responses

mahalrmoj:
A derived field which indicates to the user interface if a EoS Report is required. 

To avoid duplicating the business rule in UI

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/controller/ReferralController.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/controller/ReferralController.kt
@@ -64,7 +64,7 @@ class ReferralController(
     )
     return SentReferralDTO.from(
       referralService.assignSentReferral(sentReferral, assignedBy, assignedTo),
-      referralConcluder.requiresEndOfServiceReport(sentReferral)
+      referralConcluder.requiresEndOfServiceReportCreation(sentReferral)
     )
   }
 
@@ -85,14 +85,14 @@ class ReferralController(
 
     return ResponseEntity
       .created(location)
-      .body(SentReferralDTO.from(sentReferral, referralConcluder.requiresEndOfServiceReport(sentReferral)))
+      .body(SentReferralDTO.from(sentReferral, referralConcluder.requiresEndOfServiceReportCreation(sentReferral)))
   }
 
   @JsonView(Views.SentReferral::class)
   @GetMapping("/sent-referral/{id}")
   fun getSentReferral(@PathVariable id: UUID, authentication: JwtAuthenticationToken): SentReferralDTO {
     val referral = getSentReferralForAuthenticatedUser(authentication, id)
-    return SentReferralDTO.from(referral, referralConcluder.requiresEndOfServiceReport(referral))
+    return SentReferralDTO.from(referral, referralConcluder.requiresEndOfServiceReportCreation(referral))
   }
 
   @JsonView(Views.SentReferral::class)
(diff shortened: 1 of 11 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/ReferralConcluder.kt
```diff
@@ -64,4 +66,10 @@ class ReferralConcluder(
 
     return null
   }
+
+  private fun getNumberOfAttendedAppointments(referral: Referral): Int {
+    return referral.actionPlan?.let {
+      actionPlanRepository.countNumberOfAttendedSessions(it.id)
```
sldblog, line 72: Can we rename this to `countSessionsWithAttendanceRecorded` (or similar), please? "Attended" in domain terminology is when people turn up to their appointment

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/ReferralConcluder.kt
```diff
@@ -35,13 +35,15 @@ class ReferralConcluder(
     }
   }
 
+  fun requiresEosr(referral: Referral): Boolean {
+    return getNumberOfAttendedAppointments(referral) > 0
```
sldblog, line 39: And it's concluded, right?

sldblog:
👍 💯 for exposing the business rules in the business API

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/ReferralConcluder.kt
mahalrmoj: Will do

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/ReferralConcluder.kt
mahalrmoj: yes

sldblog: Can we add those conditions?

mahalrmoj: Added a checked that the referral has infact ended for some reason

rgforsyth (approved):

sldblog (approved):

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/controller/ReferralController.kt
```diff
@@ -81,22 +85,23 @@ class ReferralController(
 
     return ResponseEntity
       .created(location)
-      .body(SentReferralDTO.from(sentReferral))
+      .body(SentReferralDTO.from(sentReferral, referralConcluder.requiresEndOfServiceReportCreation(sentReferral)))
```
tomsmyers, line 88: nit: i think it's safe to assume that freshly sent referrals don't require EOSRs, but it's fine to do this anyway.

tomsmyers (approved):

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/ReferralConcluder.kt
mahalrmoj: Agreed logic ... true if:

endOfServiceReport does not exist 
and
(allSessionsDelivered) || (endRequested && atLeastOneSessionDelivered)

PR #359, opened 2021-06-18
Supplier assessment behaviour

rgforsyth:

Adds endpoint so that appointment behaviour can be submitted.

Behaviour needs to be able to be recorded against appointments for both action plan sessions & supplier assessments


rgforsyth:
Branched from supplier assessment pr branch(IC-1459) which should to go in first.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/jpa/entity/Appointment.kt
```diff
@@ -53,6 +53,6 @@ enum class Attended {
 
   @JsonValue
   open fun toLower(): String? {
-    return this.toString().lowercase()
+    return this.toString().toLowerCase()
```
tomsmyers: intellij be damned!

tomsmyers:
looks like this has a bunch of commits shared with your other PR. will hold off on review until the other is merged and this is rebased to avoid confusion

PR #362, opened 2021-06-21
IC-1722 Supplier Assessment booking requests in nDelius

mahalrmoj:
Contains the logic to request a booking for the supplier assessment appointment made in R&M 

With this PR supplier assessments are also notified to Delius so all appointments pertaining to a CRN are maintained in Delius for clash checking.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/AppointmentService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/AppointmentService.kt
@@ -31,22 +31,19 @@ class AppointmentService(
     val initialAppointmentRequired = appointment == null
     if (initialAppointmentRequired) {
       val deliusAppointmentId = communityAPIBookingService.book(referral, null, appointmentTime, durationInMinutes, appointmentType)
-      val createdAppointment = createAppointment(durationInMinutes, appointmentTime, deliusAppointmentId, createdByUser)
-      return appointmentRepository.save(createdAppointment)
+      return createAppointment(durationInMinutes, appointmentTime, deliusAppointmentId, createdByUser)
     }
 
     val updateCurrentAppointmentRequired = appointment!!.attended == null
     if (updateCurrentAppointmentRequired) {
       val deliusAppointmentId = communityAPIBookingService.book(referral, appointment, appointmentTime, durationInMinutes, appointmentType)
-      val updatedAppointment = updateAppointment(appointment, durationInMinutes, appointmentTime, deliusAppointmentId)
-      return appointmentRepository.save(updatedAppointment)
+      return updateAppointment(appointment, durationInMinutes, appointmentTime, deliusAppointmentId)
     }
 
     val additionalAppointmentRequired = appointment.attended == Attended.NO
     if (additionalAppointmentRequired) {
       val deliusAppointmentId = communityAPIBookingService.book(referral, null, appointmentTime, durationInMinutes, appointmentType)
-      val additionalAppointment = createAppointment(durationInMinutes, appointmentTime, deliusAppointmentId, createdByUser)
-      return appointmentRepository.save(additionalAppointment)
+      return createAppointment(durationInMinutes, appointmentTime, deliusAppointmentId, createdByUser)
     }
 
     throw IllegalStateException("Is it not possible to update an appointment that has already been attended")
@@ -58,13 +55,15 @@ class AppointmentService(
     deliusAppointmentId: Long?,
     createdByUser: AuthUser,
   ): Appointment {
-    return Appointment(
-      id = UUID.randomUUID(),
-      appointmentTime = appointmentTime,
-      durationInMinutes = durationInMinutes,
-      deliusAppointmentId = deliusAppointmentId,
-      createdBy = authUserRepository.save(createdByUser),
-      createdAt = OffsetDateTime.now()
+    return appointmentRepository.save(
+      Appointment(
+        id = UUID.randomUUID(),
+        appointmentTime = appointmentTime,
+        durationInMinutes = durationInMinutes,
+        deliusAppointmentId = deliusAppointmentId,
+        createdBy = authUserRepository.save(createdByUser),
+        createdAt = OffsetDateTime.now()
+      )
     )
   }
 
@@ -77,6 +76,6 @@ class AppointmentService(
     appointment.durationInMinutes = durationInMinutes
     appointment.appointmentTime = appointmentTime
     appointment.deliusAppointmentId = deliusAppointmentId
-    return appointment
+    return appointmentRepository.save(appointment)
   }
 }
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/AppointmentService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/AppointmentService.kt
@@ -2,8 +2,10 @@ package uk.gov.justice.digital.hmpps.hmppsinterventionsservice.service
 
 import org.springframework.stereotype.Service
 import uk.gov.justice.digital.hmpps.hmppsinterventionsservice.jpa.entity.Appointment
+import uk.gov.justice.digital.hmpps.hmppsinterventionsservice.jpa.entity.AppointmentType
 import uk.gov.justice.digital.hmpps.hmppsinterventionsservice.jpa.entity.Attended
 import uk.gov.justice.digital.hmpps.hmppsinterventionsservice.jpa.entity.AuthUser
+import uk.gov.justice.digital.hmpps.hmppsinterventionsservice.jpa.entity.Referral
 import uk.gov.justice.digital.hmpps.hmppsinterventionsservice.jpa.repository.AppointmentRepository
 import uk.gov.justice.digital.hmpps.hmppsinterventionsservice.jpa.repository.AuthUserRepository
 import java.time.OffsetDateTime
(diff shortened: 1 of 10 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/AppointmentService.kt
```diff
@@ -14,54 +16,67 @@ import javax.transaction.Transactional
...
-    val appointment = Appointment(
+    return Appointment(
       id = UUID.randomUUID(),
       appointmentTime = appointmentTime,
       durationInMinutes = durationInMinutes,
+      deliusAppointmentId = deliusAppointmentId,
       createdBy = authUserRepository.save(createdByUser),
       createdAt = OffsetDateTime.now()
     )
-    appointmentRepository.save(appointment)
```
sldblog, line 54: Out of curiosity, what was the intent of moving `save` outside?

The function's verb "create"/"update" kind of implies it'd be "created", which I'd assume implies persistence

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/CommunityAPIBookingService.kt
```diff
@@ -70,8 +74,8 @@ class CommunityAPIBookingService(
       appointmentStart = appointmentTime,
       appointmentEnd = appointmentTime.plusMinutes(durationInMinutes.toLong()),
       officeLocationCode = officeLocation,
-      notes = getNotes(referral, resourceUrl, "Appointment"),
-      countsTowardsRarDays = true, // Fixme: For assessment booking this should be false and will pass in when assessment booking is done
+      notes = getNotes(referral, resourceUrl, "${get(appointmentType, notesFieldQualifier)} Appointment"),
+      countsTowardsRarDays = get(appointmentType, countsTowardsRarDays),
```
sldblog, line 78: I was under the impression that `countsTowardsRarDays` was already mapped in the structures of the Delius reference data; do we need it here?

The concept of "counts towards RAR days" is something our domain shouldn't need to understand or configure. It's the difference between "delivery" and "non-delivery" appointments that matters in the domain, and we can pass that along, right?

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/AppointmentServiceTest.kt
```diff
@@ -7,87 +7,117 @@ import com.nhaarman.mockitokotlin2.times
...
+      appointmentTime = appointmentTime,
+      durationInMinutes = durationInMinutes,
+      deliusAppointmentId = deliusAppointmentId,
+    )
+    whenever(appointmentRepository.save(any())).thenReturn(savedAppointment)
+
+    // When
+    val newAppointment = appointmentService.createOrUpdateAppointment(referral, null, durationInMinutes, appointmentTime, SUPPLIER_ASSESSMENT, createdByUser)
+
+    // Then
```
sldblog, line 67: Would it be good to have a conversation to pick a consistent style throughout the project?

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/AppointmentService.kt
```diff
@@ -14,54 +16,67 @@ import javax.transaction.Transactional
...
+      val deliusAppointmentId = communityAPIBookingService.book(referral, appointment, appointmentTime, durationInMinutes, appointmentType)
+      val updatedAppointment = updateAppointment(appointment, durationInMinutes, appointmentTime, deliusAppointmentId)
+      return appointmentRepository.save(updatedAppointment)
     }
 
-    val appointmentWithNonAttendance = appointment.attended == Attended.NO
-    if (appointmentWithNonAttendance) {
-      return createAppointment(durationInMinutes, appointmentTime, createdByUser)
+    val additionalAppointmentRequired = appointment.attended == Attended.NO
+    if (additionalAppointmentRequired) {
```
sldblog, line 46: _observation_ if all `createOrUpdateAppointment` calls would use `createAppointment` we essentially wouldn't need 3 branches, just

```kt
val deliusAppointmentId = communityAPIBookingService.book(referral, appointment, appointmentTime, durationInMinutes, appointmentType)
val appointment = createAppointment(durationInMinutes, appointmentTime, deliusAppointmentId, createdByUser)
return appointmentRepository.save(appointment)
```

the dangling appointments are negligible; it would however give us the history of the potentially different users who create/move them around before attendance

sldblog (commented):
looking good, just a couple of questions

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/AppointmentService.kt
```diff
@@ -14,54 +16,67 @@ import javax.transaction.Transactional
...
+      val deliusAppointmentId = communityAPIBookingService.book(referral, null, appointmentTime, durationInMinutes, appointmentType)
+      val additionalAppointment = createAppointment(durationInMinutes, appointmentTime, deliusAppointmentId, createdByUser)
+      return appointmentRepository.save(additionalAppointment)
     }
+
     throw IllegalStateException("Is it not possible to update an appointment that has already been attended")
```
sldblog, line 52: Could we make this check at the very beginning as a pre-condition? It feels important to be able to clearly express, however, currently, someone has to follow through all branches and build the negation of all possible conditions

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/AppointmentService.kt
mahalrmoj: I think that was the intent of this structure

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/AppointmentService.kt
mahalrmoj: Save was moved outside as the appointment needs to have the delius appointment id added to it before being saved

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/CommunityAPIBookingService.kt
mahalrmoj: I think you're probably right that community-api can determine this based on being told whether we are booking for a supplier assessment or a service delivery but this was unclear at the start. I can raise a tech debt ticket to cover this.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/AppointmentService.kt
mahalrmoj: It can only arise if we are 1) not making a fresh booking 2) not doing a reschedule for an unattended appointment and 3) not creating an additional appointment for a missed appointment, i.e. should never arise. 
Expressing 1-3 as a pre condition (rather than a fall through "all other situations") I think would be overly complicated to reason

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/AppointmentService.kt
tomsmyers: is this still true? looks like we pass the delius id into this method. would be good to put the save back in here if possible

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/CommunityAPIBookingService.kt
```diff
@@ -17,24 +19,26 @@ import javax.validation.constraints.NotNull
 class CommunityAPIBookingService(
   @Value("\${community-api.appointments.bookings.enabled}") private val bookingsEnabled: Boolean,
   @Value("\${interventions-ui.baseurl}") private val interventionsUIBaseURL: String,
-  @Value("\${interventions-ui.probation-links.view-appointment}") private val interventionsUIViewAppointment: String,
+  @Value("#{\${interventions-ui.probation-appointment-links}}") private val interventionsUIViewAppointmentLinks: Map<AppointmentType, String>,
```
tomsmyers, line 22: what does the pound do ? `#` vs `$`?

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/CommunityAPIBookingService.kt
```diff
@@ -95,6 +99,10 @@ class CommunityAPIBookingService(
 
   fun isDifferentTimings(existingAppointment: Appointment, appointmentTime: OffsetDateTime, durationInMinutes: Int): Boolean =
     !existingAppointment.appointmentTime.isEqual(appointmentTime) || existingAppointment.durationInMinutes != durationInMinutes
+
+  private inline fun <reified T> get(appointmentType: AppointmentType, map: Map<AppointmentType, T>): T {
```
tomsmyers, line 103: nit: don't think `T` needs to be reified?

tomsmyers (approved):
nice work! will be great to get action plan appointments using this.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/AppointmentService.kt
mahalrmoj: Will move the save - you are correct

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/CommunityAPIBookingService.kt
mahalrmoj: Necessary for a map configuration

PR #367, opened 2021-06-22
IC-1971: show referrals both managed and started by the PP only once

sldblog:

Removes duplication for managed/started referrals in the PP dashboard.

## How to review?

Commit-by-commit, please -- the first two commits are reorg.

The crux is
- 35ea245: adds the failing test
- 4fb6bd9: adds the fix

**Expected behaviour**
A single referral shows up once in the probation practitioner's dashboard.

**Actual behaviour**
If a PP user creates a referral for a service user who they manage, their dashboard shows the referral twice.

**What happens if we don't fix this?**
Users would like to make sure they did not duplicate. Quote from user:

> [...] tried to cancel one thinking there were two and I may have cancelled both?


tomsmyers (approved):
nice - thank you!

sldblog:
thank you! 🙇‍♂️

PR #372, opened 2021-06-23
IC-1980: Update delivery details when updating supplier assessment appointments

andrewburrowsmoj:

As with action plan appointments, we also need to ensure that supplier appointments also contain delivery details.

Ensure it's impossible to not have any appointments without a delivery.


andrewburrowsmoj:
Note to reviewer: Will need to update pact tests if this Ryan's pact tests are merged before mine

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/AppointmentService.kt
```diff
@@ -25,57 +31,100 @@ class AppointmentService(
...
     appointmentType: AppointmentType,
-    createdByUser: AuthUser
+    createdByUser: AuthUser,
+    appointmentDeliveryType: AppointmentDeliveryType,
+    appointmentDeliveryAddress: AddressDTO? = null,
   ): Appointment {
 
     val initialAppointmentRequired = appointment == null
     if (initialAppointmentRequired) {
       val deliusAppointmentId = communityAPIBookingService.book(referral, null, appointmentTime, durationInMinutes, appointmentType)
```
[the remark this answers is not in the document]
sldblog: there's a default we need to use, but I don't think that's our responsibility to know

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/AppointmentService.kt
```diff
@@ -25,57 +31,100 @@ class AppointmentService(
...
-      return createAppointment(durationInMinutes, appointmentTime, deliusAppointmentId, createdByUser)
+      return createAppointment(durationInMinutes, appointmentTime, deliusAppointmentId, createdByUser, appointmentDeliveryType, appointmentDeliveryAddress)
     }
 
     val updateCurrentAppointmentRequired = appointment!!.attended == null
     if (updateCurrentAppointmentRequired) {
```
sldblog, line 46: _this does not block this PR, can be discussed after it's merged_

Can we remove all branching here, by creating a new appointment every time `createOrUpdateAppointment` is called?

Ie. instead of figuring out create-or-update, we can always create an appointment and make the previous one superseded? It would significantly reduce the complexity here

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/AppointmentService.kt
sldblog, line 41: Double-checking the order of things

1. `communityAPIBookingService.book` with time and context details
2. create appointment and `appointmentRepository.saveAndFlush(appointment)`
3. create appointment delivery and `appointmentRepository.saveAndFlush(appointment)`
4. set the address if applicable and `appointmentDeliveryRepository.saveAndFlush(appointmentDelivery)`

is this ok?

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/controller/SupplierAssessmentControllerTest.kt
```diff
@@ -42,10 +46,10 @@ class SupplierAssessmentControllerTest {
     whenever(userMapper.fromToken(token)).thenReturn(user)
     whenever(referralService.getSentReferralForUser(referral.id, user)).thenReturn(referral)
     whenever(supplierAssessmentService.getSupplierAssessmentById(any())).thenReturn(supplierAssessment)
-    whenever(supplierAssessmentService.createOrUpdateSupplierAssessmentAppointment(supplierAssessment, durationInMinutes, appointmentTime, user)).thenReturn(supplierAssessment.currentAppointment)
+    whenever(supplierAssessmentService.createOrUpdateSupplierAssessmentAppointment(supplierAssessment, durationInMinutes, appointmentTime, user, appointmentDeliveryType)).thenReturn(supplierAssessment.currentAppointment)
 
     val response = supplierAssessmentController.updateSupplierAssessmentAppointment(referral.id, update, token)
-
+    verify(appointmentValidator, times(1)).validateUpdateAppointment(any())
```
sldblog, line 52: _nit_ Is it significant how many times does this happen?

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/integration/SetupAssistant.kt
```diff
@@ -215,17 +215,34 @@ class SetupAssistant(
...
-    appointment: Appointment = appointmentFactory.create(createdBy = createPPUser())
+    appointment: Appointment = appointmentFactory.create(createdBy = createPPUser()),
+    appointmentDeliveryType: AppointmentDeliveryType? = null,
+    appointmentDeliveryAddress: AddressDTO? = null
   ) {
     appointmentRepository.save(appointment)
     supplierAssessment.appointments.add(appointment)
-    supplierAssessmentRepository.save(supplierAssessment)
+    supplierAssessmentRepository.saveAndFlush(supplierAssessment)
+    if (appointmentDeliveryType != null) {
```
sldblog, line 231: _nit_ `appointmentDeliveryType` is never null in the test usage, so this branch could be skipped

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/controller/SupplierAssessmentController.kt
```diff
@@ -24,13 +26,15 @@ class SupplierAssessmentController(
...
+    appointmentValidator.validateUpdateAppointment(updateAppointmentDTO)
     return AppointmentDTO.from(
       supplierAssessmentService.createOrUpdateSupplierAssessmentAppointment(
         supplierAssessment,
         updateAppointmentDTO.durationInMinutes,
         updateAppointmentDTO.appointmentTime,
-        user
+        user,
+        updateAppointmentDTO.appointmentDeliveryType,
+        updateAppointmentDTO.appointmentDeliveryAddress
```
sldblog, line 37: I couldn't find tests around passing this address on from the controller to SupplierAssessmentService to AppointmentService -- is this tested in an integration test?

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/SupplierAssessmentService.kt
```diff
@@ -46,6 +50,8 @@ class SupplierAssessmentService(
       appointmentTime,
       SUPPLIER_ASSESSMENT,
       createdByUser,
+      appointmentDeliveryType,
+      appointmentDeliveryAddress
     )
     supplierAssessment.appointments.add(appointment)
```
sldblog, line 56: _this does not block this PR, can be discussed after it's merged_

I'm a bit confused with the chain of `createOrUpdateAppointment` then `appointments.add` -- is it "create" and "add"?

I can see the appointments are a set, but when thinking about appointments, order seems important, so I'm not sure set is a fitting data structure. What do you think?

sldblog (approved):
👏 couple of potentially impactful questions, but not blocking

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/controller/SupplierAssessmentControllerTest.kt
andrewburrowsmoj: It's not significant, but need verification that it happened at least once.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/AppointmentService.kt
andrewburrowsmoj: yeah, I think this part of the code can be improved.
(Although this code was not part of changes I made)
I'll add a TODO here

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/AppointmentService.kt
andrewburrowsmoj: Yes, that is the correct order.

andrewburrowsmoj: Yes that is the order in which it occurs.

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/controller/SupplierAssessmentControllerTest.kt
andrewburrowsmoj: It's not significant, I'll remove the times()

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/integration/SetupAssistant.kt
andrewburrowsmoj: I've removed this

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/controller/SupplierAssessmentController.kt
andrewburrowsmoj: I've now added address into the SupplierAssessmentControllerTest

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/SupplierAssessmentService.kt
andrewburrowsmoj: 1. This logic is not part of my PR.. but I agree that it seems a bit wrong to me... If you are updating an appointment you don't really want to add it again to the supplierAssessment. I'm assuming that if it's a duplicate appointment then it will be discarded (rather than throw some type of hibernate exception)..
I'll see if I can add a hibernate test for this.


2. Yes, you make a good point, I think appointments would be ideally ordered based on some time value. Maybe something to raise in next dev meeting.

andrewburrowsmoj: Just did a test -- it's fine because the Appointment class has a equals() on appointment.id

PR #375, opened 2021-06-24
create supplier assessments for all sent referrals which do not have them

tomsmyers:

create supplier assessments for all sent referrals which do not have them

there are broken referrals in all environments which were sent before supplier assessments are created on referrals send.

this migration creates the assessments for all these referrals.

### DIFF
--- a/src/main/resources/db/migration/V1_66__create_missing_supplier_assessments.sql
+++ b/src/main/resources/db/migration/V1_66__create_missing_supplier_assessments.sql
@@ -1,0 +1,7 @@
+create extension if not exists pgcrypto;
+
+insert into supplier_assessment (id, referral_id)
+    select gen_random_uuid(), r.id
+    from referral r
+        left join supplier_assessment sa on sa.referral_id = r.id
+        where sa.referral_id is null and r.sent_at is not null;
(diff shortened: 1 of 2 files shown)


src/main/resources/db/migration/V1_66__create_missing_supplier_assessments.sql
```diff
@@ -0,0 +1,7 @@
+create extension pgcrypto;
+
+insert into supplier_assessment (id, referral_id)
+    select gen_random_uuid(), r.id
```
sldblog, line 4: [uuid-ossp](https://www.postgresql.org/docs/10/uuid-ossp.html) has smaller scope than the [pgcrypto](https://www.postgresql.org/docs/10/pgcrypto.html) extension, so might be better:

```suggestion
create extension if not exists "uuid-ossp";

insert into supplier_assessment (id, referral_id)
    select uuid_generate_v4(), r.id
```

sldblog: ignore me, the docs advise to use pgcrypto :)

> If you only need randomly-generated (version 4) UUIDs, consider using the gen_random_uuid() function from the pgcrypto module instead.

src/main/resources/db/migration/V1_66__create_missing_supplier_assessments.sql
```diff
@@ -0,0 +1,7 @@
+create extension pgcrypto;
```
sldblog, line 1:
```suggestion
create extension if not exists pgcrypto;
```

tomsmyers: good point thanks.

andrewburrowsmoj (approved):
lgtm (after testing in my local db)

PR #377, opened 2021-06-24
IC-1829 use user token for calling ARN API

tomsmyers:

This PR is split into three commits. they do the following:

1. refactor all the existing WebClients to use the new RestClient helper class, which was added recently. its intention is to standardize things like request headers (and as you'll see in a later commit, authentication too).
2. change the spring security webclient config to no longer use a default oauth2 client registration id across all web clients, but instead allow each webclient the ability to specify which registration it wants to use. this is then applied on a per request basis in the RestClient helper class. what this allows us to do is decide on a per request basis if we want to override this client registration with a custom auth token.
3. override calls to the ARN API with the user's auth token instead of the default 'interventions-client' client_credentials token.

ARN requires user type tokens for security reasons.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/config/WebClientConfiguration.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/config/WebClientConfiguration.kt
@@ -33,19 +33,13 @@ class WebClientConfiguration(
   }
 
   @Bean
-  @Deprecated("usage of newer 'RestClient' interface is preferred")
-  fun communityApiWebClient(authorizedClientManager: OAuth2AuthorizedClientManager): WebClient {
-    return createAuthorizedWebClient(authorizedClientManager, communityApiBaseUrl)
+  fun hmppsAuthApiClient(authorizedClientManager: OAuth2AuthorizedClientManager): RestClient {
+    return RestClient(createAuthorizedWebClient(authorizedClientManager, hmppsAuthBaseUrl))
   }
 
   @Bean
-  fun hmppsAuthApiWebClient(authorizedClientManager: OAuth2AuthorizedClientManager): WebClient {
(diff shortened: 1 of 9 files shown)


sldblog:
two test failures on `expectStatus().is2xxSuccessful` without `expectBody()`... further evidence to #376

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/config/WebClientConfiguration.kt
```diff
@@ -60,7 +60,6 @@ class WebClientConfiguration(
 
   private fun createAuthorizedWebClient(clientManager: OAuth2AuthorizedClientManager, baseUrl: String): WebClient {
     val oauth2Client = ServletOAuth2AuthorizedClientExchangeFilterFunction(clientManager)
-    oauth2Client.setDefaultClientRegistrationId("interventions-client")
```
tomsmyers, line 69: this is really the crux of this commit

sldblog: _nit_ probably worth extracting `interventions-client` to a constant

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/component/CommunityAPIClientTest.kt
```diff
@@ -66,7 +66,7 @@ class CommunityAPIClientTest {
   fun `makes async post request successfully`() {
 
     communityAPIClient = CommunityAPIClient(
-      WebClient.builder().exchangeFunction(exchangeFunction).build()
+      RestClient(WebClient.builder().exchangeFunction(exchangeFunction).build(), "")
```
sldblog, line 69: _nit_ `"irrelevant"` might be a good hint for test readers

sldblog (approved):
👍 LGTM, 👏

PR #385, opened 2021-06-28
IC-1745 allow a referral to have more than one action plan 

tomsmyers:

store a list of action plans on a referral. for now just return the last one in the list (most recently added).

revert a previous commit which explicitly stopped you adding more than one action plan to a referral.

improve the test data so you can create action plans for all sent referrals.

do not allow setting fields on action plans on creation - to do this force clients to use the 'update action plan' endpoint.

lay the groundwork for editing already submitted action plans and creating new action plans after approval.


src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/ReferralConcluderTest.kt
```diff
@@ -49,7 +49,7 @@ internal class ReferralConcluderTest {
   fun `concludes referral as cancelled when ending a referral with no sessions attempted`() {
     val timeAtStart = OffsetDateTime.now()
     val actionPlan = actionPlanFactory.create(numberOfSessions = 2)
-    val referralWithActionPlanAndNoAttemptedSessions = referralFactory.createSent(actionPlan = actionPlan)
+    val referralWithActionPlanAndNoAttemptedSessions = referralFactory.createSent(actionPlans = mutableListOf(actionPlan))
```
sldblog, line 52: _nit_ factory could support both singular and list use cases, if that reduces cognitive load on reading the test; what do you think?

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/jpa/entity/Referral.kt
```diff
@@ -71,7 +72,7 @@ class Referral(
 
   var relevantSentenceId: Long? = null,
 
-  @OneToOne(mappedBy = "referral", optional = true) @Fetch(JOIN) var actionPlan: ActionPlan? = null,
+  @OneToMany(fetch = FetchType.LAZY) @JoinColumn(name = "referral_id") var actionPlans: MutableList<ActionPlan>? = null,
```
sldblog, line 75: we have a `idx_action_plan_referral_id` "must be unique" constraint on the DB -- can we remove that too, here?

sldblog (commented):
Seems very close to making this work with lists :)

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/ReferralConcluderTest.kt
tomsmyers: i kinda figured that if referrals have lists of action plans, it might make it _harder_ to reason about tests that set individual action plans

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/jpa/entity/Referral.kt
tomsmyers: ah missed this - thanks

sldblog (approved):
Looks good to me, thanks 👍 

Things that came up in my mind but I assume will be handled in the follow-up:

- make sure lookup by referral ID is efficient as it's the primary lookup
- tests that ordering of multiple action plans is well-defined in the entities (latest last, etc.)
- tests that non-active action plans are 410 Gone on reads or 4?? on writes

sldblog:
> non-active action plans are 410 Gone on reads

Well I guess it depends. It still exists. It might be a client-side `superseded` concern instead

tomsmyers:
> > non-active action plans are 410 Gone on reads
> 
> Well I guess it depends. It still exists. It might be a client-side `superseded` concern instead

@sldblog it also depends if the old action plan was approved or not. it's something i have thought about but it warrants it's own set of changes.

src/main/resources/db/migration/V1_67__remove_unique_index_on_action_plan_referral_id.sql
```diff
@@ -0,0 +1 @@
+drop index idx_action_plan_referral_id;
```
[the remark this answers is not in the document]
tomsmyers: done

sldblog (approved):
🎉 Yay

PR #387, opened 2021-06-28
tidy up 'interventions-ui.locations' configuration

tomsmyers:

removes duplicate links, removed unused variables and code, tidies up location naming.

since this is quite a serious change, i've been really deliberate about each and every location change. they are all split out in seperate commits, with information about the recipient of the link (whether that be delius, or an email recipient), the change in naming and the change in link location (if any).

some of the links were not correct, some had confusing names. this change aims to clarify things as much as possible.


mahalrmoj (approved):
Great clean up and rationalisation of naming. I would like to verify the delius links in dev before we promote to preprod.

PR #390, opened 2021-06-29
Re-add content assertions to flaky tests

sldblog:

Re-add content assertions to flaky tests

Related to
- #376 
- #378

Honestly, this is grasping at straws. There's insufficient data about what causes these `.expectStatus().is2xxSuccessful` calls to sometimes (quite often) fail with

```
> POST http://localhost:39473/sent-referral/24121550-20ff-421b-b39f-72a1abca1c77/assign

{"status":500,"error":"unexpected error","message":"Could not commit JPA transaction; ⏎
  nested exception is javax.persistence.RollbackException: Error while committing the transaction"}
```

Adding these might just add delay, _hiding_ the issue behind artificial lag.

However, it's getting very painful on `main` being blocked and needing several retries, so... let's try?


sldblog:
Never mind: it clearly doesn't work

image

PR #396, opened 2021-07-01
IC-2010: Store and send Delius Office Location Code to CommunityAPI

andrewburrowsmoj:

Store and send Delius Office Location Code to CommunityAPI when booking or rescheduling an appointment

Allows to users to specify a Delius office location for an appointment booking

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/controller/ActionPlanSessionController.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/controller/ActionPlanSessionController.kt
@@ -40,7 +40,8 @@ class ActionPlanSessionController(
       updateAppointmentDTO.durationInMinutes,
       user,
       updateAppointmentDTO.appointmentDeliveryType,
-      updateAppointmentDTO.appointmentDeliveryAddress
+      updateAppointmentDTO.appointmentDeliveryAddress,
+      updateAppointmentDTO.npsOfficeCode,
     )
     return ActionPlanSessionDTO.from(actionPlanSession)
   }
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/dto/PCCRegionDTO.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/dto/PCCRegionDTO.kt
@@ -6,12 +6,14 @@ import uk.gov.justice.digital.hmpps.hmppsinterventionsservice.jpa.entity.PCCRegi
 class PCCRegionDTO(
   val id: PCCRegionID,
   val name: String,
+  val npsRegion: NPSRegionDTO,
 ) {
   companion object {
     fun from(pccRegion: PCCRegion): PCCRegionDTO {
       return PCCRegionDTO(
         id = pccRegion.id,
-        name = pccRegion.name
+        name = pccRegion.name,
+        npsRegion = NPSRegionDTO.from(pccRegion.npsRegion)
       )
     }
   }
--- a/src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/AppointmentServiceTest.kt
+++ b/src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/AppointmentServiceTest.kt
@@ -56,6 +56,32 @@ class AppointmentServiceTest {
   }
 
   @Test
+  fun `can create new appointment with delius office location delivery`() {
+    // Given
+    val durationInMinutes = 60
+    val appointmentTime = OffsetDateTime.now()
+    val referral = referralFactory.createSent()
+    val deliusAppointmentId = 99L
+    val npsOfficeCode = "CRSEXT"
+
+    whenever(communityAPIBookingService.book(referral, null, appointmentTime, durationInMinutes, SUPPLIER_ASSESSMENT))
+      .thenReturn(deliusAppointmentId)
+    val savedAppointment = appointmentFactory.create(
+      appointmentTime = appointmentTime,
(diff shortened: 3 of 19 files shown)


src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/validator/AppointmentValidatorTest.kt
```diff
@@ -79,7 +89,7 @@ internal class AppointmentValidatorTest {
       }
 
       @Test
-      fun `empty address for non nps office appointment throws validation error`() {
+      fun `empty address for a non nps delius appointment throws validation error`() {
```
sldblog, line 92: In the user/domain language we still call this NPS location; can we keep the previous way?

src/main/resources/db/migration/V1_68__rename_nps_office_code_column_on_appointment_delivery.sql
```diff
@@ -0,0 +1 @@
+alter table appointment_delivery rename column nps_office_code to delius_office_location_code;
```
sldblog, line 1: Out of
- `nps_office_code`
- `delius_office_location_code`

the most semantically correct probably would be a third `probation_office_id` -- and using the reference data to make any dereferencing in the translation layer

Context: if we model our domain based on real-life, we're selecting an _office_, not a location code; our context doesn't require location IDs, that becomes relevant when we transition over the boundary to Delius' language

andrewburrowsmoj: ah so `probation_office_id` is the id on the first column in this csv? https://github.com/ministryofjustice/hmpps-reference-data/blob/main/probation-offices-v0.csv

I personally think it would be better with still using `delius_crs_location_id` for the following reasons:

1. I don't think it's safe to rely on an external data source id since we can't control if this changes. If we had the data out database this would be different since we can add constraints.
2. Similar to 1, I feel it's useful to have a record of the actual code we sent to delius. Yes, this is in reference-data but the connection is weaker and no protected from changes.
3. It means we don't have to implement any reference data integration on services side reducing the complexity for booking an appointment

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/AppointmentServiceTest.kt
```diff
@@ -49,6 +49,32 @@ class AppointmentServiceTest {
...
+      durationInMinutes = durationInMinutes,
+      deliusAppointmentId = deliusAppointmentId,
+    )
+    whenever(appointmentRepository.save(any())).thenReturn(savedAppointment)
+
+    // When
+    val newAppointment = appointmentService.createOrUpdateAppointment(referral, null, durationInMinutes, appointmentTime, SUPPLIER_ASSESSMENT, createdByUser, AppointmentDeliveryType.IN_PERSON_MEETING_PROBATION_OFFICE, deliusOfficeLocationCode = deliusOfficeLocationCode)
+
+    // Then
+    verifyResponse(newAppointment, null, true, deliusAppointmentId, appointmentTime, durationInMinutes, AppointmentDeliveryType.PHONE_CALL, deliusOfficeLocationCode)
```
tomsmyers: why is this still returning PHONE_CALL type?

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/controller/ActionPlanSessionController.kt
```diff
@@ -40,7 +40,8 @@ class ActionPlanSessionController(
       updateAppointmentDTO.durationInMinutes,
       user,
       updateAppointmentDTO.appointmentDeliveryType,
-      updateAppointmentDTO.appointmentDeliveryAddress
+      updateAppointmentDTO.appointmentDeliveryAddress,
+      updateAppointmentDTO.deliusOfficeLocationCode,
```
tomsmyers, line 44: i saw some discussion about the best way to refer to this data - is this what we settled on vs nps_office_code or whatever else was discussed?

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/dto/PCCRegionDTO.kt
```diff
@@ -6,12 +6,14 @@ import uk.gov.justice.digital.hmpps.hmppsinterventionsservice.jpa.entity.PCCRegi
 class PCCRegionDTO(
   val id: PCCRegionID,
   val name: String,
+  val npsRegion: NPSRegionDTO,
```
tomsmyers, line 9: what's the reason for this change?

tomsmyers (approved):
looks good andrew - thanks

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/dto/PCCRegionDTO.kt
andrewburrowsmoj: It's in the commit message:

```
IC-2010: Add nps region to pcc region. This is to ensure that the ui is able to understand which nps regions belong to a given pcc region when filtering for office locations.
```

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/controller/ActionPlanSessionController.kt
andrewburrowsmoj: Sorry it's been a while since I've revisted this PR. I've changed back to the usage of npsOfficeCode, I was originally thinking we'd want to make it more generic.

PR #397, opened 2021-07-01
IC-2013: Hide cancelled referrals from providers

sldblog:

Hide cancelled referrals from providers

Service providers should not see referrals made to them that were
withdrawn (cancelled) by the probation staff

This is to prevent unnecessary visibility of offender details

🙋‍♂️ Note: it's undefined what is the cut-off point here; we might still want
to keep this if there was a supplier assessment scheduled, but I'm
sticking to the "cancelled" definition here

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/jpa/entity/Referral.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/jpa/entity/Referral.kt
@@ -99,6 +99,8 @@ class Referral(
     return createdBy
   }
 
+  fun cancelled(): Boolean = concludedAt != null && endRequestedAt != null && endOfServiceReport == null
+
   val currentActionPlan: ActionPlan?
     get() = actionPlans?.lastOrNull()
 }
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/ReferralService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/ReferralService.kt
@@ -245,7 +245,7 @@ class ReferralService(
     }
 
     update.serviceUser?.let {
-      if (it.crn != null && it.crn != referral.serviceUserCRN) {
+      if (it.crn != referral.serviceUserCRN) {
         errors.add(FieldError(field = "serviceUser.crn", error = Code.FIELD_CANNOT_BE_CHANGED))
       }
     }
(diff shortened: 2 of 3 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/ReferralService.kt
```diff
@@ -244,7 +245,7 @@ class ReferralService(
     }
 
     update.serviceUser?.let {
-      if (it.crn != null && it.crn != referral.serviceUserCRN) {
+      if (it.crn != referral.serviceUserCRN) {
```
sldblog, line 248: fixed warning as `crn` is always non-null

tomsmyers: nit: it would be good to keep these types of changes in seperate commits

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/jpa/entity/Referral.kt
```diff
@@ -5,6 +5,8 @@ import org.hibernate.annotations.FetchMode
 import org.hibernate.annotations.FetchMode.JOIN
 import java.time.LocalDate
 import java.time.OffsetDateTime
+import java.util.Objects.isNull
```
tomsmyers, line 8: usages of these seem kinda non-kotlin-y

tomsmyers: IMO i think this reads clearer as just `concludedAt != null && endRequestedAt !=null && endOfServiceReport == null`

tomsmyers: this means referrals with EOSRs still stick around forever?

sldblog: for now, in this dashboard, yes.
I expect there'll be filtering options to the endpoint later on to support the "live" and "ended" dashboards

sldblog: moved to `null` equality checks in 514cb097

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/ReferralService.kt
sldblog: separated to d580d03af85ed5df900d47678fa8bd25fccd3d49

tomsmyers (approved):

PR #402, opened 2021-07-05
Update NDMIS report to no longer assume a single action plan per referral

tomsmyers:
To achieve this we add two new subqueries, one to select the first created action
plans, and one to select the most recently created action plans. We populate
'date_first_action_plan_submitted' from the formed query and other action plan
related values e.g. 'number of sessions' and the derived 'concluded_at' from
the latter.

here we have also added the code to populate the 'date_of_first_action_plan_approval' field (in e0f47b49093aa2fd18d96a5a6786f2d9b55d27bb)

fix NDMIS report which currently has duplicate referrals id rows.

### DIFF
--- a/helm_deploy/hmpps-interventions-service/reports/crs_performance_report.sql
+++ b/helm_deploy/hmpps-interventions-service/reports/crs_performance_report.sql
@@ -20,6 +20,11 @@ COPY (
     from action_plan
     group by referral_id
   ),
+  first_approved_action_plans AS (
+     select referral_id, min(approved_at) as approved_at
+     from action_plan
+     group by referral_id
+  ),
   current_action_plans AS (
       select id, referral_id, number_of_sessions from action_plan
       where (referral_id, created_at) in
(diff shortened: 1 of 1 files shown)


helm_deploy/hmpps-interventions-service/reports/crs_performance_report.sql
```diff
@@ -14,6 +14,25 @@ COPY (
...
+  first_submitted_action_plans AS (
+    select referral_id, min(submitted_at) as submitted_at
+    from action_plan
+    group by referral_id
+  ),
+  first_approved_action_plans AS (
+     select referral_id, min(approved_at) as approved_at
+     from action_plan
+     group by referral_id
+  ),
```
mahalrmoj, line 27: The two above could be combined into a single statement returning both columns from a performance viewpoint

helm_deploy/hmpps-interventions-service/reports/crs_performance_report.sql
```diff
@@ -14,6 +14,25 @@ COPY (
...
+  current_action_plans AS (
+      select id, referral_id, number_of_sessions from action_plan
+      where (referral_id, created_at) in
+        (
+          select referral_id, max(created_at)
+          from action_plan
+          group by referral_id
+        )
   )
```
mahalrmoj, line 36: Alternatively
```
current_action_plan as (
  SELECT r.id, ap.number_of_sessions,
  ROW_NUMBER() OVER(PARTITION BY r.id ORDER BY ap.approved_at DESC) as approved_at_desc_seq,
  FROM referral r
  INNER JOIN action_plan ap ON ap.referral_id = r.id
)
```
```
    LEFT JOIN current_action_plan cap ON (cap.referral_id = r.id and cap.approved_at_desc_seq = 1)
```

mahalrmoj (approved):
Just some alternatives but not essential

helm_deploy/hmpps-interventions-service/reports/crs_performance_report.sql
mahalrmoj: Some people prefer SQL Window Functions and others Group By

PR #403, opened 2021-07-06
IC-1116: Referal Performance report data

rgforsyth:

Creates an endpoint that will retrieve referral data between two given days.

To provide performance data.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/jpa/entity/ReferralReportData.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/jpa/entity/ReferralReportData.kt
@@ -1,0 +1,29 @@
+package uk.gov.justice.digital.hmpps.hmppsinterventionsservice.jpa.entity
+
+import java.time.Instant
+import java.util.UUID
+
+interface ReferralReportData {
+  val referralLink: String?
+  val referralReference: String
+  val referralId: UUID
+  val contractId: String
+  val organisationId: String
+  val referringOfficerEmail: String?
+  val caseworkerId: UUID?
+  val serviceUserCRN: String
+  val dateReferralReceived: Instant
+  val dateSAABooked: Instant?
+  val dateSAAAttended: Instant?
+  val dateFirstActionPlanSubmitted: Instant?
+  val dateOfFirstActionPlanApproval: Instant?
+  val dateOfFirstAttendedSession: Instant?
+  val outcomesToBeAchievedCount: Int?
+  val outcomesAchieved: Double?
+  val countOfSessionsExpected: Int?
+  val countOfSessionsAttended: Int?
+  val endRequestedByPPAt: Instant?
+  val endRequestedByPPReason: CancellationReason?
+  val dateEOSRSubmitted: Instant?
+  val concludedAt: Instant?
+}
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/jpa/repository/ReferralRepository.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/jpa/repository/ReferralRepository.kt
@@ -27,79 +27,79 @@ interface ReferralRepository : JpaRepository<Referral, UUID> {
 
   @Query(
     value =
-    """WITH attended_sessions AS (
-        select count(app.id) AS attended, min(app.appointment_time) as first_appointment, aps.action_plan_id, act.referral_id
-            from appointment app 
-                join action_plan_session_appointment apsa on app.id = apsa.appointment_id
-                join action_plan_session aps on apsa.action_plan_session_id = aps.id
-                join action_plan act on aps.action_plan_id = act.id
-            where attended in ('YES', 'LATE')
-            group by aps.action_plan_id, act.referral_id),
-        outcomes_to_be_achieved AS (
-            select count(des.id) as outcomes_to_be_achieved, ref.id AS referral_id
-                from referral ref
-                    inner join referral_selected_service_category rssc on rssc.referral_id = ref.id
-                    inner join desired_outcome des on des.service_category_id = rssc.service_category_id
-                group by ref.id),
-        achieved_outcomes AS (
-            select sub.partially_achieved + sub.achieved_count AS total, referral_id
-                from (select ref.id AS referral_id,
-                    sum(case when eosro.achievement_level = 'PARTIALLY_ACHIEVED' then 0.5 end) partially_achieved,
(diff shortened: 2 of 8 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/jpa/repository/ReferralRepository.kt
```diff
@@ -20,4 +24,73 @@ interface ReferralRepository : JpaRepository<Referral, UUID> {
...
+            "aco.total AS outcomesAchieved, " +
+            "act.number_of_sessions AS countOfSessionsExpected, " +
+            "ats.attended AS countOfSessionsAttended, " +
+            "cast(ref.end_requested_at as TIMESTAMP WITH TIME ZONE) AS endRequestedByPPAt, " +
+            "ref.end_requested_by_id AS endRequestedByPPReason, " +
+            "eosr.submitted_at AS dateEOSRSubmitted, " +
+            "cast(ref.concluded_at as TIMESTAMP WITH TIME ZONE) AS concludedAt " +
+                "from referral ref " +
+                    "left join attended_sessions ats on ats.referral_id = ref.id " +
+                    "left join action_plan act on act.referral_id = ref.id " +
```
tomsmyers, line 84: does this assume there is one action plan per referral?

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/jpa/repository/ReferralRepository.kt
```diff
@@ -20,4 +24,73 @@ interface ReferralRepository : JpaRepository<Referral, UUID> {
...
+            "select app.created_at as appointment_booked_time, " +
+            "case when (app.attended in ('LATE', 'YES')) then app.appointment_time else null end as attended_time, " +
+            "sup.referral_id " +
+                "from appointment app " +
+                    "join supplier_assessment_appointment saa on app.id = saa.appointment_id " +
+                    "join supplier_assessment sup on saa.supplier_assessment_id = sup.id ) " +
+        "select " +
+            "null AS referralLink, " +
+            "ref.reference_number AS referralReference, " +
+            "cast(ref.id as varchar) AS referralId, " +
```
tomsmyers, line 62: why do we have to do these casts?

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/jpa/repository/ReferralRepository.kt
```diff
@@ -20,4 +24,73 @@ interface ReferralRepository : JpaRepository<Referral, UUID> {
   // queries for draft referrals
   fun findByIdAndSentAtIsNull(id: UUID): Referral?
   fun findByCreatedByIdAndSentAtIsNull(userId: String): List<Referral>
+
+  @Query(value =
+    "WITH attended_sessions AS ( " +
```
tomsmyers, line 29: you can use triple quotes for multiline strings to avoid all the extra quotes and plus symbols.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/jpa/repository/ReferralRepository.kt
rgforsyth: When using a native query, UUID's could not be automatically mapped and were throwing errors, therefore had to be cast

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/jpa/entity/ReferralReportData.kt
```diff
@@ -0,0 +1,29 @@
+package uk.gov.justice.digital.hmpps.hmppsinterventionsservice.jpa.entity
+
+import java.time.Instant
+import java.util.UUID
+
+interface ReferralReportData {
```
tomsmyers, line 6: this isn't an entity (no @Entity annotation). it's also identical to the DTO. can we just have one class to represent this thing?

rgforsyth: As far as we could see, In order for the native query to work and map, you have to use an interface.

tomsmyers: i see. i also now notice there are some differences in the types. why do we have to use Instant here ? our time stamp sql values map directly to OffsetDateTime elsewhere, why not here?

regarding this not being an entity, we should just define this class where it's used, since having it in the entity directory suggests this class represents a model in the database.

rgforsyth: Again, this was to get the mapping's to work properly. The sql would only seem to return as an SQL Timestamp and would not map to offsetDateTime directly

tomsmyers: this is the reason JPA exists 😝

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/ReportingService.kt
```diff
@@ -0,0 +1,34 @@
...
+
+@Service
+@Transactional
+class ReportingService(
+  val referralRepository: ReferralRepository,
+  val serviceProviderUserAccessScopeMapper: ServiceProviderAccessScopeMapper,
+  @Value("\${interventions-ui.baseurl}") private val interventionsUIBaseURL: String,
+  @Value("\${interventions-ui.locations.referral-details}") private val interventionsUIReferralDetailsLocation: String,
+) {
+  fun getReportData(to: OffsetDateTime, from: OffsetDateTime, user: AuthUser): List<ReferralReportDataDTO> {
```
tomsmyers, line 22: nit: `from, to` is a more natural order for these params

nit: it's not clear what 'user' is for in the param name. this makes me wonders, is it the responsibility of the reporting service to look up the access scope? could this method signature be like "getReportDataForContracts" and leave that up to the caller? that way if, down the line, we want to adds methods to say, just get report data for cases assigned to a certain user, we could add a method like 'getReportDataForReferralAssignee' that takes a user param. 

we don't have to do it now, because it doesn't affect the API, but it's good to consider how this class might evolve, and how the method signatures look to uninformed future readers.

PR #407, opened 2021-07-07
IC-2058: sql and entity changes for nominated access of referrals

andrewburrowsmoj:

Defines the data structure for nominated access

To facilitate future development work for nominated access and provide a discussion before starting it.

### Notes

Nominated access has two parts:
1. Service provider sub contractor access rights to the referral.
2. Case assignment based on the access right rules.

The user must select the service providers (subcontractors) which should have access to this referral. It's assumed that the prime provider will always have access.
When access has been granted to one or more (subcontractors) then the user can select the case workers for the service provider categorised based on subcontractor (and prime-provider).
There will eventually be some rules based on assigning a referral and nominated access subcontractors.

Questions:
Are we going to be querying hmpps-auth for all users belonging to a specific role? 
I assume that we don't want to be recording details of user's roles in our database and should be using hmpps-auth for this purpose.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/jpa/entity/Referral.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/jpa/entity/Referral.kt
@@ -36,6 +36,15 @@ class SelectedDesiredOutcomesMapping(
 @Entity
 @Table(indexes = arrayOf(Index(columnList = "created_by_id")))
 class Referral(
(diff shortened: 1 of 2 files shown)


tomsmyers (commented):
this looks good andrew - however before merging let's just check in at stand up about the current state of the requirements on this. there has been some chatter about this all changing.

sldblog:
given this is 1.5 months old, is this still relevant?

tomsmyers:
@andrewburrowsmoj @sldblog closing this PR until we have more clarity on what is required.

PR #410, opened 2021-07-08
Bump gradle's max memory to 512MB (from 256MB)

sldblog:

Bump gradle's max memory to 512MB (from 256MB)
- on all jobs that call `./gradlew`

The previous value is defined in the hmpps orb and coming from the
`executor: hmpps/java` setting

This change is because we started seeing
```
> Task :compileTestKotlin
Exception: java.lang.OutOfMemoryError
```
on CI

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -33,6 +33,8 @@ jobs:
   validate:
     executor: hmpps/java
     docker: *db_docker_config
+    environment:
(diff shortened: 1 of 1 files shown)


tomsmyers (approved):
are the rest of the options the existing defaults?

sldblog:
> are the rest of the options the existing defaults?

Yes

PR #412, opened 2021-07-09
IC-2033 store history of referral assignments

tomsmyers:

stores a history of referral assignments as a list of the referral entity. 

allow reassignments without losing the data on who was previously assigned.


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/jpa/entity/Referral.kt
```diff
@@ -103,4 +100,10 @@ class Referral(
 
   val currentActionPlan: ActionPlan?
     get() = actionPlans?.lastOrNull()
+
+  val currentAssignment: ReferralAssignment?
+    get() = assignments.lastOrNull()
```
andrewburrowsmoj, line 105: Will `last` always return the latest? Perhaps might need to sort by assignedAt.

src/main/resources/db/migration/V1_70__referral_assignment_history.sql
```diff
@@ -0,0 +1,10 @@
+create table referral_assignments
```
andrewburrowsmoj, line 1: the table name here is plural whilst most other tables use singular

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/jpa/entity/Referral.kt
tomsmyers: good point - it _should_ since the list is always appended to, but can't hurt to make it explicity

src/main/resources/db/migration/V1_70__referral_assignment_history.sql
tomsmyers: hmm, i think the plural makes sense to the reader, because the table is used in a collection. but i'll have a look at the existing stuff and make it as consistent as possible.

andrewburrowsmoj: Yeah, sorry, I think it's best to keep it as plural. It's not a straight noun as other tables are, but rather a possessive noun, so 2nd order nouns would be beneficial to have plural/singular distinction.

tomsmyers: 👍

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/jpa/entity/Referral.kt
andrewburrowsmoj: yeah. I was thinking of more when it's returned by a hibernate fetch and how it orders it. Normally it would be based on the natural order of the database (using `ctid`) but not 100% sure it will always be in the order of insertion (maybe if there's an update for instance).

andrewburrowsmoj (approved):
lgtm

src/main/resources/db/migration/V1_70__referral_assignment_history.sql
```diff
@@ -0,0 +1,21 @@
...
+    constraint fk_assigned_by_id_auth_user foreign key (assigned_by_id) references auth_user,
+    constraint fk_assigned_to_id_auth_user foreign key (assigned_to_id) references auth_user
+);
+
+insert into referral_assignments
+select id, assigned_at, assigned_by_id, assigned_to_id
+from referral;
+
+alter table referral
+    drop column assigned_at;
```
sldblog, line 17: _nit_ rename would provide a simpler way to rollback, but I appreciate that's probably not necessary now

src/main/resources/db/migration/V1_70__referral_assignment_history.sql
```diff
@@ -0,0 +1,21 @@
+create table referral_assignments
+(
```
mahalrmoj, line 2: Because there is no pk - should we have an index so it avoids a table scan?

src/main/resources/db/migration/V1_70__referral_assignment_history.sql
```diff
@@ -0,0 +1,21 @@
...
+insert into referral_assignments
+select id, assigned_at, assigned_by_id, assigned_to_id
+from referral;
+
+alter table referral
+    drop column assigned_at;
+alter table referral
+    drop column assigned_by_id;
+alter table referral
+    drop column assigned_to_id;
```
mahalrmoj, line 21: Should you take a copy of the table before doing this migration so you have a rollback point?

mahalrmoj:
Nice solution

src/main/resources/db/migration/V1_70__referral_assignment_history.sql
tomsmyers: good idea. on referral_id ?

tomsmyers: done

src/main/resources/db/migration/V1_70__referral_assignment_history.sql
tomsmyers: as discussed on slack, going to keep the old columns around as suggested by david to achieve this.

src/main/resources/db/migration/V1_70__referral_assignment_history.sql
tomsmyers: done

PR #417, opened 2021-07-13
Add referral to appointment

rgforsyth:

Adds a reference to the referral id on an appointment.

We currently have two types of appointments (supplier assessment & action plan). Finding the referral for a given appointment currently means we have to search the database against both possible paths to an appointment, as we can't go from an appointment entity backwards to action plans/saa to referral.

Having the referral easily accessible from an appointment will allow future changes, where we don't have the referral readily available, but needed, such as the generic appointment/notify services easier and cleaner to develop.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/jpa/entity/Appointment.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/jpa/entity/Appointment.kt
@@ -47,7 +47,7 @@ data class Appointment(
   @PrimaryKeyJoinColumn
   var appointmentDelivery: AppointmentDelivery? = null,
 
-  @ManyToOne(fetch = FetchType.LAZY) var referral: Referral,
+  @ManyToOne(fetch = FetchType.LAZY) var referral: Referral? = null,
 
   @Id val id: UUID,
 ) {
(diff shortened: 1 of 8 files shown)


src/main/resources/db/migration/V1_71__appointment_references_referral.sql
```diff
@@ -0,0 +1,28 @@
...
+    from referral ref
+             inner join supplier_assessment sup on sup.referral_id = ref.id
+             inner join supplier_assessment_appointment saa on saa.supplier_assessment_id = sup.id
+             inner join appointment app on app.id = saa.appointment_id
+    where app.id = apo.id
+)
+where apo.referral_id is null;
+
+alter table appointment
+    alter column referral_id set not null
```
mahalrmoj, line 28: Will this cause problems during deployment? When the migration scripts are done and the new code yet to be deployed, any existing transactions will fail as they wont populate this field. Is this the case @tomsmyers?

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/jpa/entity/Appointment.kt
```diff
@@ -46,6 +47,8 @@ data class Appointment(
   @PrimaryKeyJoinColumn
   var appointmentDelivery: AppointmentDelivery? = null,
 
+  @ManyToOne(fetch = FetchType.LAZY) var referral: Referral,
```
mahalrmoj, line 50: Originally the choice was a referral id - I like the idea of the referral but what was the reason for the change

src/main/resources/db/migration/V1_71__appointment_references_referral.sql
tomsmyers: yeh i agree. i think we need to perform this step in a seperate deploy.

rgforsyth: i'll remove it

mahalrmoj: Not sure that completely solves it. Because of the race condition, not all "referral ids" will get populated and so some will be null. Not sure how to solve this one

tomsmyers (approved):

PR #424, opened 2021-07-15
add data migration & make referral on appointment not null

rgforsyth:

The second part of [pr-417](https://github.com/ministryofjustice/hmpps-interventions-service/pull/417). This adds a new migration that will update the existing appointments with a referral_id. Then finally make referral not null on the appointment.

To ensure that all appointments have a referral_id going forward.

### DIFF
--- a/src/main/resources/db/migration/V1_72__appointment_referral_data.sql
+++ b/src/main/resources/db/migration/V1_72__appointment_referral_data.sql
@@ -1,0 +1,22 @@
+Update appointment apo
+set referral_id = (
+    select ref.id
+    from referral ref
+             inner join action_plan act on ref.id = act.referral_id
+             inner join action_plan_session aps on aps.action_plan_id = act.id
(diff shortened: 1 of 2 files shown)


src/main/resources/db/migration/V1_72__appointment_referral_data.sql
```diff
@@ -0,0 +1,22 @@
+Update appointment apo
+set referral_id = (
```
sldblog, line 2: _not a blocker_
This kind of thing feels to be simpler through one-off management tasks, does spring have a capacity to run those?

I'm thinking of things like in Rails/Phoenix where we could `kubectl exec` into a pod and run `rake/mix task` as a one-off

mahalrmoj:
This can't be merged until the previous one has been deployed otherwise they may get deployed together which is what you are trying to repvent

rgforsyth:
The first pr has been deployed to prod

tomsmyers (approved):

PR #426, opened 2021-07-15
Change initial assessment appointment feedback endpoints to be referenced by referral

andrewburrowsmoj:

Changes the endpoints for supplier assessment appointment feedback to be callable by referral.id rather than appointment.id

Caller does not have to worry about the current supplier assessment appointment when submitting feedback

## Note:
Nothing current uses these endpoints. I will write pact tests in next PR


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/controller/SupplierAssessmentController.kt
```diff
@@ -38,4 +49,58 @@ class SupplierAssessmentController(
...
+    authentication: JwtAuthenticationToken
+  ): AppointmentDTO {
+    val user = userMapper.fromToken(authentication)
+    val supplierAssessmentAppointment = getSupplierAssessmentAppointment(referralId, user)
+    return AppointmentDTO.from(appointmentService.submitSessionFeedback(supplierAssessmentAppointment, user))
+  }
+
+  private fun getSupplierAssessmentAppointment(referralId: UUID, user: AuthUser): Appointment {
+    val referral = referralService.getSentReferralForUser(referralId, user)
+    if (referral == null) {
```
tomsmyers, line 95: tiny nit: you can avoid the `== null` check with a `?: ...`

tomsmyers (approved):
looks great - thank you!

PR #427, opened 2021-07-15
Audit: track authorised scope of service providers

sldblog:

🕵️ Track authorised scope of service providers (9e684ed9)
🕵️ (bonus) Track user names when we track user IDs (53836209)
📐 Predictability: ensures providers and contracts are ordered in the scope (8e7ee35e)

Currently, we can prove what information we have in our database and we have tests that give us confidence that the filtering behaviour is correct

However, provider users also take external input to the filtering, which is what providers and groups were active at the time on their auth profile

**Hypothesis**
Logging this data point allows us to completely reproduce the filter for the `/sent-referrals` query

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/authorization/ServiceProviderAccessScopeMapper.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/authorization/ServiceProviderAccessScopeMapper.kt
@@ -64,6 +64,7 @@ class ServiceProviderAccessScopeMapper(
         "InterventionsAuthorizedProvider",
         mapOf(
           "userId" to user.id,
+          "userName" to user.userName,
           "userAuthSource" to user.authSource,
           "contracts" to it.contracts.joinToString(",") { c -> c.contractReference },
           "providers" to it.serviceProviders.joinToString(",") { p -> p.id },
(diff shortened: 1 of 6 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/authorization/ServiceProviderAccessScopeMapper.kt
```diff
@@ -57,7 +59,19 @@ class ServiceProviderAccessScopeMapper(
...
-    )
+    ).also {
+      telemetryClient.trackEvent(
+        "InterventionsAuthorizedProvider",
+        mapOf(
+          "userId" to user.id,
+          "userName" to user.userName,
+          "userAuthSource" to user.authSource,
+          "contracts" to it.contracts.joinToString(",") { c -> c.contractReference },
+          "providers" to it.serviceProviders.joinToString(",") { p -> p.id },
```
andrewburrowsmoj, line 70: Maybe should order to ensure equality when contracts/providers are the same but difference order

sldblog: Good shout. Perhaps I should do that on the query level 🤔

sldblog: Done in 8e7ee35ee5bf7bde57929cb73456165390294273

andrewburrowsmoj (approved):

sldblog:
Thank you!

PR #430, opened 2021-07-16
IC-1116 email report results to requester on completion

tomsmyers:

IC-1116 email report results to requester on completion

let the requesting SP know when the report is ready. 

the UI for this is not build yet, but it will simply stream the report directly from s3 for the logged in user.

we can be smart about permissions in the UI download controller too. since the filename contains the requesting user's id, we can make rules like "only the user that requested the file can download it".

### DIFF
--- a/helm_deploy/values-dev.yaml
+++ b/helm_deploy/values-dev.yaml
@@ -24,3 +24,4 @@ env:
   COMMUNITYAPI_BASEURL: "https://community-api-secure.test.delius.probation.hmpps.dsd.io"
   ASSESSRISKSANDNEEDS_BASEURL: "https://assess-risks-and-needs-dev.hmpps.service.justice.gov.uk"
   SENTRY_ENVIRONMENT: "dev"
+  AWS_S3_ENABLED: "true"
--- a/script/localstack/buckets.sh
+++ b/script/localstack/buckets.sh
@@ -1,0 +1,4 @@
+#!/bin/bash
+set -x
+awslocal s3 mb s3://interventions-bucket-local
+set +x
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/config/S3Configuration.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/config/S3Configuration.kt
@@ -1,0 +1,34 @@
+package uk.gov.justice.digital.hmpps.hmppsinterventionsservice.config
+
+import org.springframework.beans.factory.annotation.Value
+import org.springframework.context.annotation.Bean
+import org.springframework.context.annotation.Configuration
+import software.amazon.awssdk.auth.credentials.AwsBasicCredentials
+import software.amazon.awssdk.regions.Region
+import software.amazon.awssdk.services.s3.S3Client
+import java.net.URI
+
+@Configuration
+class S3Configuration(
+  @Value("\${aws.s3.provider}") private val provider: String,
+  @Value("\${aws.s3.region}") private val region: String,
+  @Value("\${aws.s3.access-key-id}") private val accessKeyId: String,
+  @Value("\${aws.s3.secret-access-key}") private val secretAccessKey: String,
+  @Value("\${aws.s3.endpoint.uri:}") private val uri: String,
+) {
+  @Bean
+  fun s3Client(): S3Client {
+    val builder = S3Client.builder()
+      .region(Region.of(region))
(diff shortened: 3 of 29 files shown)


script/localstack/buckets.sh
```diff
@@ -0,0 +1,4 @@
+#!/bin/bash
+set -x
+awslocal s3 mb s3://interventions-bucket-local
```
andrewburrowsmoj, line 3: I think it would be worthwhile adding something to the readme about this script or perhaps a line on how to create a S3 bucket on localstack?

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/config/S3Configuration.kt
```diff
@@ -0,0 +1,34 @@
...
+import software.amazon.awssdk.services.s3.S3Client
+import java.net.URI
+
+@Configuration
+class S3Configuration(
+  @Value("\${aws.s3.provider}") private val provider: String,
+  @Value("\${aws.s3.region}") private val region: String,
+  @Value("\${aws.s3.access-key-id}") private val accessKeyId: String,
+  @Value("\${aws.s3.secret-access-key}") private val secretAccessKey: String,
+  @Value("\${aws.s3.endpoint.uri:}") private val uri: String,
```
andrewburrowsmoj, line 17: Is the ':' a typo?

helm_deploy/values-dev.yaml
```diff
@@ -24,3 +24,4 @@ env:
   COMMUNITYAPI_BASEURL: "https://community-api-secure.test.delius.probation.hmpps.dsd.io"
   ASSESSRISKSANDNEEDS_BASEURL: "https://assess-risks-and-needs-dev.hmpps.service.justice.gov.uk"
   SENTRY_ENVIRONMENT: "dev"
+  AWS_S3_ENABLED: "true"
```
andrewburrowsmoj, line 27: Where is this variable used?
I can only see that `aws.s3.enabled` is used but from the application.yaml....

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/reporting/serviceprovider/performance/PerformanceReportJobListener.kt
```diff
@@ -0,0 +1,60 @@
...
+
+  override fun afterJob(jobExecution: JobExecution) {
+    if (jobExecution.status == BatchStatus.COMPLETED) {
+      val path = Path(jobExecution.executionContext.getString("output.file.path"))
+
+      s3Service.publishFileToS3(path, "reports/service-provider/performance/")
+
+      emailSender.sendEmail(
+        notifyTemplateId,
+        jobExecution.jobParameters.getString("user.email"),
```
andrewburrowsmoj, line 45: Is this parameter supposed to be set in the ReportingService?

andrewburrowsmoj, line 45: Sorry... not 100% sure where this is being set? The ReportingService doesn't seem to set it.

script/localstack/buckets.sh
tomsmyers: good idea ! will do

helm_deploy/values-dev.yaml
tomsmyers: these env vars and the configuration file entries are interchangeable. 

it's the same with the env vars like `COMMUNITYAPI_BASEURL` which translates to `community-api.baseurl`

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/config/S3Configuration.kt
tomsmyers: good spot! no! it's used to provide a default value if the configuration property is missing. in this case, when running locally, the URI is set, when running in production, the URI configuration is missing (and unused). the `:` defaults it to `""`. the service would not run without it.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/reporting/serviceprovider/performance/PerformanceReportJobListener.kt
tomsmyers: good spot! this is a convincing reason to add parameter validation as suggested previously

PR #435, opened 2021-07-19
Add a native SQL path for obtaining Sent Referral Summary for Service…

andrewburrowsmoj:
Provides a quicker retrieval of SentReferral summary details to be showed on the dashboard for service providers.

Some users with a lot of referrals are experiencing timeouts. We believe this is caused by too many SQL queries being called for a single request. We're aiming to reduce this request to less than a second in order to stop timeout issues.

### Note

This is a temporarily solution to mitigate issues with production which calls upto a 1000 subquries when fetching SentReferrals for the dashboard.
We suspect that a complete solution will involve pagination.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/jpa/repository/ReferralRepository.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/jpa/repository/ReferralRepository.kt
@@ -17,6 +17,7 @@ interface ReferralRepository : JpaRepository<Referral, UUID> {
 			serviceUserFirstName,
 			serviceUserLastName from (	
 	select
+			cast(r.id as varchar) AS referralId,
 			cast(r.sent_at as TIMESTAMP WITH TIME ZONE) as sentAt,
 			r.reference_number as referenceNumber,
       dfc.id as dynamicFrameworkContractId,
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/jpa/repository/ReferralRepository.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/jpa/repository/ReferralRepository.kt
@@ -1,12 +1,49 @@
 package uk.gov.justice.digital.hmpps.hmppsinterventionsservice.jpa.repository
 
 import org.springframework.data.jpa.repository.JpaRepository
+import org.springframework.data.jpa.repository.Query
 import uk.gov.justice.digital.hmpps.hmppsinterventionsservice.jpa.entity.AuthUser
 import uk.gov.justice.digital.hmpps.hmppsinterventionsservice.jpa.entity.Referral
 import uk.gov.justice.digital.hmpps.hmppsinterventionsservice.jpa.entity.ServiceProvider
+import uk.gov.justice.digital.hmpps.hmppsinterventionsservice.jpa.entity.ServiceProviderSentReferralSummary
 import java.util.UUID
 
 interface ReferralRepository : JpaRepository<Referral, UUID> {
+  @Query(
+    value = """select sentAt,
+			referenceNumber,
+			assignedToUserName,
+			interventionTitle,
+			serviceUserFirstName,
+			serviceUserLastName from (	
+	select
+			cast(r.sent_at as TIMESTAMP WITH TIME ZONE) as sentAt,
+			r.reference_number as referenceNumber,
+      dfc.id as dynamicFrameworkContractId,
+			au.user_name as assignedToUserName,
+			i.title as interventionTitle,
+			rsud.first_name as serviceUserFirstName,
+			rsud.last_name as serviceUserLastName,
+			row_number() over(partition by r.id order by ra.assigned_at desc) as assigned_at_desc_seq		
+	from referral r
+			 inner join intervention i on i.id = r.intervention_id
+			 inner join referral_service_user_data rsud on rsud.referral_id = r.id
+			 inner join dynamic_framework_contract dfc on dfc.id = i.dynamic_framework_contract_id
+			 left join dynamic_framework_contract_sub_contractor dfcsc on dfcsc.dynamic_framework_contract_id = dfc.id
+			 left join referral_assignments ra on ra.referral_id = r.id
+			 left join auth_user au on au.id = ra.assigned_to_id
+			 left join end_of_service_report eosr on eosr.referral_id = r.id
+	where
+		  r.sent_at is not null
+		  and not (r.concluded_At is not null
+			and r.end_Requested_At is not null
+			and eosr.id is null)
+		and ( dfc.prime_provider_id in :serviceProviders or dfcsc.subcontractor_provider_id in :serviceProviders )
+) a where assigned_at_desc_seq = 1""",
+    nativeQuery = true
+  )
+  fun referralSummaryForServiceProviders(serviceProviders: List<String>): List<ServiceProviderSentReferralSummary>
+
   // queries for service providers
   fun findAllByInterventionDynamicFrameworkContractPrimeProviderInAndSentAtIsNotNull(providers: Iterable<ServiceProvider>): List<Referral>
   fun findAllByInterventionDynamicFrameworkContractSubcontractorProvidersInAndSentAtIsNotNull(providers: Iterable<ServiceProvider>): List<Referral>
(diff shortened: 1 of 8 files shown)


src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/jpa/repository/ReferralRepositoryTest.kt
```diff
@@ -1,59 +1,312 @@
...
+      referral.referral.referenceNumber!!,
+      referral.intervention.title,
+      referral.contract.contractReference,
+      referral.referral.currentAssignee?.id,
+      referral.serviceUser.firstName,
+      referral.serviceUser.lastName
+    )
 }
+
+data class Multiple(
```
andrewburrowsmoj, line 295: I'm a bit confused about the use of this class.
Why can't you just use Referral object?

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/jpa/repository/ReferralRepositoryTest.kt
```diff
@@ -1,59 +1,312 @@
...
+    fun `will exclude referral if draft`() {
+      val referralWithNoAssignee = createDraftReferral(true)
+
+      val summaries = referralRepository.referralSummaryForServiceProviders(listOf(referralWithNoAssignee.serviceProvider.id))
+
+      assertThat(summaries.size).isEqualTo(0)
+    }
+  }
+
+  private fun createReferral(
```
andrewburrowsmoj, line 204: why don't we have this in the factory classes, as it's what they're there for. Could be useful elsewhere.

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/jpa/repository/ReferralRepositoryTest.kt
```diff
@@ -1,59 +1,312 @@
...
+      asPrime -> dynamicFrameworkContractFactory.create(primeProvider = serviceProvider, contractReference = random(10))
+      else -> dynamicFrameworkContractFactory.create(subcontractorProviders = setOf(serviceProvider), contractReference = random(10))
+    }
+    val intervention = interventionFactory.create(contract = contract)
+    val referral = referralFactory.createDraft(intervention = intervention)
+    val serviceUser = serviceUserFactory.create(random(15), random(16), referral)
+    return Multiple(referral.currentAssignee, serviceProvider, contract, intervention, referral, serviceUser)
+  }
+
+  private fun random(length: Int): String {
```
andrewburrowsmoj, line 267: I personally don't like the use of random in tests; I think testing should be done with defined constraints, randomness introduces uncertainty.

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/jpa/repository/ReferralRepositoryTest.kt
mahalrmoj: It was just a convenience although all available via referral so will do as you say

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/jpa/repository/ReferralRepositoryTest.kt
mahalrmoj: The actual value of a service provider id adds no value to the test so rather than make them up I used random strings

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/jpa/repository/ReferralRepositoryTest.kt
mahalrmoj: Where should these go?

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/jpa/repository/ReferralRepositoryTest.kt
mahalrmoj: Done

mahalrmoj: This has now gone

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/jpa/repository/ReferralRepository.kt
```diff
@@ -1,12 +1,49 @@
...
+			serviceUserLastName from (	
+	select
+			cast(r.sent_at as TIMESTAMP WITH TIME ZONE) as sentAt,
+			r.reference_number as referenceNumber,
+      dfc.id as dynamicFrameworkContractId,
+			au.user_name as assignedToUserName,
+			i.title as interventionTitle,
+			rsud.first_name as serviceUserFirstName,
+			rsud.last_name as serviceUserLastName,
+			row_number() over(partition by r.id order by ra.assigned_at desc) as assigned_at_desc_seq
```
sldblog, line 27: Do we have a business rule to show assigned cases first?

If yes, could we use a simpler `ORDER BY`, please? `ORDER BY COALESCE(ra.assigned_at, r.sent_at) DESC`

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/jpa/repository/ReferralRepository.kt
```diff
@@ -1,12 +1,49 @@
...
+			 inner join dynamic_framework_contract dfc on dfc.id = i.dynamic_framework_contract_id
+			 left join dynamic_framework_contract_sub_contractor dfcsc on dfcsc.dynamic_framework_contract_id = dfc.id
+			 left join referral_assignments ra on ra.referral_id = r.id
+			 left join auth_user au on au.id = ra.assigned_to_id
+			 left join end_of_service_report eosr on eosr.referral_id = r.id
+	where
+		  r.sent_at is not null
+		  and not (r.concluded_At is not null
+			and r.end_Requested_At is not null
+			and eosr.id is null)
```
sldblog, line 40: probably easier to read if the "not" group is on one line (plus capitalisation for native SQL)
```suggestion
		  and not (r.concluded_at is not null and r.end_requested_at is not null and eosr.id is null)
```

sldblog (commented):
Nice! One ask about primary key IDs and one about ORDER BY

Others are nits, can be deferred

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/jpa/repository/ReferralRepository.kt
```diff
@@ -1,12 +1,49 @@
...
+			i.title as interventionTitle,
+			rsud.first_name as serviceUserFirstName,
+			rsud.last_name as serviceUserLastName,
+			row_number() over(partition by r.id order by ra.assigned_at desc) as assigned_at_desc_seq		
+	from referral r
+			 inner join intervention i on i.id = r.intervention_id
+			 inner join referral_service_user_data rsud on rsud.referral_id = r.id
+			 inner join dynamic_framework_contract dfc on dfc.id = i.dynamic_framework_contract_id
+			 left join dynamic_framework_contract_sub_contractor dfcsc on dfcsc.dynamic_framework_contract_id = dfc.id
+			 left join referral_assignments ra on ra.referral_id = r.id
```
sldblog, line 33: this join should cause each referral to duplicate in case of multiple assignments
having a superseded flag by design would help this a lot

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/jpa/repository/ReferralRepository.kt
```diff
@@ -1,12 +1,49 @@
...
+    value = """select sentAt,
+			referenceNumber,
+			assignedToUserName,
+			interventionTitle,
+			serviceUserFirstName,
+			serviceUserLastName from (	
+	select
+			cast(r.sent_at as TIMESTAMP WITH TIME ZONE) as sentAt,
+			r.reference_number as referenceNumber,
+      dfc.id as dynamicFrameworkContractId,
```
sldblog, line 23: 🙋‍♂️ `dynamicFrameworkContractId` is unused in the outer query

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/jpa/repository/ReferralRepository.kt
sldblog: I see, that's why the `row_number` was used

sldblog (approved):

PR #438, opened 2021-07-20
Use service provider id rather than name for native query to collect …

andrewburrowsmoj:
…referrals summary for sp user. This is used for the sp user dashboard

Uses service provide id (not name) to search for referrals.

To ensure dashboards return correct data


mahalrmoj (commented):
We should have a service test to cover these changes

PR #439, opened 2021-07-20
Use service provider id rather than name

andrewburrowsmoj:
Use service provider id rather than name for native query to collect referrals summary for sp user. This is used for the sp user dashboard


mahalrmoj (approved):
LGTM

PR #440, opened 2021-07-20
Handle UUIDs between native query and endpoint responses

andrewburrowsmoj:

Convert UUIDs to UUIDs in the API presentation layer.

Using UUIDs in native queries is weird; returning text values and converting them to UUIDs works.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/jpa/repository/ReferralRepository.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/jpa/repository/ReferralRepository.kt
@@ -10,17 +10,20 @@ import java.util.UUID
 
 interface ReferralRepository : JpaRepository<Referral, UUID> {
   @Query(
-    value = """select referralId, sentAt,
+    value = """select 
+      referralId, 
+      sentAt,
 			referenceNumber,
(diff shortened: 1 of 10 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/jpa/repository/ReferralRepository.kt
```diff
@@ -10,17 +10,20 @@ import java.util.UUID
...
 interface ReferralRepository : JpaRepository<Referral, UUID> {
   @Query(
-    value = """select referralId, sentAt,
+    value = """select 
+      referralId, 
+      sentAt,
 			referenceNumber,
-			assignedToUserName,
 			interventionTitle,
+      dynamicFrameworkContractId,
```
sldblog, line 18: We don't use this anywhere in `ServiceProviderSentReferralSummaryDTO`. Does this have to be here?

andrewburrowsmoj: yes, we use in ReferralAccessFilter 
(I'm just writing tests for these)

mahalrmoj (approved):
LGTM

PR #449, opened 2021-07-23
IC-1998: add sns messaging for supplier assessment

rgforsyth:

Adds ability to sent a SNS event for supplier assessment for the same scenario's as action plan appointments.

To emit SNS events for supplier assessments.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/SNSService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/SNSService.kt
@@ -8,6 +8,8 @@ import uk.gov.justice.digital.hmpps.hmppsinterventionsservice.events.ActionPlanA
 import uk.gov.justice.digital.hmpps.hmppsinterventionsservice.events.ActionPlanAppointmentEventType
 import uk.gov.justice.digital.hmpps.hmppsinterventionsservice.events.ActionPlanEvent
 import uk.gov.justice.digital.hmpps.hmppsinterventionsservice.events.ActionPlanEventType
+import uk.gov.justice.digital.hmpps.hmppsinterventionsservice.events.AppointmentEvent
+import uk.gov.justice.digital.hmpps.hmppsinterventionsservice.events.AppointmentEventType
 import uk.gov.justice.digital.hmpps.hmppsinterventionsservice.events.ReferralEvent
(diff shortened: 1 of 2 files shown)


src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/SNSActionPlanAppointmentServiceTest.kt
```diff
@@ -0,0 +1,74 @@
+package uk.gov.justice.digital.hmpps.hmppsinterventionsservice.service
```
rgforsyth, line 1: This may seem slightly confusing. This file has just been renamed and therefore has not been 'added' as part of this pr. These are the tests for the old action plan appointments sns event service

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/SNSAppointmentServiceTest.kt
```diff
@@ -9,11 +9,12 @@ import org.assertj.core.api.Assertions.assertThat
 import org.junit.jupiter.api.Test
 import uk.gov.justice.digital.hmpps.hmppsinterventionsservice.component.SNSPublisher
 import uk.gov.justice.digital.hmpps.hmppsinterventionsservice.dto.EventDTO
-import uk.gov.justice.digital.hmpps.hmppsinterventionsservice.events.ActionPlanAppointmentEvent
-import uk.gov.justice.digital.hmpps.hmppsinterventionsservice.events.ActionPlanAppointmentEventType
+import uk.gov.justice.digital.hmpps.hmppsinterventionsservice.events.AppointmentEvent
```
rgforsyth, line 12: This file is named the same as the old one was, therefore it looks like there are only minimal changes. This is a new test file used for the generic sns appointment service. It is very similar to the action plan one

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/SNSService.kt
```diff
@@ -112,3 +114,35 @@ class SNSAppointmentService(
     }
   }
 }
+
+@Service
+class SNSAppointmentService(
```
tomsmyers, line 119: it's a real shame we can't combine these events - i guess the reason is that we can't generate a link to action plan appointments without the session?

tomsmyers (approved):

PR #452, opened 2021-07-27
IC-2109 publish an event on appointment scheduling and email the responsible PP for initial assessments.

tomsmyers:

publishes an event when an appointment is scheduled (either for the first time or thereafter) and sends an email to the PP for initial assessments. 

email the PP on supplier assessments being scheduled.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/NotifyService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/NotifyService.kt
@@ -152,7 +152,9 @@ class NotifyActionPlanAppointmentService(
 class NotifyAppointmentService(
   @Value("\${notify.templates.appointment-not-attended}") private val appointmentNotAttendedTemplateID: String,
   @Value("\${notify.templates.concerning-behaviour}") private val concerningBehaviourTemplateID: String,
+  @Value("\${notify.templates.initial-assessment-scheduled}") private val initialAssessmentScheduledTemplateID: String,
   @Value("\${interventions-ui.baseurl}") private val interventionsUIBaseURL: String,
+  @Value("\${interventions-ui.locations.probation-practitioner.intervention-progress}") private val ppInterventionProgressUrl: String,
   @Value("\${interventions-ui.locations.probation-practitioner.supplier-assessment-feedback}") private val ppSAASessionFeedbackLocation: String,
   private val emailSender: EmailSender,
   private val hmppsAuthService: HMPPSAuthService,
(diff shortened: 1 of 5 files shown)


src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/NotifyAppointmentServiceTest.kt
```diff
@@ -37,6 +37,7 @@ class NotifyAppointmentServiceTest {
 
   private fun notifyService(): NotifyAppointmentService {
     return NotifyAppointmentService(
+      "template",
```
andrewburrowsmoj, line 40: I would have a different value for `template` here to confirm we're using the right templateId on line 116

andrewburrowsmoj (approved):
LGTM. I'm not sure why the lint has  decided to format it with a space in the methods.

(I was wondering how we can the existence of `28256bed-6433-485c-857d-cb52a7937dff` template on notify service?

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/NotifyAppointmentServiceTest.kt
tomsmyers: done

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/NotifyService.kt
```diff
@@ -195,10 +197,22 @@ class NotifyAppointmentService(
...
           )
         }
+        AppointmentEventType.SCHEDULED -> {
+          if (event.appointmentType == AppointmentType.SUPPLIER_ASSESSMENT) {
+            emailSender.sendEmail(
+              initialAssessmentScheduledTemplateID, ppDetails.email,
+              mapOf(
+                "ppFirstName" to ppDetails.firstName,
+                "referenceNumber" to referral.referenceNumber!!,
+                "referralUrl" to generateResourceUrl(interventionsUIBaseURL, ppReferralDetailsUrl, referral.id).toString(),
```
tomsmyers, line 211: @andrewburrowsmoj i added this line since your review based on a conversation with D and Leigh.

andrewburrowsmoj: Ok, that makes sense - especially since there is no way to navigate from feedback details to referral details.

PR #454, opened 2021-07-27
IC-2063 send PP emails to delius responsible officer instead of referral sender

tomsmyers:

send PP emails to delius responsible officer instead of referral sender.

there has to be a bunch of defensive code in here to handle cases where RO names or email addresses are missing. we are waiting on info from the business about how to handle these cases (i.e. do we fall back to referral sender)

get the emails to the most relevent PPs

## Probation context

We are told that people-on-probation (newer term for service users) can be managed by different responsible officers over time. This might be due to workload management of these officers, or other factors. To ensure we and our service providers contact the _current_ person, we need to use the "responsible officer" for a person-on-probation.

We are assuming there's exactly one on record in Delius.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/CommunityAPIOffenderService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/CommunityAPIOffenderService.kt
@@ -18,6 +18,12 @@ data class ServiceUserAccessResult(
   val messages: List<String>,
 )
 
+data class ResponsibleOfficer(
+  val firstName: String?,
+  val email: String?,
+  val staffId: Long?,
+)
+
 private data class UserAccessResponse(
   val exclusionMessage: String?,
   val restrictionMessage: String?,
@@ -27,12 +33,27 @@ private data class StaffDetailsResponse(
   val staffIdentifier: String,
 )
 
+private data class ContactableHumanResponse(
+  val forenames: String?,
+  val surname: String?,
+  val email: String?,
+  val phoneNumber: String?,
+)
+
+private data class OffenderManagerResponse(
+  val staff: ContactableHumanResponse?,
+  val staffId: Long?,
+  val isResponsibleOfficer: Boolean?,
+)
+
 @Service
 class CommunityAPIOffenderService(
   @Value("\${community-api.locations.offender-access}") private val offenderAccessLocation: String,
   @Value("\${community-api.locations.managed-offenders}") private val managedOffendersLocation: String,
   @Value("\${community-api.locations.staff-details}") private val staffDetailsLocation: String,
+  @Value("\${community-api.locations.offender-managers}") private val offenderManagersLocation: String,
   private val communityApiClient: RestClient,
+  private val telemetryService: TelemetryService,
 ) {
   fun checkIfAuthenticatedDeliusUserHasAccessToServiceUser(user: AuthUser, crn: String): ServiceUserAccessResult {
     val userAccessPath = UriComponentsBuilder.fromPath(offenderAccessLocation)
@@ -85,5 +106,38 @@ class CommunityAPIOffenderService(
       }
       .block()
       ?.staffIdentifier
+  }
+
+  fun getResponsibleOfficer(crn: String): ResponsibleOfficer {
+    val offenderManagersPath = UriComponentsBuilder.fromPath(offenderManagersLocation)
+      .buildAndExpand(crn)
+      .toString()
+
+    val responsibleOfficers = communityApiClient.get(offenderManagersPath)
+      .retrieve()
+      .bodyToFlux(OffenderManagerResponse::class.java)
+      .filter { it.isResponsibleOfficer == true }
+      .map { ResponsibleOfficer(it.staff?.forenames?.substringBefore(' '), it.staff?.email, it.staffId) }
+      .collectList()
+      .block()
+
+    // as with many community API endpoints, we have a few assumptions about the data.
+    // if there are no ROs, we can't recover. if there are more than one, we just take
+    // the first.
+    responsibleOfficers.size.let {
+      when {
+        it == 0 -> telemetryService.reportInvalidAssumption(
+          "service users always have a responsible officer",
+          mapOf("crn" to crn),
+          recoverable = false
+        )
+        it > 1 -> telemetryService.reportInvalidAssumption(
+          "service users only have one responsible officer",
+          mapOf("crn" to crn, "numberOfResponsibleOfficers" to it.toString()),
+        )
+      }
+    }
+
+    return responsibleOfficers.first()
   }
 }
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/NotifyService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/NotifyService.kt
@@ -35,13 +35,14 @@ class NotifyActionPlanService(
   @Value("\${interventions-ui.locations.service-provider.action-plan}") private val spActionPlanLocation: String,
   private val emailSender: EmailSender,
   private val hmppsAuthService: HMPPSAuthService,
+  private val referralService: ReferralService,
 ) : ApplicationListener<ActionPlanEvent>, NotifyService {
 
   @AsyncEventExceptionHandling
   override fun onApplicationEvent(event: ActionPlanEvent) {
     when (event.type) {
       ActionPlanEventType.SUBMITTED -> {
-        val recipient = hmppsAuthService.getUserDetail(event.actionPlan.referral.sentBy!!)
+        val recipient = referralService.getResponsibleProbationPractitioner(event.actionPlan.referral)
         val location = generateResourceUrl(interventionsUIBaseURL, ppActionPlanLocation, event.actionPlan.referral.id)
         emailSender.sendEmail(
           actionPlanSubmittedTemplateID,
@@ -76,20 +77,20 @@ class NotifyEndOfServiceReportService(
   @Value("\${interventions-ui.baseurl}") private val interventionsUIBaseURL: String,
   @Value("\${interventions-ui.locations.probation-practitioner.end-of-service-report}") private val ppEndOfServiceReportLocation: String,
   private val emailSender: EmailSender,
-  private val hmppsAuthService: HMPPSAuthService,
+  private val referralService: ReferralService,
 ) : ApplicationListener<EndOfServiceReportEvent>, NotifyService {
 
   @AsyncEventExceptionHandling
   override fun onApplicationEvent(event: EndOfServiceReportEvent) {
     when (event.type) {
       EndOfServiceReportEventType.SUBMITTED -> {
-        val userDetail = hmppsAuthService.getUserDetail(event.endOfServiceReport.referral.getResponsibleProbationPractitioner())
+        val recipient = referralService.getResponsibleProbationPractitioner(event.endOfServiceReport.referral)
         val location = generateResourceUrl(interventionsUIBaseURL, ppEndOfServiceReportLocation, event.endOfServiceReport.id)
         emailSender.sendEmail(
           endOfServiceReportSubmittedTemplateID,
-          userDetail.email,
+          recipient.email,
           mapOf(
-            "ppFirstName" to userDetail.firstName,
+            "ppFirstName" to recipient.firstName,
             "referralReference" to event.endOfServiceReport.referral.referenceNumber!!,
             "endOfServiceReportLink" to location.toString(),
           )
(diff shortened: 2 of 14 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/NotifyService.kt
```diff
@@ -20,10 +20,18 @@ import uk.gov.justice.digital.hmpps.hmppsinterventionsservice.exception.AsyncEve
...
+open class NotifyService {
   fun generateResourceUrl(baseURL: String, path: String, vararg args: Any): URI {
     return UriComponentsBuilder.fromHttpUrl(baseURL).path(path).buildAndExpand(*args).toUri()
   }
+
+  fun responsibleOfficerFirstName(responsibleOfficer: ResponsibleOfficer): String {
+    return responsibleOfficer.name?.split(' ')?.first() ?: ""
+  }
+
+  fun responsibleOfficerEmail(responsibleOfficer: ResponsibleOfficer): String {
```
tomsmyers, line 32: this is the defensive code i mentioned

tomsmyers: (and above)

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/CommunityAPIOffenderService.kt
```diff
@@ -86,4 +104,24 @@ class CommunityAPIOffenderService(
...
+
+  fun getResponsibleOfficer(crn: String): ResponsibleOfficer {
+    val offenderManagersPath = UriComponentsBuilder.fromPath(offenderManagersLocation)
+      .buildAndExpand(crn)
+      .toString()
+
+    // technically this endpoint can return more than one offender manager that
+    // is a responsible officer. in practice this should not happen. if it does
+    // we just take the first one we come across.
+    return communityApiClient.get(offenderManagersPath)
```
andrewburrowsmoj, line 116: Possibly could order it by `fromDate` on `CommunityOrPrisonOffenderManager` and get the latest that is set in the past.

It seems this is populated from `ALLOCATION_DATE` from `OFFENDER_MANAGER` table. 

The `END_DATE` is skipped in population of the POJO, so maybe can make a change to include it to also filter out "ended" responsibile officers.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/CommunityAPIOffenderService.kt
```diff
@@ -86,4 +104,24 @@ class CommunityAPIOffenderService(
...
+      .retrieve()
+      .bodyToFlux(OffenderManagerResponse::class.java)
+      .filter { it.isResponsibleOfficer == true }
+      .map { ResponsibleOfficer(parseName(it.staff?.forenames, it.staff?.surname), it.staff?.email) }
+      .blockFirst()
+  }
+
+  private fun parseName(forenames: String?, surname: String?): String? {
```
andrewburrowsmoj, line 124: Do you have any tests for this? 

(I also think it would be helpful to contain the unit tests in the same commit)

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/CommunityAPIOffenderService.kt
tomsmyers: i don't think this is quite accurate - the response from 'allOffenderManagers' first filters OM's by `endDate == null && Optional.ofNullable(activeFlag).orElse(0L) == 1L && !isDeleted()` (`OffenderManager::isActive`), then populates 'isResponsibleOfficer' by checking `endDateTime == null` (`ResponsibleOfficer::isActive`). so ended OMs and ROs are not included. 

whilst there are no DB constraints for ensuring there cannot be more than one active RO at any given time, the only way to update ROs via the API ensure that previous ROs have end dates set. so i am totally happy that this method is sufficient without adding additional complexity.

worst case that there are multiple active ROs isn't that impactful, since we still select one of them.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/CommunityAPIOffenderService.kt
tomsmyers: currently no - i can add. i'll move the code to a public utility method called 'mungeNames' or something.

on the latter i agree - but the prospect of the rebase and potential manual merges put me off spending the time to do it. how important do you think this is in terms of the commit history ?

andrewburrowsmoj: The last comment was just as a general rule as it makes it easier to review - I was not expecting you to make that change in this PR.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/CommunityAPIOffenderService.kt
andrewburrowsmoj: I understand now, thank you

andrewburrowsmoj (approved):

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/CommunityAPIOffenderService.kt
```diff
@@ -86,4 +106,20 @@ class CommunityAPIOffenderService(
       .block()
       ?.staffIdentifier
   }
+
+  fun getResponsibleOfficer(crn: String): ResponsibleOfficer {
+    val offenderManagersPath = UriComponentsBuilder.fromPath(offenderManagersLocation)
```
sldblog, line 111: _nit_ Can we inline `offenderManagersLocation`? They won't ever change and should be used through a single code path anyway, right?
```suggestion
    val offenderManagersPath = UriComponentsBuilder.fromPath("/secure/offenders/crn/{crn}/allOffenderManagers")
```

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/NotifyService.kt
```diff
@@ -20,10 +20,18 @@ import uk.gov.justice.digital.hmpps.hmppsinterventionsservice.exception.AsyncEve
...
 import java.net.URI
 
-interface NotifyService {
+open class NotifyService {
   fun generateResourceUrl(baseURL: String, path: String, vararg args: Any): URI {
     return UriComponentsBuilder.fromHttpUrl(baseURL).path(path).buildAndExpand(*args).toUri()
   }
+
+  fun responsibleOfficerFirstName(responsibleOfficer: ResponsibleOfficer): String {
+    return responsibleOfficer.name?.split(' ')?.first() ?: ""
```
sldblog, line 29: This name is constructed during `ResponsibleOfficer(stringUtils.joinNullable(it.staff?.forenames, it.staff?.surname), ...)` and then split here -- can we pass down first/last names separately to avoid splitting after joining?

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/NotifyService.kt
```diff
@@ -20,10 +20,18 @@ import uk.gov.justice.digital.hmpps.hmppsinterventionsservice.exception.AsyncEve
...
+  }
+
+  fun responsibleOfficerEmail(responsibleOfficer: ResponsibleOfficer): String {
+    return responsibleOfficer.email ?: throw RuntimeException("no email address for service user's responsible officer")
```
sldblog, line 33: Could we throw an exception class that is semantically explicit about "our assumptions about received data were invalid"? For simpler monitoring. Maybe:
```suggestion
    return responsibleOfficer.email ?: throw ExternalDataExpectationFailed("no email address for service user's responsible officer")
```
or -- we could use the `InterventionsDataAssumptionInvalid` telemetry in the other comment

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/CommunityAPIOffenderService.kt
```diff
@@ -86,4 +106,20 @@ class CommunityAPIOffenderService(
...
+
+    // technically this endpoint can return more than one offender manager that
+    // is a responsible officer. in practice this should not happen. if it does
+    // we just take the first one we come across.
+    return communityApiClient.get(offenderManagersPath)
+      .retrieve()
+      .bodyToFlux(OffenderManagerResponse::class.java)
+      .filter { it.isResponsibleOfficer == true }
+      .map { ResponsibleOfficer(stringUtils.joinNullable(it.staff?.forenames, it.staff?.surname), it.staff?.email) }
+      .blockFirst()
```
sldblog, line 123: Given the amount of data-related assumptions we had before, could we pop up a "our assumption about the data was invalid" telemetry here if there are `!= 1` matches?

Feels like this would make our assumptions explicit and monitorable

```kt
telemetryClient.trackEvent(
  "InterventionsDataAssumptionInvalid",
  mapOf(
    "message": "expected only one responsible officer",
    "got": "2",
    // plus CRN and referralId
  ),
  null
)
```

sldblog:
📝 Added a "probation context" section to the description

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/CommunityAPIOffenderService.kt
tomsmyers: this follows existing patterns which group urls in config e.g. 

```
community-api:
  locations:
    sent-referral: "/secure/offenders/crn/{crn}/referral/start/context/{contextName}"
    ended-referral: "/secure/offenders/crn/{crn}/referral/end/context/{contextName}"
    reschedule-appointment: "/secure/offenders/crn/{crn}/appointments/{appointmentId}/reschedule/context/{contextName}"
    appointment-outcome-request: "/secure/offenders/crn/{crn}/appointments/{appointmentId}/outcome/context/{contextName}"
    notification-request: "/secure/offenders/crn/{crn}/sentences/{sentenceId}/notifications/context/{contextName}"
    offender-access: "/secure/offenders/crn/{crn}/user/{username}/userAccess"
    managed-offenders: "/secure/staff/staffIdentifier/{staffIdentifier}/managedOffenders"
    staff-details: "/secure/staff/username/{username}"
```

or 

```
hmppsauth:
  api:
    locations:
      auth-user-groups: "/api/authuser/{username}/groups"
      auth-user-detail: "/api/authuser/{username}"
      user-email: "/api/user/{username}/email"
      user-detail: "/api/user/{username}"
 ```

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/NotifyService.kt
tomsmyers: we still need to split because the firstname can be any number of names.

i was trying my best to come up with the most sane representation of the name in the ResponsibleOfficer data class, regardless of how it's used. 

i have a habit of optimizing for stuff like this too but in reality string joining and splitting is the least of our worries when it comes to performance.

IMO

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/NotifyService.kt
tomsmyers: like this idea - i'll have a play arond with it

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/NotifyService.kt
sldblog: I guess I'm worried about getting "surnames" in the mix -- if someone has `forenames="", surnames="Smith"`, addressing them as "Smith" might be awkward

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/CommunityAPIOffenderService.kt
sldblog: yes -- do we benefit from that setup?

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/NotifyService.kt
tomsmyers: in your example case, would it be better to address them as `""` or `"smith"` though? at least it shows some reassurance they are the intended recipient of the email.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/CommunityAPIOffenderService.kt
tomsmyers: we can ask the team what they think. it doesn't bother me having to look them up, and i like having them all together. but i can see your point.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/CommunityAPIOffenderService.kt
tomsmyers: done

PR #459, opened 2021-07-28
IC-2110: Provided nps office code to community-api when rescheduling an appointment

andrewburrowsmoj:

If the SP user reschedules an appointment with a different npsOffice, then this needs to be updated on delius.
Even where the timings are the same, it should still be considered a reschedule.

Provide ability to reschedule with an NPS office location. (Phone call, Video call and in person meeting all use the default nps office code)

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/CommunityAPIBookingService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/CommunityAPIBookingService.kt
@@ -36,20 +36,20 @@ class CommunityAPIBookingService(
       return existingAppointment?.deliusAppointmentId
     }
 
-    return processingBooking(referral, existingAppointment, appointmentTime, durationInMinutes, appointmentType, npsOfficeCode ?: this.defaultOfficeLocation)
+    return processingBooking(referral, existingAppointment, appointmentTime, durationInMinutes, appointmentType, npsOfficeCode)
   }
 
-  private fun processingBooking(referral: Referral, existingAppointment: Appointment?, appointmentTime: OffsetDateTime, durationInMinutes: Int, appointmentType: AppointmentType, npsOfficeCode: String): Long? {
+  private fun processingBooking(referral: Referral, existingAppointment: Appointment?, appointmentTime: OffsetDateTime, durationInMinutes: Int, appointmentType: AppointmentType, npsOfficeCode: String?): Long? {
     return existingAppointment?.let {
-      if (!isRescheduleBooking(existingAppointment, appointmentTime, durationInMinutes)) {
+      if (!isRescheduleBooking(existingAppointment, appointmentTime, durationInMinutes, npsOfficeCode)) {
         // nothing to do !
         return existingAppointment.deliusAppointmentId
       }
 
-      val appointmentRequestDTO = buildAppointmentRescheduleRequestDTO(appointmentTime, durationInMinutes, npsOfficeCode)
+      val appointmentRequestDTO = buildAppointmentRescheduleRequestDTO(appointmentTime, durationInMinutes, npsOfficeCode ?: defaultOfficeLocation)
       makeBooking(referral.serviceUserCRN, it.deliusAppointmentId!!, appointmentRequestDTO, communityApiRescheduleAppointmentLocation)
     } ?: run {
-      val appointmentRequestDTO = buildAppointmentCreateRequestDTO(referral, appointmentTime, durationInMinutes, appointmentType, npsOfficeCode)
+      val appointmentRequestDTO = buildAppointmentCreateRequestDTO(referral, appointmentTime, durationInMinutes, appointmentType, npsOfficeCode ?: defaultOfficeLocation)
       makeBooking(referral.serviceUserCRN, referral.relevantSentenceId!!, appointmentRequestDTO, communityApiBookAppointmentLocation)
     }
   }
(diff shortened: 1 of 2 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/CommunityAPIBookingService.kt
```diff
@@ -81,11 +81,11 @@ class CommunityAPIBookingService(
   }
 
   private fun buildAppointmentRescheduleRequestDTO(appointmentTime: OffsetDateTime, durationInMinutes: Int, npsOfficeCode: String): AppointmentRescheduleRequestDTO {
-    // TODO: CommunityApi and Delius currently don't allow for rescheduling with a different location
     return AppointmentRescheduleRequestDTO(
       updatedAppointmentStart = appointmentTime,
       updatedAppointmentEnd = appointmentTime.plusMinutes(durationInMinutes.toLong()),
       initiatedByServiceProvider = true, // fixme - needs to come from the user - defaulted to SP Initiated Reschedule
+      officeLocationCode = npsOfficeCode
```
sldblog, line 88: To clarify, are we intending to send data containing "delius location" or "probation office" codes?

To my knowledge,

- _probation office_ refers to a building, a venue on its own
- _delius location_ refers to the composite ID identifying _probation office_ used by a _provider_

I'd _love_ if we could use the _probation office_ but if we have to use the _location_, could we rename the variable across the board (and optionally create a type alias?), please?

sldblog: (I do believe "office" is the right abstraction here, but it seems the community-api endpoint requires something else?)

andrewburrowsmoj: Would you prefer that we call it `probationOffice` rather than `npsOfficeCode`?

andrewburrowsmoj: I original did a change in an old PR where I used `deliusOfficeLocationCode` for variable name, but I reverted that and  kept the usage of `npsOfficeCode`

I remember the change was quite extensive and touching a lot of classes. 
Given it doesn't change the function, are you okay with this being in another PR?

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/CommunityAPIBookingService.kt
```diff
@@ -101,10 +101,13 @@ class CommunityAPIBookingService(
...
   }
 
-  fun isRescheduleBooking(existingAppointment: Appointment, appointmentTime: OffsetDateTime, durationInMinutes: Int): Boolean =
-    isDifferentTimings(existingAppointment, appointmentTime, durationInMinutes)
+  fun isRescheduleBooking(existingAppointment: Appointment, appointmentTime: OffsetDateTime, durationInMinutes: Int, npsOfficeCode: String?): Boolean =
+    isDifferentTimings(existingAppointment, appointmentTime, durationInMinutes) || isDifferentLocation(existingAppointment, npsOfficeCode)
 
-  fun isDifferentTimings(existingAppointment: Appointment, appointmentTime: OffsetDateTime, durationInMinutes: Int): Boolean =
+  private fun isDifferentLocation(existingAppointment: Appointment, npsOfficeCode: String?): Boolean {
+    return !npsOfficeCode.equals(existingAppointment.appointmentDelivery?.npsOfficeCode)
```
mahalrmoj, line 108: npsOfficeCode is defined as nullable - should this check be made safe

mahalrmoj: the only place npsOfficeCode should not be null is on the main method beacuse then it gets defaulted?

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/CommunityAPIBookingService.kt
```diff
@@ -129,6 +132,7 @@ data class AppointmentRescheduleRequestDTO(
   val updatedAppointmentStart: OffsetDateTime?,
   val updatedAppointmentEnd: OffsetDateTime?,
   val initiatedByServiceProvider: Boolean,
+  val officeLocationCode: String?,
```
mahalrmoj, line 135: Does this still need to be nullable?

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/CommunityAPIBookingService.kt
andrewburrowsmoj: Sorry, I a bit unsure of the question.

`npsOfficeCode` being null means the appointment was either a phone call, video call or custom office location.
`npsOfficeCode` not being null means the appointment was in a defined PP office.


I've changed the assignment of the default `npsOfficeCode` to only be set when defining the DTO request to send to delius.
We do not store the default npsOfficeCode in the database when it's a non-PP office. So I need to have this method return true when the appointment is a phone call and the appointment change is also a phone call. In which case they would both be null.

Perhaps the variable would make more sense to be called: `userSuppliedNpsOfficeCode` rather than `npsOfficeCodeToSubmitToDelius`.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/CommunityAPIBookingService.kt
andrewburrowsmoj: Yes, the reason I made it nullable here was to reflex the usage of the DTO on `CommunityApi` which allows nullable.

I agree that it should be non-nullable as that was a backwards compatible change you did.

(I think also updateAppointmentStart and End should be non-nullable too)

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/CommunityAPIBookingService.kt
andrewburrowsmoj: discussion on slack:

David is happy for this to be in a future PR

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/CommunityAPIBookingService.kt
mahalrmoj: LGTM

mahalrmoj (approved):
LGTM

PR #460, opened 2021-07-28
Add superseded flag to assignments

sldblog:

Add superseded flag to assignments

**But** not backfill "superseded" flags to true, yet, to avoid gaps created by old deployment still in use after the migration runs.

To make it easier to select the latest assignment in queries (which should remove the need for `row_number` and be able to do a JOIN ON without subqueries) and debugging sessions


tomsmyers (approved):
nice - thanks !

do you prefer `AND superseded = 'false'` or `AND latest = 'true'` for this?

sldblog:
I don't know; it feels relying on something that is "true by default" and set everything to "false" when it gets updated somehow feels like a weaker interaction as I'm "negating" a previous state. It's a coin flip though. What do we prefer?

| using "latest" | using "superseded" |
| --- | --- |
| update current entries to `latest = false` | update current entries to `superseded = true` |
| create new entry with `latest = true` | create new entry with `superseded = false` |

PR #461, opened 2021-07-28
IC-0000 Set HmppsAuth Retry to 2 after Timeouts of 3 seconds

mahalrmoj:
This PR causes all hmpps-auth calls to timeout after 3 seconds and then to retry 2 times.

R&M have seen an occassional big fluctuation in the response to time.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/config/WebClientConfiguration.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/config/WebClientConfiguration.kt
@@ -19,8 +19,8 @@ import java.time.Duration
 
 @Configuration
 class WebClientConfiguration(
-  @Value("\${webclient.connect-timeout-seconds}") private val connectTimeoutSeconds: Long,
-  @Value("\${webclient.read-timeout-seconds}") private val readTimeoutSeconds: Int,
+  @Value("\${webclient.connect-timeout-seconds}") private val defaultConnectTimeoutSeconds: Long,
+  @Value("\${webclient.read-timeout-seconds}") private val defaultReadTimeoutSeconds: Int,
   @Value("\${webclient.hmpps-auth.read-timeout-seconds}") private val hmppsReadTimeoutSeconds: Int,
   @Value("\${webclient.hmpps-auth.connect-timeout-seconds}") private val hmppsAuthConnectTimeoutSeconds: Long,
   @Value("\${webclient.write-timeout-seconds}") private val writeTimeoutSeconds: Int,
@@ -71,15 +71,11 @@ class WebClientConfiguration(
     return authorizedClientManager
   }
 
-  private fun createAuthorizedWebClient(clientManager: OAuth2AuthorizedClientManager, baseUrl: String): WebClient {
-    return createAuthorizedWebClient(clientManager, baseUrl, readTimeoutSeconds, connectTimeoutSeconds)
-  }
-
   private fun createAuthorizedWebClient(
     clientManager: OAuth2AuthorizedClientManager,
     baseUrl: String,
-    readTimeoutSeconds: Int,
-    connectTimeoutSeconds: Long,
+    readTimeoutSeconds: Int = defaultReadTimeoutSeconds,
+    connectTimeoutSeconds: Long = defaultConnectTimeoutSeconds,
   ): WebClient {
     val oauth2Client = ServletOAuth2AuthorizedClientExchangeFilterFunction(clientManager)
 
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/config/WebClientConfiguration.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/config/WebClientConfiguration.kt
@@ -21,6 +21,8 @@ import java.time.Duration
 class WebClientConfiguration(
   @Value("\${webclient.connect-timeout-seconds}") private val connectTimeoutSeconds: Long,
   @Value("\${webclient.read-timeout-seconds}") private val readTimeoutSeconds: Int,
+  @Value("\${webclient.hmpps-auth.read-timeout-seconds}") private val hmppsReadTimeoutSeconds: Int,
+  @Value("\${webclient.hmpps-auth.connect-timeout-seconds}") private val hmppsAuthConnectTimeoutSeconds: Long,
   @Value("\${webclient.write-timeout-seconds}") private val writeTimeoutSeconds: Int,
   @Value("\${community-api.baseurl}") private val communityApiBaseUrl: String,
   @Value("\${hmppsauth.baseurl}") private val hmppsAuthBaseUrl: String,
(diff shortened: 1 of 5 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/config/WebClientConfiguration.kt
```diff
@@ -70,6 +72,15 @@ class WebClientConfiguration(
   }
 
   private fun createAuthorizedWebClient(clientManager: OAuth2AuthorizedClientManager, baseUrl: String): WebClient {
```
tomsmyers, line 74: i think this would be neater as a single method with default parameters i.e. 

```
  private fun createAuthorizedWebClient(
    clientManager: OAuth2AuthorizedClientManager,
    baseUrl: String,
    readTimeoutSeconds: Int = readTimeoutSeconds,
    connectTimeoutSeconds: Long = connectTimeoutSeconds,
  ): WebClient
```

tomsmyers: there is some name shadowing going on here too which would be good to avoid. perhaps we could call the existing configuration vars `defaultReadTimeoutSeconds` etc

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/HMPPSAuthService.kt
```diff
@@ -56,6 +57,7 @@ class HMPPSAuthService(
       .retrieve()
       .onStatus({ HttpStatus.NOT_FOUND == it }, { Mono.just(null) })
       .bodyToFlux(AuthGroupResponse::class.java)
+      .retry(maxRetryAttempts)
```
tomsmyers, line 60: i wonder how we could get these 'auto-applied' to all calls for this client? i wonder if there's a way by adding a filter function to the base webclient ?

tomsmyers (commented):
looks good - couple of nitty comments. 

one thing which i think is missing from this though (which is somewhat trickier to figure out) is how to implement retries for the hmpps auth token exchange. it's all in spring config, so we can't just stick a retry on a reactive call, but there must be a way to do it. the motivation behind doing this would be to avoid these errors which crop up infrequently in prod `[invalid_token_response] An error occurred while attempting to retrieve the OAuth 2.0 Access Token Response: 500`.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/HMPPSAuthService.kt
mahalrmoj: Isn't this is a feature of the flux rather than the client consuming it.

mahalrmoj:
> An error occurred while attempting to retrieve the OAuth
It seems to be hardwired to 30 seconds
```
	private com.nimbusds.oauth2.sdk.TokenResponse getTokenResponse(AuthorizationGrant authorizationCodeGrant,
			URI tokenUri, ClientAuthentication clientAuthentication) {
		try {
			// Send the Access Token request
			TokenRequest tokenRequest = new TokenRequest(tokenUri, clientAuthentication, authorizationCodeGrant);
			HTTPRequest httpRequest = tokenRequest.toHTTPRequest();
			httpRequest.setAccept(MediaType.APPLICATION_JSON_VALUE);
			httpRequest.setConnectTimeout(30000);
			httpRequest.setReadTimeout(30000);
			return com.nimbusds.oauth2.sdk.TokenResponse.parse(httpRequest.send());
		}
		catch (ParseException | IOException ex) {
			OAuth2Error oauth2Error = new OAuth2Error(INVALID_TOKEN_RESPONSE_ERROR_CODE,
					"An error occurred while attempting to retrieve the OAuth 2.0 Access Token Response: "
							+ ex.getMessage(),
					null);
			throw new OAuth2AuthorizationException(oauth2Error, ex);
		}
	}

```

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/HMPPSAuthService.kt
mahalrmoj: This is now changed to do retryWhen for TimeoutException and ConnectTimeoutException

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/HMPPSAuthService.kt
```diff
@@ -104,15 +115,43 @@ class HMPPSAuthService(
...
+
+  fun <T> Flux<T>.withRetryPolicy(): Flux<T> {
+    return this
+      .retryWhen(
+        Retry.max(maxRetryAttempts)
+          .filter {
+            it.cause is ReadTimeoutException || it.cause is ConnectTimeoutException
+          }
+          .doBeforeRetry { retrySignal ->
+            logger.debug("Retrying: ${retrySignal.totalRetries()}, ${retrySignal.totalRetriesInARow()}, ${retrySignal.failure()}")
```
tomsmyers, line 140: nit: structured logs preferred

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/HMPPSAuthService.kt
```diff
@@ -104,15 +117,52 @@ class HMPPSAuthService(
...
+        Retry.max(maxRetryAttempts)
+          .filter { isTimeoutException(it) }
+          .doBeforeRetry { logRetrySignal(it) }
+      )
+  }
+
+  fun isTimeoutException(it: Throwable): Boolean {
+    // Timeout for NO_RESPONSE is wrapped in a WebClientRequestException
+    return it is ReadTimeoutException || it is ConnectTimeoutException ||
+      it.cause is ReadTimeoutException || it.cause is ConnectTimeoutException
```
sldblog, line 155: Do we need to go into causes?

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/HMPPSAuthService.kt
```diff
@@ -104,15 +117,52 @@ class HMPPSAuthService(
...
+      .retryWhen(
+        Retry.max(maxRetryAttempts)
+          .filter { isTimeoutException(it) }
+          .doBeforeRetry { logRetrySignal(it) }
+      )
+  }
+
+  fun isTimeoutException(it: Throwable): Boolean {
+    // Timeout for NO_RESPONSE is wrapped in a WebClientRequestException
+    return it is ReadTimeoutException || it is ConnectTimeoutException ||
```
sldblog, line 154: `io.netty.handler.timeout.TimeoutException` and `java.net.ConnectException` are probably change-safe "intent-specific" superclasses

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/HMPPSAuthService.kt
sldblog: Asking because if we do, it's a recursive data structure, so ... is it sufficient to stop at level 1?

mahalrmoj: In the case of NO_RESPONSE (as apposed to delayed) spring does seem to wrap it with a WebClientRequestException so we do need to look inside.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/HMPPSAuthService.kt
mahalrmoj: They are very netty specific and hang off io.netty.channel.ChannelException which in turn extends RuntimeException.

tomsmyers (approved):

PR #469, opened 2021-08-02
Bug fix attendance feedback submitted fields

andrewburrowsmoj:

Fixes two bugs:

1. The attendance feedback submitted by was being populated from the `attendanceSubmittedBy` and not the `attendanceFeedbackSubmittedBy`. The risk of this is `NONE` because we currently don't do anything with the `submittedBy` field (UI PR is ready but the pact test failed), in addition it would likely be the same officer in 99% of cases.

2. The second bug is where the feedback submitted boolean field was being populated to true or false based on the `attendanceSubmittedAt` and not the `attendanceFeedbackSubmittedAt`. This only applies to action plan appointments and not initial assessment appointments.

This is more serious because it has the effect of the feedback shown as `COMPLETED` if the user fills in the attendance fields but has not yet submitted the feedback. This would occur if the user presses back after completing attendance feedback or if they close their browser.

I tested this locally with the above scenario and this is what I see for the feedback:

image

I should expect to be able to `Give feedback` with status `SCHEDULED`.

Prevent users from being locked out of completing action plan session feedback if they press back.


tomsmyers (approved):
good catch - thank yoU!

PR #470, opened 2021-08-02
IC-1116 launch service provider reporting job with an async job launcher

tomsmyers:

launch service provider reporting job with an async job launcher instead of the default synchronous launcher

ensure the reporting controller method returns without waiting for the job to complete.


andrewburrowsmoj (approved):
LGTM.
Though I think I'd prefer to be explicit with naming the bean and a qualifier in the annotation. Thus if it can't find it then it fails (maybe something goes wrong with the configuration but it's ignored) ?

PR #473, opened 2021-08-03
IC-1116 add explicit migrations for spring batch database tables

tomsmyers:

add explicit migrations for spring batch database tables taken from the spring batch repo.

this also fixes the issue detailed here https://mojdt.slack.com/archives/C01DYKJUKDX/p1627982089042100

doing this ourselves means we do not rely on spring batch to manage our database schema for us, and we can slot this migration in with our regular flyway migrations making it visible, repeatable and revertable.


sldblog (approved):

src/main/resources/application.yml
```diff
@@ -43,7 +43,7 @@ spring:
     job:
       enabled: false  # don't run jobs at application startup
     jdbc:
-      initialize-schema: always
+      initialize-schema: never # we manage the spring batch schema with our own migrations
```
sldblog, line 46: can we add a test for this in the `validate_db` job?

tomsmyers: i think this case is covered with existing integration tests. 

these tests were previously failing before i introduced the changes in the test config.

PR #479, opened 2021-08-04
IC-2135: Added appointment session type to appointment table.

andrewburrowsmoj:
This is a required parameter when creating or updating an appointment.

The types can either be ONE_TO_ONE or GROUP.

I've defaulted the existing appointments to ONE_TO_ONE but there's no way of knowing the session type of historical appointments.

The UI PR can be found here: https://github.com/ministryofjustice/hmpps-interventions-ui/pull/652

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/controller/ActionPlanSessionController.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/controller/ActionPlanSessionController.kt
@@ -40,6 +40,7 @@ class ActionPlanSessionController(
       updateAppointmentDTO.durationInMinutes,
       user,
       updateAppointmentDTO.appointmentDeliveryType,
+      updateAppointmentDTO.sessionType,
       updateAppointmentDTO.appointmentDeliveryAddress,
       updateAppointmentDTO.npsOfficeCode,
     )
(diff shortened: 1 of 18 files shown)


src/main/resources/db/migration/V1_74__appointment_session_type.sql
```diff
@@ -0,0 +1,5 @@
+CREATE TYPE appointment_session_type AS ENUM ('ONE_TO_ONE','GROUP');
+
+alter table appointment_delivery add column appointment_session_type appointment_session_type default 'ONE_TO_ONE';
+
+alter table appointment_delivery alter column appointment_session_type drop default;
```
sldblog, line 5: _zero downtime deploy behaviour_

🙋‍♂️ As far as I can see, it's possible that
- old service code
- running after this migration executed
- handles a delivery creation

In that case, the old code wouldn't supply the `appointment_session_type` column, and would then result in a SQL error without a default.

It's safer to either
- `drop default` in a subsequent separate deployment or post-rollout migration
- execute creating the new column migration before rolling out this code + `drop default`

andrewburrowsmoj: ok. I've removed the `drop default`.

tomsmyers (approved):
LGTM - will there be a follow up commit here to drop the default value then?

PR #485, opened 2021-08-05
IC-1116 mock 'ServiceProviderAccessScopeMapper' in contract tests

tomsmyers:

 mock 'ServiceProviderAccessScopeMapper' in contract tests

a new contract test for downloading the performance report calls this code path and it needs to be mocked. 

mocking this is nothing to worry about in these tests. we aren't testing the access logic, just validating the API contracts.


andrewburrowsmoj (approved):
Ok, makes sense given that it's a contract test and we have integration tests for access rights.

PR #488, opened 2021-08-06
IC-2165: Added ability to relocate an existing appointment

andrewburrowsmoj:

Added a new community-api endpoint for relocation of an existing appointment.

When rescheduling an appointment but only changing the location (npsOfficeCode) then we need to send a different request to delius. This is because delius only considers rescheduling to be applicable if the appointment timing details are different.

Unblock the ability to schedule an appointment for an NPS office location.

### DIFF
--- a/src/main/resources/application.yml
+++ b/src/main/resources/application.yml
@@ -115,6 +115,7 @@ community-api:
     ended-referral: "/secure/offenders/crn/{crn}/referral/end/context/{contextName}"
     book-appointment: "/secure/offenders/crn/{crn}/sentence/{sentenceId}/appointments/context/{contextName}"
     reschedule-appointment: "/secure/offenders/crn/{crn}/appointments/{appointmentId}/reschedule/context/{contextName}"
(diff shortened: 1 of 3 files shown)


src/main/resources/application.yml
```diff
@@ -113,6 +113,7 @@ community-api:
     ended-referral: "/secure/offenders/crn/{crn}/referral/end/context/{contextName}"
     book-appointment: "/secure/offenders/crn/{crn}/sentence/{sentenceId}/appointments/context/{contextName}"
     reschedule-appointment: "/secure/offenders/crn/{crn}/appointments/{appointmentId}/reschedule/context/{contextName}"
+    relocate-appointment: "/secure/offenders/crn/{crn}/appointments/{appointmentId}/relocate/context/{contextName}"
```
mahalrmoj, line 116: A context is not required in this situation as there is no mapping required that is specific to Interventions (aka CRS aka R&M)

andrewburrowsmoj: Thanks, fixed.

PR #492, opened 2021-08-09
IC-2160 add approved action plan ids  to sent referral dto

rgforsyth:

Adds all approved action plan id's to the sent-referral dto.
Added a new field `currentActionPlanId` to eventually replace `actionPlanId`. This has the same value, just a more specific name. The old field is left temporarily to avoid issues.

To provide the front end with the id's of all approved action plan's for a given referral.


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/dto/SentReferralDTO.kt
```diff
@@ -12,6 +12,8 @@ class SentReferralDTO(
   val assignedTo: AuthUserDTO?,
   val referral: DraftReferralDTO,
   val actionPlanId: UUID?,
+  val currentActionPlanId: UUID?,
```
rgforsyth, line 15: Essentially a re-name of `actionPlanId`. Leaving the old field here temporarily until the front end has been updated to `currentActionPlanId`

tomsmyers: just a heads up - we should never remove the old field, to avoid potential backwards incompatibility issues.

tomsmyers (approved):
looks good - it would be good to include rationale for adding these new fields in the commit message.

PR #497, opened 2021-08-10
IC-1275 Notification to Delius of End of Service report is now synchronous

mahalrmoj:
Notification to Delius of End of Service report is now synchronous.

Prior to this PR, the notification was initiated before Referral End but was sent asynchronously. However, to avoid simultaneous object modification issues in Delius it must complete before referral end is sent. This was always the intention


tomsmyers (approved):
pasing `Unit` to the client post method looks a bit weird, but as long as it works it's all good

sldblog:
🙋‍♂️ It's probably unrealistic to shut off notifications on a person's record when an activity ends
The model doesn't seem to be flexible enough for our needs

PR #501, opened 2021-08-11
IC-2171 add endpoint to return approved action plans for a referral

rgforsyth:

New endpoint that will return all approved action plan summaries for a referral.

To provide the front end with a list of approved action plan's for a referral that can be displayed.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/jpa/repository/ActionPlanRepository.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/jpa/repository/ActionPlanRepository.kt
@@ -7,7 +7,7 @@ import java.util.UUID
 
 interface ActionPlanRepository : JpaRepository<ActionPlan, UUID> {
   fun findByIdAndSubmittedAtIsNull(id: UUID): ActionPlan?
-  fun findByReferralId(referralId: UUID): ActionPlan?
+  fun findByReferralId(referralId: UUID): List<ActionPlan>?
 
   @Query(
     "select count(sesh) from ActionPlanSession sesh join sesh.appointments appt " +
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/jpa/repository/ActionPlanRepository.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/jpa/repository/ActionPlanRepository.kt
@@ -7,7 +7,7 @@ import java.util.UUID
 
 interface ActionPlanRepository : JpaRepository<ActionPlan, UUID> {
   fun findByIdAndSubmittedAtIsNull(id: UUID): ActionPlan?
-  fun findAllByReferralIdAndApprovedAtIsNotNull(referralId: UUID): List<ActionPlan>?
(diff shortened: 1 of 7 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/ActionPlanService.kt
```diff
@@ -100,9 +100,9 @@ class ActionPlanService(
       ?: throw EntityNotFoundException("action plan not found [id=$id]")
   }
 
-  fun getActionPlanByReferral(referralId: UUID): ActionPlan {
+  fun getActionPlanByReferral(referralId: UUID): List<ActionPlan> {
```
rgforsyth, line 103: Old method that was no longer being used, so updated to work with multiple action plans.

tomsmyers (dismissed):
looks good - however i think this would be better implemented with a simple repository method then you don't have to have logic in the DTO constructor.

`actionPlanRepository.findAllByReferralIdAndApprovedAtIsNotNull(referralId: UUID)`

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/ActionPlanService.kt
tomsmyers: or if we use the suggested repository method it would be `getApprovedActionPlansByReferral`

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/jpa/repository/ActionPlanRepository.kt
```diff
@@ -7,7 +7,7 @@ import java.util.UUID
 
 interface ActionPlanRepository : JpaRepository<ActionPlan, UUID> {
   fun findByIdAndSubmittedAtIsNull(id: UUID): ActionPlan?
-  fun findByReferralId(referralId: UUID): ActionPlan?
+  fun findAllByReferralIdAndApprovedAtIsNotNull(referralId: UUID): List<ActionPlan>?
```
tomsmyers, line 10: nit: does this ever return null or always empty list?

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/ActionPlanService.kt
```diff
@@ -104,9 +104,9 @@ class ActionPlanService(
       ?: throw EntityNotFoundException("action plan not found [id=$id]")
   }
 
-  fun getActionPlanByReferral(referralId: UUID): ActionPlan {
-    return actionPlanRepository.findByReferralId(referralId)
-      ?: throw EntityNotFoundException("action plan not found [referralId=$referralId]")
+  fun getApprovedActionPlansByReferral(referralId: UUID): List<ActionPlan> {
+    return actionPlanRepository.findAllByReferralIdAndApprovedAtIsNotNull(referralId)
+      ?: throw EntityNotFoundException("no action plans found for referral [referralId=$referralId]")
```
tomsmyers, line 109: see other comment above - i would probably expect this to just return an empty list in this case

tomsmyers (approved):

PR #502, opened 2021-08-11
IC-1116 tweak the calculation for 'number of attended' sessions in SP performance report

tomsmyers:

- we simplify the code by taking only sessions from the most recently approved AP.
  this is ok, since all the previous sessions/appointments must be transfered to
  the latest approved AP when it's approved.
- the calculation is moved to ActionPlanService and more unit tests are included.
- the count of number of attended sessions is only included if there is an approved
  AP. otherwise it is left out. previously it defaulted to 0.

primarily - it's to stop the count of attended sessions defaulting to 0, but with some improvements along the way.


mahalrmoj (approved):
Nicely simplified - LGTM

PR #507, opened 2021-08-12
enable s3 in all environments

tomsmyers:

now that we have provisioned the buckets for all these envs we can turn support on in the application.

allow SP reports to be written for all envs.


mahalrmoj (approved):
LGTM

PR #511, opened 2021-08-13
use a hibernate specific reader in the SP reporting job to ensure there is a hibernate session active

tomsmyers:

changes the sp performance report reader to use a hibernate specific cursor reader which fixes the lazy loading exception in the item processor.


mahalrmoj (approved):
LGTM

PR #516, opened 2021-08-17
IC-1833 approve new action plan if there is an existing approved one

rgforsyth:

Allows the approval of a new draft action plan for a referral when it already has an approved one. This includes transferring all the action plan sessions from the previously approved action plan to the new one, using the existing approval endpoint.

To allow the approval of a new draft action plan when there has previously been on approved.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/ActionPlanService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/ActionPlanService.kt
@@ -37,7 +37,6 @@ class ActionPlanService(
     activities: List<ActionPlanActivity>,
     createdByUser: AuthUser
   ): ActionPlan {
-
     val draftActionPlan = ActionPlan(
       id = randomUUID(),
       numberOfSessions = numberOfSessions,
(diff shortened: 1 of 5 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/ActionPlanService.kt
```diff
@@ -90,15 +89,30 @@ class ActionPlanService(
...
     actionPlan.approvedAt = OffsetDateTime.now()
     actionPlan.approvedBy = authUserRepository.save(user)
 
-    actionPlanSessionsService.createUnscheduledSessionsForActionPlan(actionPlan)
     val approvedActionPlan = actionPlanRepository.save(actionPlan)
     actionPlanEventPublisher.actionPlanApprovedEvent(approvedActionPlan)
     return approvedActionPlan
   }
 
+  private fun reassignSessions(fromActionPlanId: UUID, toActionPlan: ActionPlan) {
```
andrewburrowsmoj, line 107: I am wondering if it would be better to clone rather than move it across?
This is because if they wish to view the previous action plan there will be no session tied to it?

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/ActionPlanService.kt
```diff
@@ -90,15 +89,30 @@ class ActionPlanService(
...
+    val sessions = actionPlanSessionRepository.findAllByActionPlanId(fromActionPlanId)
+    sessions.forEach {
+      it.actionPlan = toActionPlan
+      actionPlanSessionRepository.save(it)
+    }
+    actionPlanSessionsService.createUnscheduledSessionsForActionPlan(toActionPlan, sessions.count())
```
andrewburrowsmoj, line 113: Could it not be the case that there could already be some unscheduled sessions already after copying them over?

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/ActionPlanService.kt
sldblog: +1, would be better to not mutate data, that can get hairy over time

We have a few options

- copying/cloning
- removing the action plan ID -- delivery_session link, and use referral -- delivery_session link, so that same sessions are used across _all_ actions plans

sldblog:
📝 From standup: this PR can probably be closed

PR #521, opened 2021-08-18
IC-2170 set dummy values in performance report to null

rgforsyth:

Remove dummy values from performance reporting.

The 'dummy' values are causes issues currently, not need to be replace by null/empty values.


andrewburrowsmoj (approved):

sldblog (commented):
Just to make this visible, the ticket as-is may not solve the users' needs -- I'm following up

sldblog:
📝 My concern was that we won't be able to distinguish between genuinely missing entries and placeholder entries, but this will be handled by communication.

sldblog (approved):
LGTM

📝 Checked the output, it will show up as empty string in CSVs

🙋‍♂️ We could now fill the first two values with actual values

PR #529, opened 2021-08-20
Fix performance report job

andrewburrowsmoj:
Replaced user.id parameter in performanceReportReader Job with authUser. This variable now holds the AuthUser entity.

The reason for this change is because the job was failing for some users because they didn't have any user details persisted into the database.
The job was trying to retrieve the user details from the db based on the user id passed in the job.

Since the contracts are held on the authUser passed in via Auth we don't need to persist and retrieve it from the database.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/ReportingService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/ReportingService.kt
@@ -28,6 +28,8 @@ class ReportingService(
       performanceReportJob,
       JobParametersBuilder()
         .addString("user.id", user.id)
+        .addString("user.authSource", user.authSource)
+        .addString("user.userName", user.userName)
         .addString("user.firstName", userDetail.firstName)
         .addString("user.email", userDetail.email)
         .addDate("from", batchUtils.parseLocalDateToDate(from))
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/reporting/serviceprovider/performance/PerformanceReportJobConfiguration.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/reporting/serviceprovider/performance/PerformanceReportJobConfiguration.kt
@@ -17,7 +17,7 @@ import org.springframework.context.annotation.Bean
 import org.springframework.context.annotation.Configuration
 import org.springframework.core.io.FileSystemResource
 import uk.gov.justice.digital.hmpps.hmppsinterventionsservice.authorization.ServiceProviderAccessScopeMapper
-import uk.gov.justice.digital.hmpps.hmppsinterventionsservice.authorization.UserMapper
+import uk.gov.justice.digital.hmpps.hmppsinterventionsservice.jpa.entity.AuthUser
 import uk.gov.justice.digital.hmpps.hmppsinterventionsservice.jpa.entity.Referral
 import uk.gov.justice.digital.hmpps.hmppsinterventionsservice.reporting.BatchUtils
 import java.util.Date
@@ -25,7 +25,6 @@ import java.util.Date
 @Configuration
 @EnableBatchProcessing
 class PerformanceReportJobConfiguration(
-  private val userMapper: UserMapper,
   private val serviceProviderAccessScopeMapper: ServiceProviderAccessScopeMapper,
   private val jobBuilderFactory: JobBuilderFactory,
   private val stepBuilderFactory: StepBuilderFactory,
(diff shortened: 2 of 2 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/ReportingService.kt
```diff
@@ -27,12 +28,12 @@ class ReportingService(
     asyncJobLauncher.run(
       performanceReportJob,
       JobParametersBuilder()
-        .addString("user.id", user.id)
         .addString("user.firstName", userDetail.firstName)
         .addString("user.email", userDetail.email)
         .addDate("from", batchUtils.parseLocalDateToDate(from))
         .addDate("to", batchUtils.parseLocalDateToDate(to.plusDays(1))) // 'to' is inclusive
```
sldblog, line 34: > 'to' is inclusive

_nit_ renaming the argument to `toIncluding` would make this comment unnecessary and verifiable

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/reporting/serviceprovider/performance/AuthUserJobParameter.kt
```diff
@@ -0,0 +1,12 @@
+package uk.gov.justice.digital.hmpps.hmppsinterventionsservice.reporting.serviceprovider.performance
+
+import org.springframework.batch.core.JobParameter
+import uk.gov.justice.digital.hmpps.hmppsinterventionsservice.jpa.entity.AuthUser
+
+data class AuthUserJobParameter(
+  val authUser: AuthUser
+) : JobParameter("") {
```
sldblog, line 8: Does this work?

The [documentation](https://www.docs4dev.com/javadoc/en/org/springframework/batch/spring-batch-core/4.2.4.RELEASE/org/springframework/batch/core/JobParameter.html) says

> Only the following types can be parameters: String, Long, Date, and Double

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/ReportingService.kt
```diff
@@ -27,12 +28,12 @@ class ReportingService(
...
         .addString("timestamp", Instant.now().toEpochMilli().toString())
+        .addParameter("authUser", AuthUserJobParameter(user))
```
sldblog, line 36: _nit_ There's

```
    // this is really not ideal - but the intention here is validate the user has
    // a valid access scope before launching the job (which will also call this method).
    serviceProviderAccessScopeMapper.fromUser(user)
```
above

What if we renamed it to `serviceProviderAccessScopeMapper.validatedScopeFrom(user)`? It'd make the comment unnecessary, as it'd be obvious from the code that it validates access

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/reporting/serviceprovider/performance/AuthUserJobParameter.kt
sldblog: Job arguments are likely serialised in https://hmpps-interventions-service-dev.apps.live-1.cloud-platform.service.justice.gov.uk/meta/schema/tables/batch_job_execution_params.html, which would make this difficult to work

sldblog (commented):
(reflecting on the comments and how to make those observations apparent from the code itself)

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/reporting/serviceprovider/performance/AuthUserJobParameter.kt
andrewburrowsmoj: Passed in user parameters to construct authUSer instead

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/ReportingService.kt
andrewburrowsmoj: I think `to` being inclusive would be the expected behaviour - I don't think there would be confustion on the variable - I think it's mainly comment to explain why a day has been added (because converting from localDate to date you need to add 24 hours)

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/ReportingService.kt
andrewburrowsmoj: Yeah, I agree that the `fromUser` method doesn't really explain what is happening. I think renaming the method to `obtainAccessScope(user)` and the class `ServiceProviderAccessAuthorizator`?

sldblog (approved):
LGTM
Does this need any integration test changes?

andrewburrowsmoj:
No, there wasn't much testing around this. I am going to confirm in Dev

PR #530, opened 2021-08-20
Do not upgrade awssdk with patch level changes

sldblog:

Do not upgrade awssdk with patch level changes

All of the AWS SDK libraries get updated at the same time, and most of
the time the updates are not relevant

Any relevant updates will be posted via minor version changes


andrewburrowsmoj (approved):
Nice change

PR #535, opened 2021-08-23
IC-1116 use the caseworker's username instead of id to populate the 'caseworker_id' field in the SP performance report

tomsmyers:
# What does this pull request do?

IC-1116 use the caseworker's username instead of id to populate the 'caseworker_id' field in the SP performance report

the reason for this is that the report reader expects this field to be the caseworker's email address.
whilst we can't guarantee that this field contains the email address, it is very likely.


andrewburrowsmoj (approved):

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/reporting/serviceprovider/performance/PerformanceReportProcessor.kt
```diff
@@ -28,7 +28,7 @@ class PerformanceReportProcessor(
       referralId = referral.id,
       contractReference = contract.contractReference,
       organisationId = contract.primeProvider.id,
-      currentAssigneeId = referral.currentAssignee?.id,
+      currentAssigneeId = referral.currentAssignee?.userName,
```
andrewburrowsmoj, line 31: Maybe rename the var too?

tomsmyers: done

sldblog:
>  whilst we can't guarantee that this field contains the email address, it is very likely.

Can we do an assumption assertion on it?

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/reporting/serviceprovider/performance/PerformanceReportProcessor.kt
```diff
@@ -28,7 +28,7 @@ class PerformanceReportProcessor(
       referralId = referral.id,
       contractReference = contract.contractReference,
       organisationId = contract.primeProvider.id,
-      currentAssigneeId = referral.currentAssignee?.id,
+      currentAssigneeEmail = referral.currentAssignee?.userName,
```
sldblog, line 31: should we use 

https://github.com/ministryofjustice/hmpps-interventions-service/blob/d599e6f0f6b892e3e5a237e5ddc208111f4190f1/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/HMPPSAuthService.kt#L99

here?

PR #538, opened 2021-08-24
Ic 1116 improve logs in batch processing

tomsmyers:

0510e6fc add a log message which details the performance report parameters in the reader
a17d844f reduce log level from INFO to DEBUG for some low level report processing messages

reduce polluting logs in prod. improve debugability of the sp performance report job.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/reporting/serviceprovider/performance/PerformanceReportJobConfiguration.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/reporting/serviceprovider/performance/PerformanceReportJobConfiguration.kt
@@ -1,5 +1,7 @@
 package uk.gov.justice.digital.hmpps.hmppsinterventionsservice.reporting.serviceprovider.performance
 
+import mu.KLogging
+import net.logstash.logback.argument.StructuredArguments.kv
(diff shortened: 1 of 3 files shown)


sldblog (approved):

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/reporting/serviceprovider/performance/PerformanceReportJobConfiguration.kt
```diff
@@ -45,6 +49,15 @@ class PerformanceReportJobConfiguration(
   ): HibernateCursorItemReader<Referral> {
     // this reader returns referral entities which need processing for the report.
     val contracts = serviceProviderAccessScopeMapper.fromUser(AuthUser(userId, authSource, userName)).contracts
+
+    logger.info(
```
tomsmyers, line 53: i really wish we could pass a list of contract refs to the job. then we could just use spring batch's logging to tell us the job params.

tomsmyers: we could just comma seperate them. but as @andrewburrowsmoj has mentioned before, we have a 250 char limit. could there ever be that many contracts for one user that it would overflow it??

PR #539, opened 2021-08-24
IC-1116 Improve the way contracts references are passed to the SP performance report batch job.

tomsmyers:

Instead of passing the auth user fields to the batch job, we construct a comma seperated
list of the contracts. This has a few benefits:
 1) We do not have to load the user access scopes twice (again in the step reader)
 2) The job parameters are more reflective of the actual work being done.
 3) The contracts are logged automatically by the batch framework when the job is launched.
 
 e.g. 
 ```
 2021-08-24 12:57:02.730  INFO 65856 --- [lTaskExecutor-1] o.s.b.c.l.support.SimpleJobLauncher      : Job: [SimpleJob: [name=performanceReportJob]] launched with the following parameters: [{contractReferences=0001,0002,0003,0004,0005,0006, user.id=6c4036b7-e87d-44fb-864f-5a06c1c492f3, user.firstName=Robin, user.email=test.interventions.sp.1@digital.justice.gov.uk, from=1609459200000, to=1640995200000, timestamp=1629806222643}]
```

The maximum length of string job params is 250 chars. This limits us to users with approx 27 contracts associated with their account. This is very unlikely to ever happen. If it does we can increase the size of the string param in the sql migration scripts.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/reporting/serviceprovider/performance/PerformanceReportJobConfiguration.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/reporting/serviceprovider/performance/PerformanceReportJobConfiguration.kt
@@ -1,7 +1,5 @@
 package uk.gov.justice.digital.hmpps.hmppsinterventionsservice.reporting.serviceprovider.performance
 
-import mu.KLogging
-import net.logstash.logback.argument.StructuredArguments.kv
 import org.hibernate.SessionFactory
 import org.springframework.batch.core.Job
 import org.springframework.batch.core.Step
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Review by andrewburrowsmoj — APPROVED** [review]
I think your assumptions are reasonable based on Sarah's reply.

**Comment by sldblog** [comment]
_observation_ There's no real benefit of running a `varchar(250)` instead of a `text`, apart from semantics (both store the length in the beginning)

If our semantic is to "potentially any length" then we should migrate to `text`

> This limits us to users with approx 27 contracts associated with their account. This is very unlikely to ever happen

True, although we could also run into problems without the number of contracts increasing:

```
.addString("contractReferences", contracts.joinToString(",") { it.contractReference })
```

- an across-the board change to use IDs instead of contract references may change this as part of a refactor, which would increase the size but no tests would pick it up
- we may rework how long references are in the future for some reason (contract management consolidation?), with the same effects

(Essentially, innocent changes, no test failures, but runtime might still fail for certain users)

So it seems it's safer to upgrade the field now
~ or make it in a way that each contract gets queued in its own "step" and merge the results at the end

**Comment by sldblog** [comment]
cc @tomsmyers @andrewburrowsmoj

**Comment by sldblog** [comment]
_observation_ I'm pondering about the language of the feature... The ticket says

> give all service providers access to the basic information that's being used to measure their performance
> {snip} As a SERVICE PROVIDER MANAGER {snip}

We are a little disconnected this; we don't have a provider manager role, therefore we rely on the contracts of a certain user to make sure the scope is not breached. In other words, if we followed the domain-language, this job could be configured with

```
.addString("providerId", providerId)
```

Just wondering if this difference is OK?

PR #547, opened 2021-08-31
Ic 1397/provide ability to obtain list of case notes

andrewburrowsmoj:

Provided ability to request for a list of paginated case notes.

The request /sent-referral/{referralId}/case-notes can accept a Pageable request parameters. Pageable is a defined spring interface where each value can be nullable. If none are provided then the default values are used - in this case page-size=10, page-number=0.

PageableHandlerMethodArgumentResolver bean had to be added in order to accept pageable request parameters.

Ensure that we have a way to obtain a list of case notes for a referral with pagination to reduce stress on the server.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/jpa/repository/CaseNoteRepository.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/jpa/repository/CaseNoteRepository.kt
@@ -1,7 +1,11 @@
 package uk.gov.justice.digital.hmpps.hmppsinterventionsservice.jpa.repository
 
-import org.springframework.data.jpa.repository.JpaRepository
+import org.springframework.data.domain.Page
+import org.springframework.data.domain.Pageable
+import org.springframework.data.repository.PagingAndSortingRepository
 import uk.gov.justice.digital.hmpps.hmppsinterventionsservice.jpa.entity.CaseNote
 import java.util.UUID
 
-interface CaseNoteRepository : JpaRepository<CaseNote, UUID>
+interface CaseNoteRepository : PagingAndSortingRepository<CaseNote, UUID> {
+  fun findAllByReferralId(referralId: UUID, pageable: Pageable?): Page<CaseNote>
+}
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/CaseNoteService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/CaseNoteService.kt
@@ -1,5 +1,7 @@
 package uk.gov.justice.digital.hmpps.hmppsinterventionsservice.service
 
+import org.springframework.data.domain.Page
+import org.springframework.data.domain.Pageable
 import org.springframework.stereotype.Service
 import uk.gov.justice.digital.hmpps.hmppsinterventionsservice.jpa.entity.AuthUser
 import uk.gov.justice.digital.hmpps.hmppsinterventionsservice.jpa.entity.CaseNote
@@ -28,5 +30,9 @@ class CaseNoteService(
       sentAt = OffsetDateTime.now(),
     )
     return caseNoteRepository.save(caseNote)
+  }
+
+  fun findByReferral(referralId: UUID, pageable: Pageable?): Page<CaseNote> {
+    return caseNoteRepository.findAllByReferralId(referralId, pageable)
   }
 }
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/config/ResourceServerConfiguration.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/config/ResourceServerConfiguration.kt
@@ -5,6 +5,7 @@ import org.springframework.boot.autoconfigure.security.oauth2.resource.OAuth2Res
 import org.springframework.context.annotation.Bean
 import org.springframework.context.annotation.Configuration
 import org.springframework.context.annotation.Profile
+import org.springframework.data.web.PageableHandlerMethodArgumentResolver
 import org.springframework.security.config.annotation.web.builders.HttpSecurity
 import org.springframework.security.config.annotation.web.configuration.EnableWebSecurity
 import org.springframework.security.config.annotation.web.configuration.WebSecurityConfigurerAdapter
@@ -71,6 +72,11 @@ class ResourceServerConfiguration(private val tokenVerifier: TokenVerifier) : We
   @Profile("test")
   fun testJwtDecoder(): JwtDecoder {
     return TestJwtDecoder()
+  }
+
+  @Bean
+  fun pageableResolver(): PageableHandlerMethodArgumentResolver? {
+    return PageableHandlerMethodArgumentResolver()
   }
 }
 
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/controller/CaseNoteController.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/controller/CaseNoteController.kt
@@ -1,7 +1,12 @@
 package uk.gov.justice.digital.hmpps.hmppsinterventionsservice.controller
 
+import org.springframework.data.domain.Page
+import org.springframework.data.domain.Pageable
+import org.springframework.data.web.PageableDefault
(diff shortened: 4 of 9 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/CaseNoteService.kt
```diff
@@ -29,4 +32,9 @@ class CaseNoteService(
     )
     return caseNoteRepository.save(caseNote)
   }
+
+  fun findByReferral(referralId: UUID, pageNumber: Int, pageSize: Int): Page<CaseNote> {
+    var pageRequest = PageRequest.of(pageNumber, pageSize, Sort.by("sentAt").ascending())
+    return caseNoteRepository.findAllByReferralId(referralId, pageRequest)
```
mahalrmoj, line 38: Should this not be ordered by some field as without SQL doesn't guarantee in which order they are returned. The definition of a page row should be determined by some sort of line number shouldn't it.

andrewburrowsmoj: sorry, I don't understand the question? It is ordered by SentAt, this value is never null, it never changes, will be unique and new case notes would be appended to the end of this ordering.

andrewburrowsmoj: re-read your question!

So the sentAt time will always be different since a user can only create one case note at a time.

mahalrmoj (approved):

src/main/resources/application.yml
```diff
@@ -174,3 +174,7 @@ webclient:
     max-retry-attempts: 2
     connect-timeout-seconds: 3
     read-timeout-seconds: 3
+
+pagination:
+  max-size:
+    casenotes: 10
```
tomsmyers, line 180: what's the reason for enforcing this?

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/validator/PageRequestValidator.kt
```diff
@@ -0,0 +1,25 @@
...
+import org.springframework.stereotype.Component
+import uk.gov.justice.digital.hmpps.hmppsinterventionsservice.config.Code
+import uk.gov.justice.digital.hmpps.hmppsinterventionsservice.config.FieldError
+import uk.gov.justice.digital.hmpps.hmppsinterventionsservice.config.ValidationError
+
+@Component
+class PageRequestValidator {
+  fun validatePageRequest(pageNumber: Int, pageSize: Int, maxAllowablePage: Int) {
+    val errors = mutableListOf<FieldError>()
+    if (pageNumber < 0) {
```
tomsmyers, line 12: nit: do we need CANNOT_BE_NEGATIVE _and_ CANNOT_BE_NEGATIVE_OR_ZERO ?

tomsmyers (commented):
i think we can remove loads of complexity from this by just using the builtin spring pageable interface at controller level.

e.g. this is basically a complete example
```
  @GetMapping("/sent-referral/{referralId}/case-notes")
  fun getCaseNotesForReferral(@NotNull pageable: Pageable, @PathVariable referralId: UUID): Page<CaseNoteDTO> {
    return caseNoteRepository.findAllByReferralId(referralId, pageable).map { CaseNoteDTO.from(it) }
  }
  ```
  
it takes request params like `page`, `size`, `sort` 

e.g. `{{interventions-service-url}}/sent-referral/a2a551aa-3d11-44b1-907b-42a028852bc1/case-notes?page=0&size=5&sort=sentAt`

returns 

```
{
    "content": [
        {
            "id": "f18d7553-1765-487e-afde-16c10b339720",
            "referralId": "a2a551aa-3d11-44b1-907b-42a028852bc1",
            "subject": "note 1",
            "body": "some note",
            "sentBy": {
                "username": "TEST_INTERVENTIONS_SP_1",
                "authSource": "auth",
                "userId": "6c4036b7-e87d-44fb-864f-5a06c1c492f3"
            },
            "sentAt": "2021-08-31T17:46:50.592196+01:00"
        },
        {
            "id": "a9b0842c-b25e-4287-88a9-9cf3d309c9d6",
            "referralId": "a2a551aa-3d11-44b1-907b-42a028852bc1",
            "subject": "note 2",
            "body": "some note",
            "sentBy": {
                "username": "TEST_INTERVENTIONS_SP_1",
                "authSource": "auth",
                "userId": "6c4036b7-e87d-44fb-864f-5a06c1c492f3"
            },
            "sentAt": "2021-08-31T17:47:03.662295+01:00"
        },
        {
            "id": "f547eaa8-2f48-4d29-a578-214a2041a8cf",
            "referralId": "a2a551aa-3d11-44b1-907b-42a028852bc1",
            "subject": "note 3",
            "body": "some note",
            "sentBy": {
                "username": "TEST_INTERVENTIONS_SP_1",
                "authSource": "auth",
                "userId": "6c4036b7-e87d-44fb-864f-5a06c1c492f3"
            },
            "sentAt": "2021-08-31T17:47:10.308857+01:00"
        },
        {
            "id": "bd19fa43-f1ca-469e-8abb-56558f1d1be6",
            "referralId": "a2a551aa-3d11-44b1-907b-42a028852bc1",
            "subject": "note 4",
            "body": "some note",
            "sentBy": {
                "username": "TEST_INTERVENTIONS_SP_1",
                "authSource": "auth",
                "userId": "6c4036b7-e87d-44fb-864f-5a06c1c492f3"
            },
            "sentAt": "2021-08-31T17:47:15.955169+01:00"
        }
    ],
    "pageable": {
        "sort": {
            "sorted": true,
            "unsorted": false,
            "empty": false
        },
        "offset": 0,
        "pageNumber": 0,
        "pageSize": 5,
        "paged": true,
        "unpaged": false
    },
    "last": true,
    "totalPages": 1,
    "totalElements": 4,
    "size": 5,
    "number": 0,
    "sort": {
        "sorted": true,
        "unsorted": false,
        "empty": false
    },
    "numberOfElements": 4,
    "first": true,
    "empty": false
}
```

File: src/main/resources/application.yml (hunk 2 above)
Comment (andrewburrowsmoj) [reply src/main/resources/application.yml:180]: I added it as a separate commit; if you think it's unnecessary then I can drop the commit.

Currently, for flexibility, the caller can make a pagination request specifying the page number and the page size of the resource they want.

I believe that the interventions-service should be responsible for enforcing a max-size as a protection against pagination causing performance issues.

File: src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/validator/PageRequestValidator.kt (hunk 3 above)
Comment (andrewburrowsmoj) [reply src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/validator/PageRequestValidator.kt:12]: well for `CANNOT_BE_NEGATIVE`, if the value is `0` and it's not allowed then it would be confusing to the caller.

Similarly for `CANNOT_BE_NEGATIVE_OR_ZERO`.

The only alternative is to provide `INVALID_VALUE` instead.

File: src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/jpa/repository/CaseNoteRepository.kt [hunk 4]
```diff
@@ -1,7 +1,11 @@
...
-import org.springframework.data.jpa.repository.JpaRepository
+import org.springframework.data.domain.Page
+import org.springframework.data.domain.Pageable
+import org.springframework.data.repository.PagingAndSortingRepository
 import uk.gov.justice.digital.hmpps.hmppsinterventionsservice.jpa.entity.CaseNote
 import java.util.UUID
 
-interface CaseNoteRepository : JpaRepository<CaseNote, UUID>
+interface CaseNoteRepository : PagingAndSortingRepository<CaseNote, UUID> {
+  fun findAllByReferralId(referralId: UUID, pageable: Pageable?): Page<CaseNote>
```
tomsmyers, line 10: what happens if `pageable` is null?

tomsmyers (approved):
great work. thanks!

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/jpa/repository/CaseNoteRepository.kt
andrewburrowsmoj: if pageable is null then it returns all case notes (no pagination)

PR #549, opened 2021-09-02
Ic 1833/associate action plan sessions with referral

rgforsyth:

Moves action plan sessions to be held on the referral, rather than the action plan. Lots of test fixes as a result.
Update the url of the pp session feedback page to be consistent with the others.(referencing the referral) [Not yet applied to the UI]
This pr does not add the handling of approving new action plans. That will be followed up in a new pr.

To be able to store the action plan sessions against a referral, rather than an individual action plan. This means that sessions will not have to be duplicated or moved over to new action plans. This will help retain the history of all sessions easier going forward.


mahalrmoj (commented):
Looks good. Should we also change all references/names from ActionPlanSession to ReferralSession. Also in terms of migration, how does the rollout / switch over work.

src/main/resources/application.yml
```diff
@@ -111,7 +111,7 @@ interventions-ui:
       intervention-progress: "/probation-practitioner/referrals/{id}/progress"
       supplier-assessment: "/probation-practitioner/referrals/{id}/supplier-assessment"
       action-plan: "/probation-practitioner/referrals/{id}/action-plan"
-      session-feedback: "/probation-practitioner/action-plan/{id}/appointment/{sessionNumber}/post-session-feedback"
+      session-feedback: "/probation-practitioner/referrals/{id}/appointment/{sessionNumber}/post-session-feedback"
```
rgforsyth, line 114: Changed as requested to make url consistent with others. Is this url correct?

tomsmyers:
> Looks good. Should we also change all references/names from ActionPlanSession to ReferralSession. Also in terms of migration, how does the rollout / switch over work.

I think `DeliverySession` would be more appropriate ?

tomsmyers:
please can we squash out the `WIP` commit at some point too?

tomsmyers:
with regards to the roll out we have to do our usual 'seperate schema migrations from data migrations'. i guess we have to stage it like this?

1. add the new field to the session. and have the service start populating it alongside the action plan id in the codebase.
2. update all existing records to associate the referral id with the sessions. and stop using the action plan ids in the codebase. mark action_plan_id field in the sessions as deprecated.

src/main/resources/application.yml
tomsmyers: sure - works for me. as long as the places where this is used have been updated to use the referral id instead of action plan id! 

has this route been added to the UI?

mahalrmoj:
> with regards to the roll out we have to do our usual 'seperate schema migrations from data migrations'. i guess we have to stage it like this?
> 
> 1. add the new field to the session. and have the service start populating it alongside the action plan id in the codebase.
> 2. update all existing records to associate the referral id with the sessions. and stop using the action plan ids in the codebase. mark action_plan_id field in the sessions as deprecated.

That's sounds good to me.

mahalrmoj (commented):
1. Are we missing an alter table to make referral_id not null in action plan session? 
2. Does the deprecated column need to be made nullable?
3. Are we not correcting the crs_performance query as that will become less accurate over time?

src/main/resources/db/migration/V1_77_1__add_referral_data_to_action_plan_session.sql
```diff
@@ -0,0 +1,4 @@
+UPDATE action_plan_session aps
+SET referral_id = ap.referral_id
+FROM action_plan ap
+WHERE aps.action_plan_id = ap.id;
```
tomsmyers, line 4: we can add `where referral_id is null` to this too

src/main/resources/db/migration/V1_77_2__associate_action_plan_sessions_with_referrals_schema.sql
```diff
@@ -0,0 +1,2 @@
+alter table action_plan_session
```
tomsmyers, line 1: please can we update the data dictionary definitions too?

helm_deploy/hmpps-interventions-service/reports/crs_performance_report.sql
```diff
@@ -1,19 +1,19 @@
 COPY (
   WITH attended_sessions AS (
```
tomsmyers, line 2: we need to update these queries to not use the deprecated column

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/controller/ActionPlanSessionController.kt
```diff
@@ -48,21 +50,42 @@ class ActionPlanSessionController(
...
     @PathVariable(name = "id") actionPlanId: UUID
   ): List<ActionPlanSessionDTO> {
 
-    val actionPlanSessions = actionPlanSessionsService.getSessions(actionPlanId)
+    val actionPlan = actionPlanService.getActionPlan(actionPlanId)
+    val actionPlanSessions = actionPlanSessionsService.getSessions(actionPlan.referral.id)
+    return ActionPlanSessionDTO.from(actionPlanSessions)
+  }
+
+  @GetMapping("/referral/{id}/sessions")
```
tomsmyers, line 62: just noticed this has a slight change compared to the previous url

/appointments vs /sessions

is that intentional?

rgforsyth: This was something you added I think. I'm not sure it's actually being used at the moment? I can update it to appointments if you want?

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/controller/ActionPlanSessionController.kt
```diff
@@ -48,21 +50,42 @@ class ActionPlanSessionController(
...
     @PathVariable sessionNumber: Int,
   ): ActionPlanSessionDTO {
 
-    val actionPlanSession = actionPlanSessionsService.getSession(actionPlanId, sessionNumber)
+    val actionPlan = actionPlanService.getActionPlan(actionPlanId)
+    val actionPlanSession = actionPlanSessionsService.getSession(actionPlan.referral.id, sessionNumber)
+    return ActionPlanSessionDTO.from(actionPlanSession)
+  }
+
+  @GetMapping("/referral/{id}/sessions/{sessionNumber}",)
```
tomsmyers, line 82: same question as above here

rgforsyth: as above

PR #550, opened 2021-09-03
IC-707 Add GET /case-note/{id} endpoint to get an individual case note

tomsmyers:

Add GET /case-note/{id} endpoint to get an individual case note

allow a single case not to be retrieved


andrewburrowsmoj (approved):
lgtm

PR #551, opened 2021-09-03
Make pagination requests non zero indexed

andrewburrowsmoj:

Removes an unnecessary bean.

Makes pageable requests to start from 1 instead of 0.

Makes it easier for caller as they don't have to decrement the page number; naturally page numbers will start from 1 rather than 0 from the UI.


tomsmyers (approved):
nice

PR #555, opened 2021-09-06
IC-1833 submitting new action plan - ON HOLD

rgforsyth:

Allows a new action plan to be approved, which will add any extra sessions if there are more than the old approved action plan.

To allow a multiple action plans to be approved against a referral.


sldblog:
No longer relevant -- closing

PR #556, opened 2021-09-06
IC-339 send email notifications on new case note creation

tomsmyers:

sends emails on new case notes being sent

alert sp or pp to new case in interventions.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/jpa/entity/AuthUser.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/jpa/entity/AuthUser.kt
@@ -9,4 +9,16 @@ data class AuthUser(
   @Id @NotNull val id: String,
   @NotNull val authSource: String,
   @NotNull val userName: String,
-)
+) {
+  override fun hashCode(): Int {
+    return id.hashCode()
+  }
+
+  override fun equals(other: Any?): Boolean {
+    if (other == null || other !is AuthUser) {
+      return false
+    }
+
+    return id == other.id
+  }
+}
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/ReferralService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/ReferralService.kt
@@ -36,10 +36,12 @@ import java.time.OffsetDateTime
 import java.util.UUID
 import javax.transaction.Transactional
 
-data class ContactableProbationPractitioner(
-  val firstName: String,
-  val email: String,
-)
+data class ResponsibleProbationPractitioner(
+  override val firstName: String,
+  override val email: String,
+  val staffId: Long?,
+  val authUser: AuthUser?,
+) : ContactablePerson
 
 @Service
 @Transactional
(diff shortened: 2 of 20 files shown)


andrewburrowsmoj (approved):
lgtm

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/notifications/CaseNotesNotificationsService.kt
```diff
@@ -0,0 +1,88 @@
...
+  private fun emailAssignedCaseWorker(caseNote: CaseNote) {
+    val sender = caseNote.sentBy
+    val assignee = caseNote.referral.currentAssignee
+    val senderIsAssignee = sender == assignee
+
+      // if the case note was sent by someone other than the assignee, email the assignee
+    if (!senderIsAssignee) {
+      val assigneeDetails = hmppsAuthService.getUserDetail(
+        assignee
+          ?: run {
```
tomsmyers, line 71: assignee will never be null here

reason is `sender` is not nullable. and this block is gated on `sender == assignee`

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/jpa/entity/AuthUser.kt
```diff
@@ -9,4 +9,16 @@ data class AuthUser(
...
+  override fun hashCode(): Int {
+    return id.hashCode()
+  }
+
+  override fun equals(other: Any?): Boolean {
+    if (other == null || other !is AuthUser) {
+      return false
+    }
+
+    return id == other.id
```
andrewburrowsmoj, line 22: Whilst it will never be the case that ids match across sources, It seems that it's good form to include authSource in the equality as well.

tomsmyers: since the id is the db primary key i don't think it's nessecary

tomsmyers: i did initially include it. but later removed it.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/ReferralService.kt
```diff
@@ -36,10 +36,12 @@ import java.time.OffsetDateTime
...
 import javax.transaction.Transactional
 
-data class ContactableProbationPractitioner(
-  val firstName: String,
-  val email: String,
-)
+data class ResponsibleProbationPractitioner(
+  override val firstName: String,
+  override val email: String,
+  val staffId: Long?,
```
andrewburrowsmoj, line 42: perhaps it would help a bit to rename this to `deliusStaffId` to aid in understanding where this id is from

andrewburrowsmoj (approved):
lgtm - I don't think there is a more suitable way to rewrite the logic of `senderIsResponsibleOfficer`

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/ReferralService.kt
tomsmyers: yeh that's a good idea - thanks

PR #559, opened 2021-09-07
ic-1833 initial db change to add referal to action plan session

rgforsyth:

Adds the new referral field to the db, whilst maintaining the action plan field for now.

Pre-requisite for upcoming changes to move action plan session from being on an action to a referral.


tomsmyers (approved):
LGTM.

perhaps label the migration V1_77_0, then when you do the next bit you can group it as V1_77_1 ?

PR #560, opened 2021-09-08
IC-2211 When creating/updating a draft action plan, the number of sessions cannot be reduced as compared to previously approved plans

mahalrmoj:
Validation to ensure that when creating/updating a draft action plan, the number of sessions cannot be reduced as compared to previously approved plans

This is a constraint that exists to deal with the situation when sessions may have already been delivered to the service user


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/component/ActionPlanValidator.kt
```diff
@@ -14,6 +15,10 @@ class ActionPlanValidator {
 
     validateNumberOfSessionsIsGreaterThanZero(update, errors)
 
+    if (errors.isEmpty()) {
```
tomsmyers, line 18: i understand why this check is here, but if other checks are added to this function and the order is changed we may want to remove this check. perhaps we could group the number of sessions checks together in one method?

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/jpa/entity/Referral.kt
```diff
@@ -104,4 +104,7 @@ class Referral(
 
   val currentAssignee: AuthUser?
     get() = currentAssignment?.assignedTo
+
+  val maximumNumberOfSessionsInPreviousAllApprovedPlans: Int?
```
tomsmyers, line 108: i don't think this is nessecary. 

first, we don't need to check all the action plans. the number of sessions can never go down, so we only need to check the most recently approved. 

second, the action plan service can just do `referral.approvedActionPlan.numberOfSessions`

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/component/ActionPlanValidator.kt
```diff
@@ -29,6 +34,16 @@ class ActionPlanValidator {
     }
   }
 
+  private fun validateNumberOfSessionsIsNotReduced(update: ActionPlan, errors: MutableList<FieldError>) {
```
tomsmyers, line 37: based on the comment in the referral service, i think we can simplify this to 

```  private fun validateNumberOfSessionsIsNotReduced(update: ActionPlan, errors: MutableList<FieldError>) {
    val newNumberOfSessions = update.numberOfSessions ?: return
    val existingNumberOfSessions = update.referral.approvedActionPlan?.numberOfSessions ?: return

    if (newNumberOfSessions < existingNumberOfSessions) {
      errors.add(FieldError(field = "numberOfSessions", error = CANNOT_BE_REDUCED))
    }
  }
```

tomsmyers (approved):
LGTM 

do we need to add a call to `actionPlanValidator.validateDraftActionPlanUpdate(...)` in ActionPlanService::createDraftActionPlan ?

mahalrmoj:
Yes - will do

PR #562, opened 2021-09-14
Bug fix: Move appointments from old action plan to new action plan

andrewburrowsmoj:
Added an update sql statement that moves appointments that exist on old action plan sessions to their equivalent session number on the new action plan where:

1. There is at least 1 appointment on the old action plan sessions.
2. There are no appointments on the new action plan sessions.
3. The number of sessions on the new action plan is enough to hold all the appointments of the old action plan.

This sql will solve an issue in prod whereby we activated multi action plan views in the front-end before the work was ready in the back-end. This meant that appointments were seemingly lost to the user; so in order to make them visible once again to the user we move them to the newest action plan.

Note that this sql will only effect around 9 referrals..

### DIFF
--- a/src/main/resources/db/migration/V1_78__move_appointments_to_new_action_plan_sessions.sql
+++ b/src/main/resources/db/migration/V1_78__move_appointments_to_new_action_plan_sessions.sql
@@ -1,0 +1,63 @@
+create table action_plan_session_appointment_pre_v1_78 as select * from action_plan_session_appointment;
+
+with action_plans_to_modify as (
+    with referrals_with_multiple_approved_aps as (
+        select
+            r.id referral_id
+        from
+            referral r
+        where
+            r.id in (
+                    '05b1d52a-ae5c-49a6-8b4d-080a5101e79f',
+                    'd2d72900-a5db-4711-a0ab-d48718f5af1d',
+                    'd60d1343-2e3f-44cb-93d4-3280938e7824',
+                    'da8edf26-d831-4b3d-bd00-9618ca4a9fa3',
+                    '75d3a978-fefc-4252-84d1-45fa86d27b90',
+                    '0be1d4f5-53a1-4e11-8c35-803774f67c7a'
+                    )
+    ),
+         appointments_per_action_plan as (
+             select
+                 ap.referral_id,
+                 ap.id action_plan_id,
+                 ap.approved_at,
+                 count(aps.id) session_count,
+                 coalesce(count(appointment_id),0) appointment_count
+             from
+                 action_plan ap
+                     inner join action_plan_session aps on aps.action_plan_id = ap.id
+                     left join action_plan_session_appointment apsa on apsa.action_plan_session_id = aps.id
+             where
+                 ap.approved_at is not null
+             group by
+                 ap.referral_id,
+                 ap.id,
+                 ap.approved_at
+         )
+    select distinct
+        rwm.referral_id,
+        older_action_plan.action_plan_id older_action_plan_id,
+        newer_action_plan.action_plan_id newer_action_plan_id
+    from
+        referrals_with_multiple_approved_aps rwm
+            inner join appointments_per_action_plan older_action_plan on older_action_plan.referral_id = rwm.referral_id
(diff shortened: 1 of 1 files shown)


mahalrmoj:
Before the update is done should we keep a copy of the data as a precaution, e.g.
`create table action_plan_session_appointment_pre_v1_78 as select * from action_plan_session_appointment`

andrewburrowsmoj:
ok, good idea Rajiv - regarding create a new table.

src/main/resources/db/migration/V1_78__move_appointments_to_new_action_plan_sessions.sql
```diff
@@ -0,0 +1,65 @@
...
+                     'd60d1343-2e3f-44cb-93d4-3280938e7824',
+                     'da8edf26-d831-4b3d-bd00-9618ca4a9fa3'
+                    )
+    ),
+         appointments_per_action_plan as (
+             select
+                 ap.referral_id,
+                 ap.id action_plan_id,
+                 ap.approved_at,
+                 count(aps.id) session_count,
```
sldblog, line 26: Just checking: is `ap.number_of_sessions` different?

andrewburrowsmoj: It should be the same. 
To be honest - not sure why we have that field since we can obtain it from counting the sessions on an action plan.

src/main/resources/db/migration/V1_78__move_appointments_to_new_action_plan_sessions.sql
```diff
@@ -0,0 +1,65 @@
...
+
+with action_plans_to_modify as (
+    with referrals_with_multiple_approved_aps as (
+        select
+            r.id referral_id
+        from
+            referral r
+        where
+            r.id in (
+                     '05b1d52a-ae5c-49a6-8b4d-080a5101e79f',
```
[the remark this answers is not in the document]
andrewburrowsmoj: You are right, however I see it as a safety check to prevent updating anything that we shouldn't. 

My understanding is that even if we make it repeatable for other referrals - it will only ever be run once because it's part of migration.

I originally had the first part as: 

```
        select
            ap.referral_id
        from
            referral r
                inner join action_plan ap on ap.referral_id = r.id
        where
            ap.approved_at is not null
          and (r.concluded_at is null and r.end_requested_at is null)
        group by
            ap.referral_id
        having
                count(ap.referral_id) = 2
```

would you prefer that?

src/main/resources/db/migration/V1_78__move_appointments_to_new_action_plan_sessions.sql
sldblog: Well, we only generate the sessions when the plan is approved

src/main/resources/db/migration/V1_78__move_appointments_to_new_action_plan_sessions.sql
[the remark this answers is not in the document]
sldblog: Yes, that's why I think this shouldn't be a versioned migration, but more like a job we can run on demand

sldblog (commented):
👍 Let's try this on preprod and compare

src/main/resources/db/migration/V1_78__move_appointments_to_new_action_plan_sessions.sql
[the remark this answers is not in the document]
andrewburrowsmoj: @rajivmahal1 What are you thoughts regarding hard-coded vs dynamic?

I guess it conflicts with your comment.

src/main/resources/db/migration/V1_78__move_appointments_to_new_action_plan_sessions.sql
andrewburrowsmoj: Ok - but it shouldn't matter if we use one or the other, right?

sldblog (approved):

sldblog:
Tested on preprod before/after, we are happy with the results

PR #564, opened 2021-09-14
Scheduled preprod refresh job

sldblog:

Introduces a scheduled database refresh job to replace https://github.com/ministryofjustice/hmpps-interventions-ops/blob/4803fcd01bf77748d3659860057d06d2da293409/copy_prod_to_preprod.sh

Currently coded as a standalone script which requires setting up port-forwarding, this approach can run on the Kubernetes namespace itself to be more secure and quicker

This cronjob can also be executed with `kubectl create job --from=cronjob/db-refresh-job refresh-now-job` for an ad-hoc refresh. (At the time of writing, this requires kubectl 1.19 as the server is still 1.18)

### Notes

- The approach is taken from https://github.com/ministryofjustice/offender-case-notes/tree/27ab394d717d29a3e2af80f7cec5ab7bb1835dda/database_refresh
- Preprod DB credentials were injected to prod via ministryofjustice/cloud-platform-environments#5566

### DIFF
--- a/helm_deploy/hmpps-interventions-service/templates/cronjob-prod-to-preprod-refresh.yaml
+++ b/helm_deploy/hmpps-interventions-service/templates/cronjob-prod-to-preprod-refresh.yaml
@@ -1,0 +1,94 @@
+{{- if .Values.env_details.prod_to_preprod_refresh_job -}}
+---
+apiVersion: v1
+kind: ConfigMap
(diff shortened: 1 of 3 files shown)


andrewburrowsmoj (approved):

sldblog:
🎩 🙇

PR #566, opened 2021-09-15
Bug fix: Email not always sent for case notes

andrewburrowsmoj:
To fix a bug in production:
```
This statement has been closed.
uk.gov.justice.digital.hmpps.hmppsinterventionsservice.jpa.entity.Referral in getCurrentAssignment at line 140
uk.gov.justice.digital.hmpps.hmppsinterventionsservice.jpa.entity.Referral in getCurrentAssignee at line 106
uk.gov.justice.digital.hmpps.hmppsinterventionsservice.service.CaseNotesNotificationsService in emailAssignedCaseWorker at line 66
uk.gov.justice.digital.hmpps.hmppsinterventionsservice.service.CaseNotesNotificationsService in onApplicationEvent at line 32
```
Which is caused by failing to lazy fetch the currentAssignee from the referral because the session has been closed.

To fix, we fetch the case notes again so it can be lazy loaded.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/notifications/CaseNotesNotificationsService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/notifications/CaseNotesNotificationsService.kt
@@ -22,6 +22,7 @@ class CaseNotesNotificationsService(
   private val hmppsAuthService: HMPPSAuthService,
   private val userTypeChecker: UserTypeChecker,
   private val communityAPIOffenderService: CommunityAPIOffenderService,
+  private val caseNoteService: CaseNoteService,
 ) : ApplicationListener<CaseNoteEvent>, NotifyService {
   companion object : KLogging()
 
@@ -29,8 +30,14 @@ class CaseNotesNotificationsService(
   override fun onApplicationEvent(event: CaseNoteEvent) {
     when (event.type) {
       CaseNoteEventType.SENT -> {
-        emailAssignedCaseWorker(event.caseNote)
-        emailResponsibleProbationPractitioner(event.caseNote)
+        // get case note from repository because the linking objects (referral.currentAssignee) on the caseNote from the event
+        // may sometimes not be lazy loaded because the session may already have been closed by this point
+        caseNoteService.getCaseNoteById(event.caseNote.id)?.let { caseNote ->
+          emailAssignedCaseWorker(caseNote)
+          emailResponsibleProbationPractitioner(caseNote)
+        } ?: run {
+          throw RuntimeException("Unable to retrieve case note for id ${event.caseNote.id}")
+        }
       }
     }
   }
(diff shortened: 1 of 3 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/notifications/CaseNotesNotificationsService.kt
```diff
@@ -22,15 +24,24 @@ class CaseNotesNotificationsService(
...
 
   @AsyncEventExceptionHandling
   override fun onApplicationEvent(event: CaseNoteEvent) {
     when (event.type) {
       CaseNoteEventType.SENT -> {
-        emailAssignedCaseWorker(event.caseNote)
-        emailResponsibleProbationPractitioner(event.caseNote)
+        // get case note from repository because the linking objects (referral.currentAssignee) on the caseNote from the event
+        // may sometimes not be lazy loaded because the session may already have been closed by this point
+       caseNoteRepository
```
sldblog, line 37: Can we use repositories here?

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/notifications/CaseNotesNotificationsService.kt
```diff
@@ -22,15 +24,24 @@ class CaseNotesNotificationsService(
...
+         .findById(event.caseNote.id).ifPresentOrElse({
+           emailAssignedCaseWorker(it)
+           emailResponsibleProbationPractitioner(it)
+         },{
+           // should never happen
+          logger.error("Unable to retrieve case note for id ${event.caseNote.id}")
```
sldblog, line 43: As this is an unrecoverable failure, could we use `orElseThrow` here instead? That seems semantically to be closer to the intended behaviour, plus makes it implicit that it should not happen

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/notifications/CaseNotesNotificationsService.kt
andrewburrowsmoj: should not be a problem because we're not performing any write operations - but I've changed it to a service to be more consistent with how we encapsulate Repository methods.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/notifications/CaseNotesNotificationsService.kt
andrewburrowsmoj: Yes, it would be nicer that way. I've changed to an exception.

sldblog (approved):
👍 LGTM, thank you!

sldblog:
Sentry exceptions that can be resolved:
- https://sentry.io/organizations/ministryofjustice/issues/2650103942/?project=5807819
- some of https://sentry.io/organizations/ministryofjustice/issues/2650102896/?project=5807819 (we'd need to start dissecting those from a wrapper)

PR #568, opened 2021-09-16
backfill referral_id & make referral_id required for delivery session

rgforsyth:

backfills `referral_id`'s onto the `action_plan_session` table, and makes the column unique.

1st step in the move to a delivery session table.


sldblog:
_not a blocker, just a note_

🙋‍♂️ The branch name will show up in the deployment notes, and visible for stakeholders, for example

image

Essentially, branch names are now part of our communication to the wider probation service

sldblog (approved):

PR #569, opened 2021-09-16
Bug fix: Prevent getCaseNote failing due to caseNote not being persisted at time of call

andrewburrowsmoj:
Changed CaseNoteEvent to accept a primitive values rather than entity

This is to decouple the entity from asynch event handling. The transaction boundaries are not clear for asynchronous calls; it's best to fetch a new object from the data.

In addition, added @transactional onto CaseNoteNotificationsService.

This is to circumvent an issue we had in production whereby the caseNote.referral.assignedUser was being lazy fetched when the tied entity manager is no longer available as it's outsides the scope of the old transaction session because it's asynchronous.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/events/CaseNoteEventPublisher.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/events/CaseNoteEventPublisher.kt
@@ -5,17 +5,16 @@ import org.springframework.context.ApplicationEventPublisher
 import org.springframework.stereotype.Component
 import uk.gov.justice.digital.hmpps.hmppsinterventionsservice.component.LocationMapper
 import uk.gov.justice.digital.hmpps.hmppsinterventionsservice.controller.CaseNoteController
+import uk.gov.justice.digital.hmpps.hmppsinterventionsservice.jpa.entity.AuthUser
 import uk.gov.justice.digital.hmpps.hmppsinterventionsservice.jpa.entity.CaseNote
+import java.util.UUID
 
-enum class CaseNoteEventType {
-  SENT,
-}
-
-class CaseNoteEvent(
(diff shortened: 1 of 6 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/events/CaseNoteEventPublisher.kt
```diff
@@ -5,31 +5,37 @@ import org.springframework.context.ApplicationEventPublisher
...
   val detailUrl: String,
+  val referralId: UUID,
 ) : ApplicationEvent(source)
 
 @Component
 class CaseNoteEventPublisher(
   private val applicationEventPublisher: ApplicationEventPublisher,
   private val locationMapper: LocationMapper,
 ) {
+  // send the referralId as UUID since there's no guarantee that the session is still open
```
sldblog, line 29: True, but also true in a more general sense (i.e. "send primary key IDs instead of complex hydrated JPA entities ..."). Could we write a decision record about it, please, as part of this change?

It can go here: https://github.com/ministryofjustice/hmpps-interventions-service/tree/main/doc/adr

sldblog (approved):
👍 Thanks! Can we reproduce the failures somehow?

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/events/CaseNoteEventPublisher.kt
andrewburrowsmoj: I've added a new readme in doc/adr/0003-decouple-transactions-on-asynch-calls.md

let me know what you think

andrewburrowsmoj: (I have tested locally by building a docker container.)

PR #578, opened 2021-09-23
[Delivery Session Migration Step 3] - Use delivery session. 

rgforsyth:

Deprecates `action_plan_id` on the `action_plan_session` table.
Moves from using `action_plan_session` table to using a `delivery_session` view. This will then be followed up by step 4 which will move to using a `delivery_session` table and removing the view.

To move towards a delivery session table, rather than action plan session.


src/main/resources/application.yml
```diff
@@ -117,7 +117,7 @@ interventions-ui:
       intervention-progress: "/probation-practitioner/referrals/{id}/progress"
       supplier-assessment: "/probation-practitioner/referrals/{id}/supplier-assessment"
       action-plan: "/probation-practitioner/referrals/{id}/action-plan"
-      session-feedback: "/probation-practitioner/action-plan/{id}/appointment/{sessionNumber}/post-session-feedback"
+      session-feedback: "/probation-practitioner/referrals/{id}/appointment/{sessionNumber}/post-session-feedback"
```
tomsmyers, line 120: it looks like this route is still missing in the UI. is there a PR for this?

tomsmyers (approved):
there's lots of changes here, so let's carefully validate this behaviour in preprod before pushing out to prod, on both existing cases and new referrals.

great job working through this really thorny territory though.

tomsmyers:
also i wonder where do we fill in referral ids which that are unset for sessions that were created before 053b291609dae4bfc0de7e68eb1c042f8bfb56a5 was merged ?

src/main/resources/application.yml
rgforsyth: yeah, there is still a pr up for this (https://github.com/ministryofjustice/hmpps-interventions-ui/pull/830) It's ready to go. We were just waiting to put it in around the same time that this gets merged.

rgforsyth:
> also i wonder where do we fill in referral ids which that are unset for sessions that were created before [053b291](https://github.com/ministryofjustice/hmpps-interventions-service/commit/053b291609dae4bfc0de7e68eb1c042f8bfb56a5) was merged ?

There was another pr that was merged (step 1) that backfills the referrals: https://github.com/ministryofjustice/hmpps-interventions-service/pull/568

PR #580, opened 2021-09-23
Mark all sensitive columns as "sensitive text"

sldblog:

Mark all sensitive columns as "sensitive text" by creating an effective [type alias][domain] to `text`.

🦺 Problem: this approach cannot safely mark non-textual fields "sensitive", for example "date of birth".

The main benefit is that our data documentation will be eloquent about sensitivity.

image

As a next step, we will be able to say "give me all sensitive columns and obfuscate them" - which will be useful for the pre-prod (and beyond!) refresh process.

[domain]: https://www.postgresql.org/docs/10/sql-createdomain.html


sldblog:
Switching approach to #581

PR #585, opened 2021-09-30
IC-nnnn Processing action plan submitted event (demo)

mahalrmoj:

_Required._

_Required._


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/authorization/UserMapper.kt
```diff
@@ -11,6 +11,9 @@ import uk.gov.justice.digital.hmpps.hmppsinterventionsservice.jpa.repository.Aut
 class UserMapper(
   private val authUserRepository: AuthUserRepository,
 ) {
+  fun isUserRequest(authentication: JwtAuthenticationToken): Boolean =
```
tomsmyers, line 14: i like that this implementation means that even if users are granted the API_READ_ALL role they still can't view records for which they do not have access. 

however i dislike that we have dupliaction in the logic that determines if the token is a user token (this logic exists in UserMapper::fromToken already). the method here also does not perform the full check to validate that all the required fields are present. 

2 options: 

1) change this check to be `isClientRequest`. i think the appropriate check for this is that the token's 'sub' claim is the same as the 'client_id' claim. but we should double check this. 

2) get rid of this method altogether and use `try { userMapper.fromToken(token) } catch (e: AccessDeniedException) { ... }`

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/controller/ReferralController.kt
```diff
@@ -248,8 +248,12 @@ class ReferralController(
   }
 
   private fun getSentReferralForAuthenticatedUser(authentication: JwtAuthenticationToken, id: UUID): Referral {
-    val user = userMapper.fromToken(authentication)
-    return referralService.getSentReferralForUser(id, user)
+    return if (userMapper.isUserRequest(authentication)) {
+      val user = userMapper.fromToken(authentication)
+      referralService.getSentReferralForUser(id, user)
+    } else {
+      referralService.getSentReferralForRole(id, authentication.authorities)
```
tomsmyers, line 255: as mentioned above - should we validate the role outside the context of the referral and just provide a method to get the referral without any further checking?

PR #589, opened 2021-10-01
add 'messageAttributes' to SNS events to allow us to filter them in subscriptions

tomsmyers:

adds the eventType field as a message attribute.

 this allows us to filter messages from SNS using filter policies


mahalrmoj (approved):
Added a test

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/component/SNSPublisher.kt
```diff
@@ -45,7 +46,11 @@ class SNSPublisher(
 
   private fun buildRequestAndPublish(event: EventDTO) {
     val message = objectMapper.writeValueAsString(event)
+    val messageAttributes = mapOf(
+      "eventType" to MessageAttributeValue.builder().stringValue(event.eventType).build()
```
sldblog, line 50: 🙋 We should include the `version` too

PR #593, opened 2021-10-04
disable pact broker ssl verification as temporary fix

rgforsyth:

Matching backend change to: https://github.com/ministryofjustice/hmpps-interventions-ui/pull/862
Disables SSL verification on the tag pact version job as a temporary fix.

Allows our CI pipeline to continue until a permanent fix is made.


andrewburrowsmoj (approved):
lets go for it

PR #597, opened 2021-10-08
IC-1833 new action plan approval

rgforsyth:

Allows new action-plans to be approved and then have any extra sessions created.

To allow subsequent action plans to be approved for a given referral.

- [x] Not to be merged until all 4 steps of the action-plan-session -> delivery session pr's are live.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/jpa/repository/DeliverySessionRepository.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/jpa/repository/DeliverySessionRepository.kt
@@ -11,4 +11,7 @@ interface DeliverySessionRepository : JpaRepository<DeliverySession, UUID> {
 
   @Query("select aps from DeliverySession aps left join ActionPlan ap on ap.referral = aps.referral where ap.id = :actionPlanId and aps.sessionNumber = :sessionNumber")
   fun findAllByActionPlanIdAndSessionNumber(actionPlanId: UUID, sessionNumber: Int): DeliverySession?
+
+  @Query("select ds from DeliverySession ds left join ActionPlan ap on ap.referral = ds.referral where ap.id = :actionPlanId")
+  fun findAllByActionPlanId(actionPlanId: UUID): List<DeliverySession>
 }
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/ActionPlanService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/ActionPlanService.kt
@@ -92,12 +92,7 @@ class ActionPlanService(
   fun approveActionPlan(id: UUID, user: AuthUser): ActionPlan {
     val actionPlan = getActionPlan(id)
 
-    actionPlan.referral.approvedActionPlan?. let {
-      val sessions = deliverySessionRepository.findAllByActionPlanId(id)
-      deliverySessionService.createUnscheduledSessionsForActionPlan(actionPlan, sessions.count())
-    } ?: run {
-      deliverySessionService.createUnscheduledSessionsForActionPlan(actionPlan)
-    }
+    deliverySessionService.createUnscheduledSessionsForActionPlan(actionPlan)
 
     actionPlan.approvedAt = OffsetDateTime.now()
     actionPlan.approvedBy = authUserRepository.save(user)
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/ActionPlanService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/ActionPlanService.kt
@@ -91,10 +91,17 @@ class ActionPlanService(
 
   fun approveActionPlan(id: UUID, user: AuthUser): ActionPlan {
     val actionPlan = getActionPlan(id)
+
+    actionPlan.referral.approvedActionPlan?. let {
+      val sessions = deliverySessionRepository.findAllByActionPlanId(id)
+      deliverySessionService.createUnscheduledSessionsForActionPlan(actionPlan, sessions.count())
+    } ?: run {
+      deliverySessionService.createUnscheduledSessionsForActionPlan(actionPlan)
+    }
+
     actionPlan.approvedAt = OffsetDateTime.now()
     actionPlan.approvedBy = authUserRepository.save(user)
 
-    deliverySessionService.createUnscheduledSessionsForActionPlan(actionPlan)
     val approvedActionPlan = actionPlanRepository.save(actionPlan)
     actionPlanEventPublisher.actionPlanApprovedEvent(approvedActionPlan)
     return approvedActionPlan
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/ActionPlanService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/ActionPlanService.kt
@@ -18,6 +18,7 @@ import java.util.UUID
 import java.util.UUID.randomUUID
 import javax.persistence.EntityNotFoundException
 import javax.transaction.Transactional
+import javax.validation.ValidationException
 
 @Service
 @Transactional
@@ -91,6 +92,7 @@ class ActionPlanService(
 
   fun approveActionPlan(id: UUID, user: AuthUser): ActionPlan {
     val actionPlan = getActionPlan(id)
+    verifySafeForApproval(actionPlan)
 
     deliverySessionService.createUnscheduledSessionsForActionPlan(actionPlan)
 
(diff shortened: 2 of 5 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/ActionPlanService.kt
```diff
@@ -91,10 +91,17 @@ class ActionPlanService(
 
   fun approveActionPlan(id: UUID, user: AuthUser): ActionPlan {
     val actionPlan = getActionPlan(id)
+
+    actionPlan.referral.approvedActionPlan?. let {
+      val sessions = deliverySessionRepository.findAllByActionPlanId(id)
```
andrewburrowsmoj, line 96: Is the intent here to add on top over the session numbers from the previously approved action plan?
If so, would `id` not need to be `it.id`?

andrewburrowsmoj: Also, if true, naming the variable to be more explicity on it's intent would be nice.

i.e. `sessions` -> `previouslyApprovedSessions`

andrewburrowsmoj: I also think this logic would work better in `DeliverySessionsService` encapsulated in a single method that handles session incrementation so the caller of that class doesn't have to worry about that complexity.

sldblog:
_Comments on group review_
📝 No blockers; will be picked up next

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/ActionPlanService.kt
```diff
@@ -91,10 +91,12 @@ class ActionPlanService(
 
   fun approveActionPlan(id: UUID, user: AuthUser): ActionPlan {
     val actionPlan = getActionPlan(id)
```
andrewburrowsmoj, line 93: One more thing: Do you think we should do a NOOP if actionPlan == referral.approvedActionPlan?

I suppose in the case of the user submitting it twice by mistake or double clicking...

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/DeliverySessionService.kt
```diff
@@ -35,8 +35,12 @@ class DeliverySessionService(
   val appointmentRepository: AppointmentRepository,
 ) {
   fun createUnscheduledSessionsForActionPlan(approvedActionPlan: ActionPlan) {
+    val previouslyApprovedSessions = approvedActionPlan.referral.approvedActionPlan?. let {
+      deliverySessionRepository.findAllByActionPlanId(it.id).count()
+    } ?: 0
+
     val numberOfSessions = approvedActionPlan.numberOfSessions!!
```
andrewburrowsmoj, line 42: what happens when numberOfSession <= previouslyApprovedSessions?

rgforsyth: I was told that could never happen so there was no need to check for it. Is this still true?

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/ActionPlanService.kt
sldblog: To avoid mishandling race conditions like @andrewburrowsmoj mentioned we would need to know the ID of the action plan that is being approved

Then we could reject it if
- already approved
- not the 'latest' non-approved

andrewburrowsmoj: Thinking about it some more -- the method is already idempotent since no new sessions would be created. So perhaps this check is unnecessary..

However the operation is not atomic - so it's theoretically possible for duplicate requests to happen where the second request starts when the first request is still in flight.
The possibility of this happening is very low - except in the case of a double submit with mouse button - we normally add a javascript blocker on the front-end to prevent this.

Personally I think making things atomic is the only way to have 100% certainty - but also realise that we don't do that everywhere.

sldblog: I think this is a business rule we need to enforce -- PPs shouldn't be able to approve action plans if there's a newer submitted version already (plus add in the re-approval checks)

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/ActionPlanService.kt
```diff
@@ -91,10 +91,12 @@ class ActionPlanService(
...
+
+    deliverySessionService.createUnscheduledSessionsForActionPlan(actionPlan)
```
sldblog, line 95: should we move this below `actionPlanRepository.save(actionPlan)` so that checks on already approved plans can prevent duplicate `createUnscheduledSessionsForActionPlan`?

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/DeliverySessionService.kt
```diff
@@ -35,8 +35,12 @@ class DeliverySessionService(
   val appointmentRepository: AppointmentRepository,
 ) {
   fun createUnscheduledSessionsForActionPlan(approvedActionPlan: ActionPlan) {
+    val previouslyApprovedSessions = approvedActionPlan.referral.approvedActionPlan?. let {
```
sldblog, line 38: Would it make sense to create a `approvedActionPlan.previousApprovedVersion` function? Then, in a separate PR, we could make that race-condition-resistant if we literally could walk the version history backwards

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/jpa/repository/DeliverySessionRepository.kt
```diff
@@ -11,4 +11,7 @@ interface DeliverySessionRepository : JpaRepository<DeliverySession, UUID> {
 
   @Query("select aps from DeliverySession aps left join ActionPlan ap on ap.referral = aps.referral where ap.id = :actionPlanId and aps.sessionNumber = :sessionNumber")
   fun findAllByActionPlanIdAndSessionNumber(actionPlanId: UUID, sessionNumber: Int): DeliverySession?
+
+  @Query("select ds from DeliverySession ds left join ActionPlan ap on ap.referral = ds.referral where ap.id = :actionPlanId")
```
sldblog, line 15: For neatness, we could consider adding a default sort by session number

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/DeliverySessionService.kt
sldblog: Imo we should actively reject incorrect state -- it's a business rule that new approved sessions must be >= previous approved sessions, so if that's violated, we need to

1. not reduce the number of sessions
2. alert us

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/ActionPlanService.kt
rgforsyth: for the not latest non-approved check. Should this be the latest non-approved submitted or created action plan

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/ActionPlanService.kt
```diff
@@ -91,15 +92,31 @@ class ActionPlanService(
...
-    deliverySessionService.createUnscheduledSessionsForActionPlan(actionPlan)
     val approvedActionPlan = actionPlanRepository.save(actionPlan)
     actionPlanEventPublisher.actionPlanApprovedEvent(approvedActionPlan)
     return approvedActionPlan
   }
 
+  private fun verifySafeForApproval(actionPlan: ActionPlan) {
+    if (actionPlan.approvedAt != null) {
+      throw ValidationException("Action plan has already been approved. [id=${actionPlan.id}]")
+    } else if (actionPlan.referral.actionPlans?.filter { it.approvedAt == null }?.maxByOrNull { it.createdAt }?.id != actionPlan.id) {
```
rgforsyth, line 110: is createdAt correct?

rgforsyth: it was submittedAt

andrewburrowsmoj (approved):
LGTM

PR #617, opened 2021-10-19
Detect migration clashes against mainline

sldblog:

Extends the `ci/circleci: validate_db` job with applying migrations from `main` **before** migrating from the active branch. 

The aim is to detect clashes of version numbers, which are typical after rebases.

This still won't detect clashes if branches are old or not rebased, so it's recommended to keep branch lifetime short

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -59,6 +59,14 @@ jobs:
             sudo apt update
             sudo apt install postgresql-client
       - run:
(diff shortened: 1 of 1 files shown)


sldblog:
> p.s. why won't this detect clashes if branches are not rebased?

I'm thinking of the case where the branch is old, and `main` has moved on since; unless the CI is re-run, the check can still be ✅ even if `main` has new migrations

PR #619, opened 2021-10-20
only add rows to delivery_session_appointment that have id's in delivery_session

rgforsyth:

Fixes the previous migration to ensure that only the `deliery_session_appointments` that are needed are migrated.

To fix failing migration on creation of foreign key index.


sldblog (approved):
LGTM

PR #624, opened 2021-10-22
Make session type non nullable.

andrewburrowsmoj:
Changed the AppointmentSessionType to be non-nullable when scheduling or rescheduling an appointment.

The value is defaulted to ONE_TO_ONE in the DB for historical appointments where this value was not specified.

This value is contained within the AppointmentDelivery Entity; this entity will only ever be created as part of an appointment when a schedule request has been made.

Making this value non-nullable will be safe because even though the UI defines it as a nullable field it will never be null when making a Schedule appointment request.


tomsmyers:
closing due to age  - please reopen and rebase on main if we want to pick this up again :)

PR #633, opened 2021-10-29
uJB67WFy: split s3 bucket configuration to allow multiple buckets to be configured

tomsmyers:

* adds configuration for the ndmis 'reporting' s3 bucket
* using spring configuration properties mapping to re-use the bucket config in the s3 config bean
* updates the s3 service to operate on a specific bucket instead of a generic s3 client.

allow the application to interact with multiple s3 buckets

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/config/S3Configuration.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/config/S3Configuration.kt
@@ -1,6 +1,8 @@
 package uk.gov.justice.digital.hmpps.hmppsinterventionsservice.config
 
-import org.springframework.beans.factory.annotation.Value
+import org.springframework.boot.context.properties.ConfigurationProperties
+import org.springframework.boot.context.properties.ConstructorBinding
+import org.springframework.boot.context.properties.EnableConfigurationProperties
 import org.springframework.context.annotation.Bean
 import org.springframework.context.annotation.Configuration
 import software.amazon.awssdk.auth.credentials.AwsBasicCredentials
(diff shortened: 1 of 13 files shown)


rgforsyth (approved):

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/config/S3Configuration.kt
```diff
@@ -1,34 +1,71 @@
...
+  data class Bucket(
+    val name: String
+  )
+}
+
+@ConstructorBinding
+@ConfigurationProperties(prefix = "aws.s3")
+class S3Buckets(
+  val storage: S3BucketConfiguration,
+  val ndmis: S3BucketConfiguration,
```
sldblog, line 37: Personally, I usually prefer things to be introduced when they are first used; if for any reason we drop the reporting-via-jobs work, this change is now in and `ndmisS3Bucket()` is unused.

So I'd prefer the NDMIS work to _add_ the NDMIS-related wiring in, rather than being here (plus tests, etc.)

But this is not a blocker

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/config/S3Configuration.kt
```diff
@@ -1,34 +1,71 @@
...
-  @Value("\${aws.s3.provider}") private val provider: String,
-  @Value("\${aws.s3.region}") private val region: String,
-  @Value("\${aws.s3.access-key-id}") private val accessKeyId: String,
-  @Value("\${aws.s3.secret-access-key}") private val secretAccessKey: String,
-  @Value("\${aws.s3.endpoint.uri:}") private val uri: String,
+  private val buckets: S3Buckets,
 ) {
   @Bean
-  fun s3Client(): S3Client {
+  fun storageS3Bucket(): S3Bucket {
```
sldblog, line 46: Does this need a test?

sldblog (approved):
None of my comments are blockers actually

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/config/S3Configuration.kt
```diff
@@ -1,34 +1,71 @@
...
+  fun ndmisS3Bucket(): S3Bucket {
+    return bucketFactory(buckets.ndmis)
+  }
+
+  private fun bucketFactory(config: S3BucketConfiguration): S3Bucket {
     val builder = S3Client.builder()
-      .region(Region.of(region))
-      .credentialsProvider { AwsBasicCredentials.create(accessKeyId, secretAccessKey) }
+      .region(Region.of(config.region))
+      .credentialsProvider { AwsBasicCredentials.create(config.accessKeyId, config.secretAccessKey) }
```
sldblog, line 58: Totally out of bounds of this PR, general 💭 
Can we detect misconfiguration _before_ we attempt to write to the bucket?

tomsmyers: hm yeh it would be good. i'm not aware of a way to do that, but we could add some kind of health check to the application start up ?

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/config/S3Configuration.kt
tomsmyers: agree - the only issue is that it's not obvious how this change really works unless there are two sets of configuration. that's the main reason i added it.

PR #637, opened 2021-11-01
Include some cancelled referrals for sentReferralsSummary

andrewburrowsmoj:
The `Completed cases` dashboard for SP should now include referrals which:

1. have their End of Service Report submitted
2. or were cancelled with at least feedback submitted for their supplier assessment
3. or were cancelled with at least an action plan submitted

The /sent-referrals/summary/service-provider endpoint will now return a new field called "concluded".

The UI can use this field to move all referrals where concluded is true to the `Completed cases` dashboard. This will include the unfiltered cancelled referrals.

## Notes
This is based on a conversation with Leigh where he provided the following table:

image

This shows that we should be filtering out cancelled referrals but only until their SAA feedback has been submitted or action plan has been submitted.

We don't need to consider the other scenarios in the table since they all occur after the SAA feedback submitted or Action Plan submitted.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/dto/ServiceProviderSentReferralSummaryDTO.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/dto/ServiceProviderSentReferralSummaryDTO.kt
@@ -13,6 +13,7 @@ class ServiceProviderSentReferralSummaryDTO(
   val serviceUserLastName: String?,
   val hasEndOfServiceReport: Boolean,
   val endOfServiceReportSubmitted: Boolean,
+  val concluded: Boolean,
 ) {
   companion object {
     fun from(sentReferralSummary: ServiceProviderSentReferralSummary): ServiceProviderSentReferralSummaryDTO {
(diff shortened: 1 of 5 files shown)


sldblog:
Why do we need to return `concluded`? That sounds like we're shifting logic to the UI

sldblog:
As far as I know, the action plans should make no difference in anything, why do we filter them?

andrewburrowsmoj:
@sldblog Yes we are doing the filtering on the UI side; however this is just a temporary solution until we have pagination in place. This temporary solution uses `/sent-referrals/summary/service-provider` endpoint.

We have the standard path for `/sent-referrals` which does real filtering based on JPA specification and parameters passed into the backend.

The reason why we don't do the same for `/summary` it was simpler to implement and since it will be discarded relatively soon it was deemed unnecessary to have a perfect solution.

The temp solution involves native SQL so implementing different versions of the native SQL with slightly different parameters would make the ReferralRepository class impractically long and prone to error.

Although we could do filtering after retrieving from the database and before sending back to user.

sldblog:
Ah, sorry, I wasn't commenting at the implementation -- rather that we're forced to do definitions like "all cases that have a booked SAA that are either DNA or delivered" instead of having a domain definition (and exposed in the API) of something like "referrals in 'planning' state"

Related to this, we're discussing a state machine for the current state of referrals with @toychicken

tomsmyers:
@andrewburrowsmoj does this block our updates to refactor these queries per 'dashboard type'?

sldblog:
_Comments on group review_

- Needs rebase
- Needs a refresher and an in-depth review: to track its progress (and its blocked status), add a Trello ticket
- 📝 The "finite state machine" discussion does not block this PR, can be adopted if that simplifies things

PR #640, opened 2021-11-02
Draft oasys risk info

andrewburrowsmoj:
Added draft OAsys risk information table. 
This holds the risk information entered by the user on "make a referral" journey for draft referrals.

It is used to hold the information as part of the journey until the user clicks submit referral; at this stage the risk information is posted to ARN. This data is not used as part of sent referral.

If/When we eventually make the draft referral a draft object hosted in redis then this table will not be required. For now, we need the risk information to have the same persistence retrieval time as the rest of the draft object.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/jpa/entity/DraftOasysRiskInformation.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/jpa/entity/DraftOasysRiskInformation.kt
@@ -1,0 +1,27 @@
+package uk.gov.justice.digital.hmpps.hmppsinterventionsservice.jpa.entity
+
+import org.hibernate.annotations.Fetch
+import org.hibernate.annotations.FetchMode
+import java.time.OffsetDateTime
+import java.util.UUID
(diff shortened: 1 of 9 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/jpa/entity/DraftOasysRiskInformation.kt
```diff
@@ -0,0 +1,29 @@
...
+import java.time.OffsetDateTime
+import java.util.UUID
+import javax.persistence.Entity
+import javax.persistence.Id
+import javax.persistence.ManyToOne
+import javax.persistence.Table
+import javax.validation.constraints.NotNull
+
+@Entity
+@Table
```
tomsmyers, line 14: nit: can be remvoed

andrewburrowsmoj: done

tomsmyers (approved):
nice work - don't forget that when we add this information to the referral, we need to delete it on 'referral send'.

PR #642, opened 2021-11-03
explicitly handle error codes in the customized hmpps-auth token response client

tomsmyers:

explicitly handle error codes in the customized hmpps-auth token response client

handle status codes, not just connection/ioerrors in the token response client

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/config/WebClientConfiguration.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/config/WebClientConfiguration.kt
@@ -11,7 +11,6 @@ import org.springframework.boot.web.client.RestTemplateBuilder
 import org.springframework.boot.web.client.RestTemplateCustomizer
 import org.springframework.context.annotation.Bean
 import org.springframework.context.annotation.Configuration
-import org.springframework.http.client.ClientHttpResponse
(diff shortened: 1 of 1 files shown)


mahalrmoj (approved):

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/config/WebClientConfiguration.kt
```diff
@@ -119,8 +118,14 @@ class WebClientConfiguration(
         retryTemplate.setListeners(arrayOf(RetryLogger("token request failed; retrying")))
 
         retryTemplate.execute(
-          RetryCallback<ClientHttpResponse, RestClientException> {
-            execution.execute(request, body)
+          RetryCallback {
+            execution.execute(request, body).also {
+              // i hate that this basically duplicates `RestTemplate.handleResponse`,
+              // but it's the only way i could figure out to get this to work.
```
sldblog, line 124: _not a blocker_
could listeners also work?

sldblog (approved):

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/config/WebClientConfiguration.kt
tomsmyers: you can't alter the request processing with listeners, only run blocks of code when certain things happen in the flow

PR #644, opened 2021-11-03
Refactor the custom retrying token response client into it's own component class

tomsmyers:

Refactor the custom retrying token response client into it's own component class

encapsulate all the logic for the retying client in it's own class, which will make it easier for others to reuse.


sldblog (approved):

sldblog:
Should unit tests be written for the new class?

tomsmyers:
added unit tests

PR #645, opened 2021-11-03
add unique constraint to referral id in end of service report table

rgforsyth:

Adds a unique constraint to the referral id on the `end_of_service_report` table.

To prevent multiple eosr's being created against a single referral.

### DIFF
--- a/src/main/resources/db/migration/V1_88__eosr_unique_constraint.sql
+++ b/src/main/resources/db/migration/V1_88__eosr_unique_constraint.sql
@@ -1,0 +1,2 @@
+alter table end_of_service_report
+    add unique (referral_id);


src/main/resources/db/migration/V1_88__eosr_unique_constraint.sql
```diff
@@ -0,0 +1,2 @@
+alter table end_of_service_report
+    add unique (referral_id);
```
sldblog, line 2: _danger_
`ALTER TABLE ... ADD CONSTRAINT ... UNIQUE (...)` requires an `ACCESS EXCLUSIVE` lock -- this would block all requests trying to create or read an end of service report

> [`ACCESS EXCLUSIVE`] mode guarantees that the holder is the only transaction accessing the table in any way

Could this be split into a `CONCURRENTLY` unique index and `add constraint {...} unique using index {name}`, please?

sldblog: Well, that is -- if there's any chance of this operation taking longer than 5 seconds

sldblog (approved):
There are 7000 end of service records. That's so tiny, I think we don't need to worry about concurrency just now

PR #648, opened 2021-11-05
Trello-129 Added missing SNS Events for referral end (canx, end, completion) and end of service report submit

mahalrmoj:
https://trello.com/c/JFRLDjoJ/129-missing-sns-events-for-referral-end-canx-end-completion-and-end-of-service-report-submit

Added missing SNS Events for referral end (canx, end, completion) and end of service report submit

These are omissions from day one and are necessary for the interventions-event-listener to consume and notify nDelius

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/SNSService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/SNSService.kt
@@ -10,6 +10,8 @@ import uk.gov.justice.digital.hmpps.hmppsinterventionsservice.events.ActionPlanE
 import uk.gov.justice.digital.hmpps.hmppsinterventionsservice.events.ActionPlanEventType
 import uk.gov.justice.digital.hmpps.hmppsinterventionsservice.events.AppointmentEvent
 import uk.gov.justice.digital.hmpps.hmppsinterventionsservice.events.AppointmentEventType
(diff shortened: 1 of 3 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/SNSService.kt
```diff
@@ -78,6 +80,36 @@ class SNSReferralService(
...
+      }
+      ReferralEventType.COMPLETED -> {
+        val snsEvent = EventDTO(
+          "intervention.referral.completed",
+          "A referral has been completed",
+          event.detailUrl,
+          event.referral.concludedAt!!,
+          mapOf("referralId" to event.referral.id)
+        )
+        snsPublisher.publish(event.referral.id, event.referral.currentAssignee!!, snsEvent)
```
andrewburrowsmoj, line 111: I’m wondering if endOfServiceReport.submittedBy would be more suitable even though it's always same person (I think)

andrewburrowsmoj (approved):
LGTM.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/SNSService.kt
mahalrmoj: Having thought about this, I think the person completing the referral is the assigned caseworker logically speaking. The person submitting the end of service report happens to be that person today (which automatically initiates the completion) but the person responsible for the referral from an SP perspective is the assigned person. Therefore, I think it's ok to leave it as it is.

PR #649, opened 2021-11-05
Add auth user validity flag

andrewburrowsmoj:
Alter `auth_user` table to include new column `valid_hmpps_user` which acts as a boolean flag.

The flag communicates whether the user is a HMPPS auth user and also if the id is a valid HMPPS auth user id.

The flag defaults to null which means "We don't know yet" or "Not applicable because they're not a HMPPS auth user".

This flag was created because there was a bug with HMPPS auth where the existing auth user ids for all users were being returned as a different value. This had the consequence of having duplicate auth users in the table for the same email address. We therefore need this flag to state which of the duplicate records are valid users or not.

We will be setting this flag to true or false (where applicable) in a future script.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/jpa/entity/AuthUser.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/jpa/entity/AuthUser.kt
@@ -9,6 +9,7 @@ data class AuthUser(
   @Id @NotNull val id: String,
   @NotNull val authSource: String,
   @NotNull val userName: String,
+  var deleted: Boolean? = null,
 ) {
   override fun hashCode(): Int {
     return id.hashCode()
--- a/src/main/resources/db/migration/V1_89__auth_user_validity_flag.sql
+++ b/src/main/resources/db/migration/V1_89__auth_user_validity_flag.sql
@@ -1,0 +1,3 @@
+alter table auth_user add column deleted boolean;
+
+INSERT INTO metadata (table_name, column_name, sensitive) values ('auth_user', 'deleted', FALSE);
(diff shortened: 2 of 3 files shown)


andrewburrowsmoj:
I assume we don't need to update the AuthUser entity for this flag because the scripts will be updating directly on the db.

src/main/resources/db/migration/V1_89__auth_user_validity_flag.sql
```diff
@@ -0,0 +1,3 @@
+alter table auth_user add column valid_hmpps_user boolean;
```
tomsmyers, line 1: i'm not sure we have used the term 'hmpps user' before - i'm not sure of its meaning. can we just keep the name simple like `is_valid` or `deleted` or something more generic? 

what do you think @sldblog ?

tomsmyers:
we also need to add the field to the AuthUser entity class in case we want to use it internally (i think, anyway)

src/main/resources/db/migration/V1_89__auth_user_validity_flag.sql
sldblog: should describe our intent with this flag, yeah -- `malformed` or `duplicate` etc

sldblog: extreme case could even be `selected_for_ticket123` 😁

tomsmyers: `duplicateOf uuid` ?

andrewburrowsmoj: I've changed the column to `deleted`. It defaults to `null`.

andrewburrowsmoj:
I've added boolean value to AuthUser entity

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/jpa/entity/AuthUser.kt
```diff
@@ -9,6 +9,7 @@ data class AuthUser(
   @Id @NotNull val id: String,
   @NotNull val authSource: String,
   @NotNull val userName: String,
+  val deleted: Boolean? = null,
```
tomsmyers, line 12: i think this needs to be `var` to allow us to modify it later

tomsmyers (approved):

PR #650, opened 2021-11-05
add basic job configuration for soft deleting invalid auth user ids

tomsmyers:

adds a one-off job configuration which soft-deletes invalid auth user records from the internal database

tidy up the database after a recent mix up in hmpps auth which caused us to create invalid duplicate user records.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/jobs/oneoff/SoftDeleteInvalidUsers.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/jobs/oneoff/SoftDeleteInvalidUsers.kt
@@ -1,0 +1,58 @@
+package uk.gov.justice.digital.hmpps.hmppsinterventionsservice.jobs.oneoff
+
+import mu.KLogging
+import net.logstash.logback.argument.StructuredArguments
+import org.hibernate.SessionFactory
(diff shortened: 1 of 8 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/jobs/oneoff/SoftDeleteInvalidUsers.kt
```diff
@@ -0,0 +1,58 @@
...
+    .processor(
+      ItemProcessor { user ->
+        if (user.deleted == null) {
+          val idIsInvalid = user.authSource == "auth" && !hmppsAuthService.checkAuthUserExists(user.id)
+
+          if (idIsInvalid) {
+            logger.info("soft deleting invalid auth user record", StructuredArguments.kv("user_id", user.id))
+          }
+
+          user.deleted = idIsInvalid
```
mahalrmoj, line 45: Is this processor transactional which means this change will be committed to the database?

tomsmyers: no - the change will be commited in the context of the web request.

mahalrmoj (approved):

PR #652, opened 2021-11-09
Username used in equals method to determine if two AuthUser entities represent the same hmpps-auth user

mahalrmoj:
Context: https://mojdt.slack.com/archives/CR5ESQ8T1/p1634807215003000

As a workaround for the functionality in R&M, the determination of whether two Hmpps-Auth user entities represent the same actual user is now based on username and not hmpps-auth Id. 

This purpose of this change is to ensure R&M continues to function correctly given the consequences of the issue in hmpps-auth

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/jpa/entity/AuthUser.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/jpa/entity/AuthUser.kt
@@ -1,5 +1,6 @@
 package uk.gov.justice.digital.hmpps.hmppsinterventionsservice.jpa.entity
 
+import java.util.Objects
 import javax.persistence.Entity
 import javax.persistence.Id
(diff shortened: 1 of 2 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/jpa/entity/AuthUser.kt
```diff
@@ -11,15 +11,19 @@ data class AuthUser(
...
+    return userName.hashCode()
   }
 
   override fun equals(other: Any?): Boolean {
     if (other == null || other !is AuthUser) {
       return false
     }
 
-    return id == other.id
+    return userName == other.userName
```
tomsmyers, line 27: can we use `userName && authSource` ?

i think it's ever so slightly safer than just username (as long as it satifies all the constraints you have evaluated as part of this change).

tomsmyers (approved):
LGTM - what could possibly go wrong!?! 😀

PR #653, opened 2021-11-09
Increase maximum JVM memory

sldblog:

Increase JVM maximum heap memory size

We are seeing some OutOfMemoryErrors in production on the `api-suspect`
pods, which handle `/sent-referrals/summary/service-provider`

The hard limit for our pods in our namespace is [1000MiB](https://github.com/ministryofjustice/cloud-platform-environments/blob/88ee810e1f48efa2af9eb7c78a0d7a1f3d37c619/namespaces/live-1.cloud-platform.service.justice.gov.uk/hmpps-interventions-prod/02-limitrange.yaml#L10); the current
maximum JVM heap size is 512m

To make use of the already allocated memory without drawbacks, we can
increase the maximum heap size by about ~250m

image
This is a graph of memory and CPU use -- when we run out of memory, the memory use still isn't anywhere near the hard limit (top red line)

### DIFF
--- a/helm_deploy/hmpps-interventions-service/templates/_envs.tpl
+++ b/helm_deploy/hmpps-interventions-service/templates/_envs.tpl
@@ -12,6 +12,9 @@ env:
     value: "{{ $value }}"
   {{ end }}
 
+  - name: JAVA_OPTS
+    value: "-Xmx750m"
+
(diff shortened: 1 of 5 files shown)


tomsmyers (approved):
thanks !

PR #654, opened 2021-11-09
Ndmis reports wite to csv

rgforsyth:

Performs the write to csv part of the ndmis performance report. Then publishes the csv to s3.

To create 3 seperate performance report csv's for referrals, appointments and complexity


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/reporting/BatchUtils.kt
```diff
@@ -54,14 +88,6 @@ class HeaderWriter(private val header: String) : FlatFileHeaderCallback {
   }
 }
 
-class LoggingWriter<T> : ItemWriter<T> {
```
tomsmyers, line 57: i think we can leave this in - it's useful for developing future jobs i think

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/reporting/ndmis/performance/NdmisPerformanceReportJobConfiguration.kt
```diff
@@ -57,45 +106,59 @@ class NdmisPerformanceReportJobConfiguration(
...
       .processor(processor)
-      .writer(LoggingWriter<List<AppointmentData>>())
+      .writer(writer)
       .build()
   }
+
+  private val pushToS3Step = stepBuilderFactory["pushToS3Step"]
+    .tasklet { _: StepContribution, _: ChunkContext ->
+      listOf(referralReportPath, complexityReportPath, appointmentsReportPath).forEach { path ->
+        s3Service.publishFileToS3(ndmisS3Bucket, path, "export/csv/reports/")
```
tomsmyers, line 158: the current path we are putting the csv file is `dfinterventions/dfi/csv/reports/`. do we need to put these in the same place? leigh mentioned we will have some overlap, so probably not. but we should ask where the ndmis team want them in the bucket.

tomsmyers (approved):
LGTM

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/reporting/ndmis/performance/NdmisPerformanceReportJobConfiguration.kt
rgforsyth: the ticket says: 
Upload those files to S3 bucket (confirmed)
export/csv/reports/crs_performance_report-v2-referrals.csv

PR #656, opened 2021-11-09
Dump heap on OutOfMemory to a volume

sldblog:

Dump heap on OutOfMemory to a volume so we can analyse it later

Objective: get a heap dump when we run out of memory, so we can analyse the heap

This heap dump would need to persist between container restarts; to achieve this, we put it in a volume

We can then `kubectl cp` the files afterwards


tomsmyers (approved):
LGTM

sldblog:
I'm thinking we should merge this tomorrow

PR #665, opened 2021-11-12
make the ndmis reporting job more easily runnable from the command line

tomsmyers:

add spring batch 'JobParametersIncrementer' which allows the ndmis reporting job to run without the launcher specifying the timestamp

means you don't need to specify the timestamp param by hand, which makes running this as a k8s cronjob simpler.


rgforsyth (approved):
Nice one!

PR #666, opened 2021-11-12
stop report from breaking on null pointer exception

rgforsyth:

keeps the report processing if a null pointer is thrown

To make the report more resilient to data issues.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/reporting/ndmis/performance/ComplexityProcessor.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/reporting/ndmis/performance/ComplexityProcessor.kt
@@ -1,7 +1,6 @@
 package uk.gov.justice.digital.hmpps.hmppsinterventionsservice.reporting.ndmis.performance
 
 import mu.KLogging
-import net.logstash.logback.argument.StructuredArguments
 import org.springframework.stereotype.Component
 import uk.gov.justice.digital.hmpps.hmppsinterventionsservice.jpa.entity.Referral
 import uk.gov.justice.digital.hmpps.hmppsinterventionsservice.reporting.SentReferralProcessor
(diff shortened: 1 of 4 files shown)


tomsmyers:
i think this is better configured in one place using the built in error handling mechanism in spring batch. you need to configure a step policy and add it into the step builder call. at it's simplest it's something like this:

``` 
    stepBuilderFactory["someStep"]
    ....   
    .faultTolerant()
    .skipPolicy { t: Throwable, _: Int ->
      when (t) {
        is NullPointerException -> {
          logger.warn("skipping referral with unexpected state", t)
          true
        }
        else -> false
      }
    }
```

using this scheme you can also collect these invalid referrals into one place, and process them in a later step. we can implement that later though.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/reporting/ndmis/performance/ComplexityProcessor.kt
```diff
@@ -17,8 +17,7 @@ class ComplexityProcessor() : SentReferralProcessor<List<ComplexityData>> {
         interventionTitle = referral.intervention.title,
         serviceCategoryId = it.id,
         serviceCategoryName = it.name,
-        complexityLevelTitle = it.complexityLevels.find {
-          complexityLevel ->
+        complexityLevelTitle = it.complexityLevels.find { complexityLevel ->
```
tomsmyers, line 20: this changes the logic in the line below... 

`it` is the selected service category

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/reporting/ndmis/performance/NdmisPerformanceReportJobConfiguration.kt
```diff
@@ -161,16 +168,30 @@ class NdmisPerformanceReportJobConfiguration(
...
+      .skipPolicy(skipPolicy)
       .build()
   }
 
   private val pushToS3Step = stepBuilderFactory["pushToS3Step"]
     .tasklet { _: StepContribution, _: ChunkContext ->
       listOf(referralReportPath, complexityReportPath, appointmentsReportPath).forEach { path ->
         s3Service.publishFileToS3(ndmisS3Bucket, path, "export/csv/reports/")
-        File(path.toString()).delete()
+//        File(path.toString()).delete()
```
tomsmyers, line 180: what's the reason for commenting this line out?

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/reporting/ndmis/performance/NdmisPerformanceReportJobConfiguration.kt
```diff
@@ -161,16 +168,30 @@ class NdmisPerformanceReportJobConfiguration(
...
       }
       RepeatStatus.FINISHED
     }
     .build()
 }
+
+class NPESkipPolicy : SkipPolicy {
```
tomsmyers, line 187: should this live in the batch utils class or here?

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/reporting/ndmis/performance/NdmisPerformanceReportJobConfiguration.kt
```diff
@@ -161,16 +168,30 @@ class NdmisPerformanceReportJobConfiguration(
...
+  override fun shouldSkip(t: Throwable, skipCount: Int): Boolean {
+    return when (t) {
+      is NullPointerException -> {
+        NdmisPerformanceReportJobConfiguration.logger.warn("skipping referral with unexpected state", t)
```
tomsmyers, line 191: nit: since this class is defined on its own, it's generic in terms of what it's skipping. `skipping referrral...` may not always be correct.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/reporting/ndmis/performance/NdmisPerformanceReportJobConfiguration.kt
rgforsyth: accident. (used to test the reports locally)

tomsmyers (approved):
LGTM

PR #669, opened 2021-11-15
update path prefix

rgforsyth:

Updated the path prefix and sets the file names as variables.

Path was incorrect.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/reporting/ndmis/performance/NdmisPerformanceReportJobConfiguration.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/reporting/ndmis/performance/NdmisPerformanceReportJobConfiguration.kt
@@ -44,11 +44,16 @@ class NdmisPerformanceReportJobConfiguration(
   @Value("\${spring.batch.jobs.ndmis.performance-report.chunk-size}") private val chunkSize: Int,
 ) {
   companion object : KLogging()
+  private val pathPrefix = "dfinterventions/dfi/csv/reports/"
+  private val referralReportFilename = "crs_performance_report-v2-referrals.csv"
+  private val complexityReportFilename = "crs_performance_report-v2-complexity.csv"
+  private val appointmentReportFilename = "crs_performance_report-v2-appointments.csv"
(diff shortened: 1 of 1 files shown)


tomsmyers (approved):

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/reporting/ndmis/performance/NdmisPerformanceReportJobConfiguration.kt
```diff
@@ -44,11 +44,16 @@ class NdmisPerformanceReportJobConfiguration(
   @Value("\${spring.batch.jobs.ndmis.performance-report.chunk-size}") private val chunkSize: Int,
 ) {
   companion object : KLogging()
+  private val pathPrefix = "dfinterventions/dfi/csv/reports/"
```
sldblog, line 47: _not a blocker_
This is tightly coupled with the environment: https://github.com/ministryofjustice/cloud-platform-environments/blob/26a924086f42f2edff4502734d0dcdf3a0961364/namespaces/live-1.cloud-platform.service.justice.gov.uk/hmpps-interventions-prod/resources/reporting-landing-access.tf#L58

To be independent from this path changing in the infrastructure, as a next iteration, we could read it from a secret

sldblog (approved):
👍 Thank you

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/reporting/ndmis/performance/NdmisPerformanceReportJobConfiguration.kt
sldblog: We could also change 

https://github.com/ministryofjustice/data-engineering-data-extractor/blob/v1.1.1/modules/transfer_local_to_s3.sh#L7

https://github.com/ministryofjustice/hmpps-interventions-service/blob/main/helm_deploy/hmpps-interventions-service/templates/cronjob-data-extractor-analytics.yaml#L38-L42

to align with the dual "bucket_name", "prefix" configuration that's more aligned with the AWS SDK

PR #670, opened 2021-11-15
provide the option to use 'acl bucket owner full control' option with the S3Service

tomsmyers:

provide the option to use 'acl bucket owner full control' option with the S3Service

allows us to write to the ndmis reporting bucket

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/S3Service.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/S3Service.kt
@@ -28,7 +28,6 @@ class S3Service {
       .apply { acl?.let { acl(it) } }
       .build()
 
-
     bucket.client.putObject(objectRequest, RequestBody.fromFile(path))
   }
 }
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/S3Service.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/S3Service.kt
@@ -4,6 +4,7 @@ import mu.KLogging
 import net.logstash.logback.argument.StructuredArguments
 import org.springframework.stereotype.Service
 import software.amazon.awssdk.core.sync.RequestBody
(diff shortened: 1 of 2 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/S3Service.kt
```diff
@@ -24,8 +25,14 @@ class S3Service {
     val objectRequest = PutObjectRequest.builder()
       .bucket(bucket.bucketName)
       .key(key)
+      .apply {
+        if (bucketOwnerFullControl) {
+          acl(ObjectCannedACL.BUCKET_OWNER_FULL_CONTROL)
```
sldblog, line 30: Should this get a unit test?

sldblog (approved):
👍

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/S3Service.kt
tomsmyers: i actually changed the signature slightly - so you just pass the ACL directly. i can add a test to ensure it's appliied properly.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/S3Service.kt
```diff
@@ -24,8 +25,10 @@ class S3Service {
     val objectRequest = PutObjectRequest.builder()
       .bucket(bucket.bucketName)
       .key(key)
+      .apply { acl?.let { acl(it) } }
       .build()
 
+
```
sldblog, line 31: This seems to displease `:ktlintMainSourceSetCheck`

sldblog (approved):
👍

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/S3Service.kt
sldblog: Pushed e6d7113b3d0dfe021865d69bc6fe415675e66d4c to fix

PR #675, opened 2021-11-16
Trello-129 Added missing session feedback domain events for both initial assessments and service delivery appointments

mahalrmoj:
Added missing session feedback domain events for both initial assessments and service delivery appointments. These have been missing from day one.

These events required by the interventions-event-listener for it to update nDelius (via community-api)

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/SNSService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/SNSService.kt
@@ -143,6 +143,27 @@ class SNSActionPlanAppointmentService(
 
         snsPublisher.publish(referral.id, appointment.appointmentFeedbackSubmittedBy!!, snsEvent)
       }
+      ActionPlanAppointmentEventType.SESSION_FEEDBACK_RECORDED -> {
+        val referral = event.deliverySession.referral
+        val appointment = event.deliverySession.currentAppointment
+          ?: throw RuntimeException("event triggered for session with no appointments")
+
+        val eventType = "intervention.session-appointment.session-feedback-submitted"
+
+        val snsEvent = EventDTO(
+          eventType,
+          "Session feedback submitted for a session appointment",
+          event.detailUrl,
+          appointment.appointmentFeedbackSubmittedAt!!,
+          mapOf(
+            "serviceUserCRN" to referral.serviceUserCRN,
+            "referralId" to referral.id,
+            "deliusAppointmentId" to appointment.deliusAppointmentId.toString()
+          )
+        )
+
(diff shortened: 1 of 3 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/SNSService.kt
```diff
@@ -141,6 +141,27 @@ class SNSActionPlanAppointmentService(
...
+
+        val snsEvent = EventDTO(
+          eventType,
+          "Session feedback submitted for a session appointment",
+          event.detailUrl,
+          appointment.appointmentFeedbackSubmittedAt!!,
+          mapOf(
+            "serviceUserCRN" to referral.serviceUserCRN,
+            "referralId" to referral.id,
+            "deliusAppointmentId" to appointment.deliusAppointmentId.toString()
```
tomsmyers, line 161: should we add this field to other appointment events?

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/SNSService.kt
```diff
@@ -173,6 +194,26 @@ class SNSAppointmentService(
           mapOf("serviceUserCRN" to referral.serviceUserCRN, "referralId" to referral.id)
         )
 
+        snsPublisher.publish(referral.id, appointment.appointmentFeedbackSubmittedBy!!, snsEvent)
+      }
+      AppointmentEventType.SESSION_FEEDBACK_RECORDED -> {
```
tomsmyers, line 199: this duplication isn't nice ... is this a hang over from the recent appointment changes? is there a way to factor this out?

tomsmyers (approved):
LGTM

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/SNSService.kt
mahalrmoj: I see this as something internal to interventions but have put this on the message so that community-api has all the info it needs to update the correct appointment.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/SNSService.kt
mahalrmoj: There is much duplication between AppointmentEvent and ActionPlanAppointmentEvent. I think the latter needs to be migrated to using the later Appointment design. I have raised a task for it. 
https://trello.com/c/g7YJYkRI/154-are-actionplanappointmentevents-still-appropriate

PR #679, opened 2021-11-17
update end reason description to correct field

rgforsyth:

updates the end reason description in the referral processor to the correct field.

The end reason description was set to the wrong field.


tomsmyers (approved):

tomsmyers:
why isn't the test failing (line 59) ?

sldblog:
I'm assuming test isn't failing because it's not asserting against known initial values, meaning that the values are most likely `""`
```
    assertThat(result.endReasonDescription).isEqualTo(referral.endRequestedComments)
```

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/jpa/entity/Referral.kt
```diff
@@ -104,4 +104,16 @@ class Referral(
 
   val currentAssignee: AuthUser?
     get() = currentAssignment?.assignedTo
+
+  val endState: String? get() {
```
sldblog, line 108: nice

sldblog (approved):

PR #680, opened 2021-11-17
Switch back to the old psql-like datetime format

sldblog:

Switch back to the old psql-like datetime format

## Implementation comments

I opted for a separate data class, to make this super explicit

We reached a compromise with the NDMIS team that we would retain the
previous datetime format that psql gave us:

`2021-11-13 17:55:51.067867+00` in winter (GMT)
`2021-06-11 16:06:38.662521+00` in summer (BST)

The ISO format we use now is:

`2021-11-13T17:55:51.067867Z` in winter (GMT)
`2021-06-11T17:06:38.662521+01:00` in summer (BST)

This is because changing the ETL pipeline requires re-prioritisation of
other work, even for a datetime parsing change


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/reporting/ndmis/performance/NdmisDateTime.kt
```diff
@@ -0,0 +1,14 @@
+package uk.gov.justice.digital.hmpps.hmppsinterventionsservice.reporting.ndmis.performance
+
+import java.time.OffsetDateTime
+import java.time.ZoneOffset
+import java.time.format.DateTimeFormatter
+
+// Agreement: the v2 NDMIS report uses this format, v3 switches to proper ISO8601
+// Deadline: stop producing v2 format after 13 December 2021
+data class NdmisDateTime(val dateTime: OffsetDateTime) {
+  override fun toString(): String {
```
sldblog, line 10: This all hinges on the serialiser using `toString` -- is that the case?

tomsmyers: yes  - you can test this by adding to the `CsvLineAggregatorTest`

tomsmyers (approved):
nice solution

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/reporting/ndmis/performance/NdmisDateTime.kt
```diff
@@ -0,0 +1,14 @@
+package uk.gov.justice.digital.hmpps.hmppsinterventionsservice.reporting.ndmis.performance
+
+import java.time.OffsetDateTime
+import java.time.ZoneOffset
+import java.time.format.DateTimeFormatter
+
+// Agreement: the v2 NDMIS report uses this format, v3 switches to proper ISO8601
+// Deadline: stop producing v2 format after 13 December 2021
```
tomsmyers, line 8: it didn't sound like the NDMIS team had signed off on this?

tomsmyers:
one slight nit: i would probably have extended the OffsetDatetime class instead of wrapping it. this way it behaves the same in every other regard except for serialization to string.

tomsmyers:
oh it's a final class.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/reporting/ndmis/performance/NdmisDateTime.kt
sldblog: yeah, not yet. I'll amend the deadline when we agree

sldblog:
thanks for the review + extra test!

PR #682, opened 2021-11-18
Async exception handler to not swallow exception so that top level exception handler reports full exception stack, e.g. below null pointer source is reported

mahalrmoj:
Async handler no longer swallows exception and is now caught by top level exception handler. Full stack trace is shown giving source of errors.

```
2021-11-18 08:48:21.280 ERROR 31069 --- [TaskExecutor-42] u.g.j.d.h.h.e.AsyncEventExceptionAdvice  : Exception thrown for method annotated with @AsyncEventExceptionHandling java.lang.NullPointerException
Exception in thread "SimpleAsyncTaskExecutor-42" java.lang.NullPointerException
	at uk.gov.justice.digital.hmpps.hmppsinterventionsservice.service.SNSActionPlanService.onApplicationEvent(SNSService.kt:35)
	at uk.gov.justice.digital.hmpps.hmppsinterventionsservice.service.SNSActionPlanService.onApplicationEvent(SNSService.kt:22)
	at uk.gov.justice.digital.hmpps.hmppsinterventionsservice.service.SNSActionPlanService$$FastClassBySpringCGLIB$$5dfb72bb.invoke(<generated>)
```

To be able to pinpoint the source of the exception which is not currently the case.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/exception/AsyncEventExceptionAdvice.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/exception/AsyncEventExceptionAdvice.kt
@@ -20,6 +20,7 @@ class AsyncEventExceptionAdvice {
     } catch (exception: Throwable) {
       val event = (pjp as MethodInvocationProceedingJoinPoint).args.getOrNull(0)
       logger.error("Exception thrown for method annotated with @AsyncEventExceptionHandling {}", exception, kv("event", event))
+      throw exception
     }
   }
 }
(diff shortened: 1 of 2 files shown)


tomsmyers (commented):
doesnt this just have the same effect as removing the exception handler altogether? 

what was the original intention for logging and swallowing exceptions here?

mahalrmoj:
> doesnt this just have the same effect as removing the exception handler altogether?
> 
> what was the original intention for logging and swallowing exceptions here?

We log the event which gives us context

`event=ReferralEvent(type=COMPLETED, referral=af978e48-3b9d-42cd-a81c-62602e39fc87, detailUrl='https://hmpps-interventions-service.apps.live-1.cloud-platform.service.justice.gov.uk/sent-referral/af978e48-3b9d-42cd-a81c-62602e39fc87', source=uk.gov.justice.digital.hmpps.hmppsinterventionsservice.events.ReferralEventPublisher@1bd6831d)`

tomsmyers (approved):
it would be better if we could figure out how to get the logger to report these exceptions properly instead of throwing it, but it's ok for now.

PR #683, opened 2021-11-18
Bug Fix: Don't include old referrals for myCases

andrewburrowsmoj:
Some users were reporting that they were seeing that they were being assigned to the wrong referral on my cases dashboard.
    
After some investigation it appears that for MyCases dashboard we are returning a referral if the user has ***ever*** been assigned to it and not just if they're the latest assigned user.
    
This bug fix moves the custom SQL checks to be after filtering for the latest assigned referral has been done.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/jpa/repository/ReferralSummaryRepositoryImpl.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/jpa/repository/ReferralSummaryRepositoryImpl.kt
@@ -50,8 +50,7 @@ class ReferralSummaryRepositoryImpl : ReferralSummaryRepository {
 		  r.sent_at is not null
 		  and ( dfc.prime_provider_id in :serviceProviders or dfcsc.subcontractor_provider_id in :serviceProviders )
       and not (r.concluded_At is not null and r.end_Requested_At is not null and eosr.id is null)
-      $dashboardRestrictionCriteria
-) a where assigned_at_desc_seq = 1"""
+) a where assigned_at_desc_seq = 1 $dashboardRestrictionCriteria"""
   }
 
   override fun getSentReferralSummaries(authUser: AuthUser, serviceProviders: List<String>, dashboardType: DashboardType?): List<ServiceProviderSentReferralSummary> {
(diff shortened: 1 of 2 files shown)


src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/jpa/repository/ReferralRepositoryTest.kt
```diff
@@ -224,6 +224,18 @@ class ReferralRepositoryTest @Autowired constructor(
 
   @Nested
   inner class DashboardTypeSelection {
+
+    @Test
+    fun `myCases dashboard will not return self assigned if they're not latest`() {
```
sldblog, line 229: What do you think about?
```suggestion
    fun `does not show referrals where the user has been historically assigned but no longer the active assignee`() {
```

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/jpa/repository/ReferralSummaryRepositoryImpl.kt
```diff
@@ -50,8 +50,7 @@ class ReferralSummaryRepositoryImpl : ReferralSummaryRepository {
 		  r.sent_at is not null
 		  and ( dfc.prime_provider_id in :serviceProviders or dfcsc.subcontractor_provider_id in :serviceProviders )
       and not (r.concluded_At is not null and r.end_Requested_At is not null and eosr.id is null)
-      $dashboardRestrictionCriteria
-) a where assigned_at_desc_seq = 1"""
+) a where assigned_at_desc_seq = 1 $dashboardRestrictionCriteria"""
```
sldblog, line 53: to note, `referral_assignments` has a `superseded` flag (still needs to be backfilled) which should allow for simpler conditions -- ie. `assignment.superseded = false` would mean it's the latest one

sldblog (approved):

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/jpa/repository/ReferralSummaryRepositoryImpl.kt
sldblog: or perhaps, a better pattern would be introducing `previous_version_id` to create a linked list for _any_ versioned entry

andrewburrowsmoj: Good points. This PR has the least divergence with the existing native SQL; but perhaps we should have this in for consideration for the pagination jpa specification route?

PR #692, opened 2021-11-22
Edit oasys risk information  - submit to arn

andrewburrowsmoj:
Provides ability to post full risk information to ARN as part of mark a referral journey.

As part of make a referral journey a Draft OAsys risk information object is created. This is what is used to post to ARN if it exists.

Given that this will be deployed whilst some users are in the process of making a referral, we must still allow for an empty draft oasys risk information object - however only if the original additional risk information object is not null.

Once the risk information has been submitted then the draft information is deleted as we must not store this information given that ARN is the source of truth.


helm_deploy/values-dev.yaml
```diff
@@ -25,3 +25,4 @@ env:
   SENTRY_ENVIRONMENT: "dev"
   AWS_S3_STORAGE_ENABLED: "true"
   AWS_S3_NDMIS_ENABLED: "true"
+  ASSESSRISKSANDNEEDS_ENABLEPOSTINGFULLRISK: "true"
```
rgforsyth, line 28: Not sure if it matters, but missing newline.

rgforsyth (approved):

PR #693, opened 2021-11-22
Enable v2 NDMIS report in production

sldblog:

Enable v2 NDMIS report in production

Stakeholders are 🤩 happy with pre-prod data and behaviour


tomsmyers (approved):
LGTM - are we happy with the duration and load this causes on the production database (even though it's in the middle of the night?)

sldblog:
Personally, the uncomfortable zone for me starts at 30 minutes and the unacceptable zone starts at 60 minutes

💭 We can create a PrometheusRule about the pod's lifetime as a proxy for an immediate alert

PR #695, opened 2021-11-22
rename old action plan session tables

rgforsyth:

Rename's the old `action_plan_session` and `action_plan_session_appointment` tables

Tidy up of the action plan to delivery session db changes


sldblog (commented):
Nice 👍

The metadata CI fail can be resolved by updating the table_name in those entries

Could we perhaps update the comment for these tables in the data dictionary migration too, please? For data users, it would be useful to know where they should look for this information  now.

sldblog (approved):
Thanks!

Would be good to add a comment in the data dictionary too, but it's not a blocker

PR #696, opened 2021-11-22
Improve filtering for SP Completed dashboard

andrewburrowsmoj:
The Completed cases dashboard for SP should now include referrals which:

1. have their End of Service Report submitted
2. or were cancelled with at least feedback submitted for their supplier assessment
3. or were cancelled with at least an action plan submitted

The other dashboards now filter on concludedAt rather than EOSR to prevent some cancelled referrals being included.

## Notes:
This is based on a conversation with Leigh where he provided the following table:
image

This shows that we should be including some cancelled referrals in the Completed cases. intervention service does the filtering so only the correct cancelled referrals will be returned.


tomsmyers:
erugh all this stuff is so complicated isn't it. is there somewhere all these rules are presented in a friendly format, like on confluence, or in the readme of this repo?

andrewburrowsmoj:
Once we switch over to pagination, they will be more neatly encapsulated in the `ReferralSpecifications` class; each rule can be it's own method. We currently have `cancelled` already defined in this class.

tomsmyers (approved):

PR #697, opened 2021-11-23
Additional logging for NullPointerException when getting CurrentAssignee 

mahalrmoj:
https://trello.com/c/RkihX2uQ/159-prod-issue-nullpointerexception-in-getcurrentassignment

Add a null check on the current assignment, with logging, to the app event publisher method as a fact finding change for two reasons:
1. To confirm whether the current assignment exists when in the main transactional thread
2. By doing this, it will lazy load the current assignee and can confirm there is a lazy loading issue (as the problem will go away)
If this proves to be correct then we can formulate a more elegant solution.

```
Crashed in non-app: org.hibernate.collection.internal.AbstractPersistentCollection in withTemporarySessionIfNeeded
uk.gov.justice.digital.hmpps.hmppsinterventionsservice.jpa.entity.Referral in getCurrentAssignment at line 152
uk.gov.justice.digital.hmpps.hmppsinterventionsservice.jpa.entity.Referral in getCurrentAssignee at line 106
uk.gov.justice.digital.hmpps.hmppsinterventionsservice.service.SNSReferralService in onApplicationEvent at line 111
uk.gov.justice.digital.hmpps.hmppsinterventionsservice.service.SNSReferralService in onApplicationEvent at line 54
```

Provides additional insight into the cause of the problem.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/events/ReferralEventPublisher.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/events/ReferralEventPublisher.kt
@@ -1,5 +1,6 @@
 package uk.gov.justice.digital.hmpps.hmppsinterventionsservice.events
 
+import mu.KLogging
 import org.springframework.context.ApplicationEvent
 import org.springframework.context.ApplicationEventPublisher
 import org.springframework.stereotype.Component
@@ -22,6 +23,7 @@ class ReferralEventPublisher(
   private val applicationEventPublisher: ApplicationEventPublisher,
   private val locationMapper: LocationMapper
 ) {
+  companion object : KLogging()
 
   fun referralSentEvent(referral: Referral) {
     applicationEventPublisher.publishEvent(ReferralEvent(this, ReferralEventType.SENT, referral, getSentReferralURL(referral)))
@@ -32,6 +34,7 @@ class ReferralEventPublisher(
   }
 
   fun referralConcludedEvent(referral: Referral, eventType: ReferralEventType) {
+    referral.currentAssignee ?: logger.warn("Concluding referral has no current assignment ${referral.id} for event type $eventType")
     applicationEventPublisher.publishEvent(ReferralEvent(this, eventType, referral, getSentReferralURL(referral)))
   }
 
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/jpa/entity/AuthUser.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/jpa/entity/AuthUser.kt
@@ -24,4 +24,8 @@ data class AuthUser(
 
     return userName == other.userName && authSource == other.authSource
   }
+  companion object {
+    // System user to represent the actor for events that are automatically generated
+    val interventionsServiceUser = AuthUser("00000000-0000-0000-0000-000000000000", "fake", "hmpps-interventions-service")
+  }
 }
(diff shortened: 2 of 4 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/events/ReferralEventPublisher.kt
```diff
@@ -32,6 +34,7 @@ class ReferralEventPublisher(
   }
 
   fun referralConcludedEvent(referral: Referral, eventType: ReferralEventType) {
+    referral.currentAssignee ?: logger.warn("Concluding referral has no current assignment ${referral.id} for event type $eventType")
```
sldblog, line 37: This message feels like knowing how the published message is going to be used. Should this be on the listener side?

sldblog: Alternatively, this feels like a kind of "this instance of generic data has more constraints" which feel like calls for a separate event class

sldblog:
Looking at https://github.com/ministryofjustice/hmpps-interventions-service/blob/4a1f975b89c1e9952f19c24b4723ce9a4b2faeb4/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/SNSService.kt#L103-L111

This can legitimately happen where a referral is made by a mistake and cancelled before it's picked up.

The user is used as the _actor_ for the event; there is no direct actor here as this event is a system-generated one sitting behind a set of business rules... Perhaps we could reflect that?

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/events/ReferralEventPublisher.kt
sldblog: Not a blocker though -- the reasons in the PR (which I read too late, sorry) are sound

mahalrmoj:
> This exception can legitimately happen where a referral is made by a mistake and cancelled before it's picked up.
There are two other events that are raised by the application, one is cancelled and the other prematurely ended. In those cases the user requesting the end is used as the _actor_ of the event.

> The user is used as the _actor_ for the event; there is no direct actor here as this event is a system-generated one sitting behind a set of business rules... Perhaps we could reflect that?
Yeah this is an interesting one. The referral is automatically "completed" on the submission of the end of service report and as a system generated event we could use "interventions" as the _actor_. In that case, our potential lazy load issue would also go away. Do we have an interventions system user or would we literally user the literal of "interventions"?

sldblog:
Ah, I see. I didn't see `ReferralEventType.CANCELLED` before.

domain state machine

Based on the above, the `service ended (concluded)` event is a system-generated one -- so yeah, I think a literal `AuthUser("hmpps-interventions-service", ...)` would work fine, right?

Note: the domain events/states above don't have `intervention.referral.cancelled` events -- I think it's worth reconciling which interpretation is right

sldblog:
We probably don't want to pass `AuthUser` there; we only need to identify the user explicitly.
Auth now has a global unique ID ... we could simulate a system UUID with a zero-UUID.

mahalrmoj:
> we could simulate a system UUID with a zero-UUID

so what values do we set authSource and userName too as we currently send an AuthUser?

id/source/username
0000000-0000-0000-0000-0000000000/auth/interventions

sldblog:
Sorry, I was wondering about the time after we shift to the global UUID now provided by hmpps-auth -- it's not here yet

Now an in-memory `AuthUser("0000000-0000-0000-0000-0000000000", "fake", "hmpps-interventions-service")` is sufficient; we only use the `id` and `userName` in tracking `InterventionsDomainEvent` events in AppInsights

sldblog (approved):

PR #698, opened 2021-11-23
remove unused 'cancelled' method in referral entity

tomsmyers:

removes a method called 'cancelled' in the referral entity. this was unused. the logic is duplicated in the 'endState' getter. so if future 'us' did want to use this method they could just check `referral.endState == 'cancelled'`.

clean up


andrewburrowsmoj (approved):

sldblog (approved):
> if future 'us' did want to use this method they could just check `referral.endState == 'cancelled'

If we introduce comparisons we should probably switch to an enumerable at that point

PR #699, opened 2021-11-23
Ability to disable calling community-api with appointment outcomes when event listener is in production

mahalrmoj:
https://trello.com/c/xHVqOzpw/155-notify-appointment-outcome-via-event-listener

Ability to disable calling community-api with appointment outcomes when event listener takes over the responsibility for this.

Facilitates the migration to the interventions to delius event listener notification route


sldblog (commented):
Could we include the `helm_deploy/values-{env}.yaml` values too, please?

sldblog (approved):

PR #703, opened 2021-11-24
Enforce data documentation on tables

sldblog:

Enables CI check on whether tables have data documentation. (This PR does not enforce column definitions.)

To make sure we don't needlessly document "internal" tables, a new `domain_data` flag is added to the `metadata` table.

The current data dictionary tables are a bit patchy:
image

The aim is to make sure the _relevant_ bits are filled out.

## After these changes

The main page will look like this:
image


mahalrmoj (approved):
LGTM - nice that it generates instructions on what needs to be generated

PR #705, opened 2021-11-25
Add debug log for arn submission

andrewburrowsmoj:
Added logger debug level for dev environment and a debug log to indicate when we are sending full supplementary risk info to ARN.


src/main/resources/logback-spring.xml
```diff
@@ -71,5 +71,9 @@
         <root level="info">
             <appender-ref ref="stdout" />
         </root>
+
+        <logger name="uk.gov.justice.digital.hmpps.hmppsinterventionsservice" level="${LOGGING_LEVEL_UK_GOV_JUSTICE_DIGITAL_HMPPS_HMPPSINTERVENTIONSSERVICE:-info}" additivity="false">
```
tomsmyers, line 75: you shouldn't have to add this - i tried it out and it worked with just the env var for me

tomsmyers (approved):

src/main/resources/logback-spring.xml
andrewburrowsmoj: oh that's awsome, thanks

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/RisksAndNeedsService.kt
```diff
@@ -67,6 +67,7 @@ class RisksAndNeedsService(
   fun createSupplementaryRisk(referralId: UUID, crn: String, user: AuthUser, riskCreatedAt: OffsetDateTime, riskInformation: String, redactedRisk: RedactedRisk? = null): UUID {
 
     val request = if (canPostFullRiskInformation && redactedRisk != null) {
+      logger.debug("Sending full supplementary risk information to ARN")
```
sldblog, line 70: Would this be possible to monitor via AppInsights and queries that specifically look for dependency calls?

Context: I wanted to initially ask if this would be useful in other environments -- then wanted to ask if it's worth having a custom metric for this -- then realised dependency calls are tracked in AppInsights

PR #709, opened 2021-11-26
Show case note event details in exceptions

sldblog:

Show case note event details in exceptions

We use these `toString`s to show details about the events in case of failure. The default toString doesn't include any IDs

This lead to unhelpful events: https://sentry.io/organizations/ministryofjustice/issues/2814711914/events/4820c55ecf314928a5639d2ef1ae68c2/?project=5807819#extra

image

A custom `toString` solves this


mahalrmoj (approved):
LGTM - good spot - all other event types seem to have been given the same treatment already

PR #714, opened 2021-11-30
Deploy to pre-prod after successful dev deploy

sldblog:

In other words, remove the manual approval to pre-prod.

As the majority of our changes (99%+) are correct and safe, there is no
reason to hold back. My hypothesis is that if we get the changes to a
live-like environment sooner, we will be aware of data
problems/migration problems sooner.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -312,18 +312,13 @@ workflows:
           tag: "deployed:dev"
           requires: [deploy_dev]
           context: [hmpps-common-vars]
-      - approve_postdev:
-          type: approval
-          requires:
-            - deploy_dev
       - hmpps/deploy_env:
           name: deploy_preprod
           env: "preprod"
           show_changelog: true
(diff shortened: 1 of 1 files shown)


jacksonj04 (approved):
This looks like a sensible change to me - if we're confident in our test suite there's no real purpose in changes waiting around for manual approval.

sldblog:
Thanks @jacksonj04!

What do you think, @mahalrmoj @tomsmyers @andrewburrowsmoj @rgforsyth ?

mahalrmoj:
> Thanks @jacksonj04!
> 
> What do you think, @mahalrmoj @tomsmyers @andrewburrowsmoj @rgforsyth ?

I think in principle [and most of the time] yes. On the odd occasion, only certain things can be verified in preprod as the ref data in DEV/TEST for nDelius is not representative. This has been raised with them many times and I believe is still an issue.

PR #715, opened 2021-11-30
Switch to EKS cluster on the dev environment

sldblog:

- Switch dev env to 100% new cluster
- Remove weights to support 100% switchover instead

We will switch to the new cluster in one go. This is because the
hmpps-delius-interventions-listener and hmpps-interventions-ui both only
support complete switchovers; so let's do it the same way

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -287,25 +287,6 @@ workflows:
             - vulnerability_scan_and_monitor
           context:
             - hmpps-common-vars
-      - hmpps/deploy_env:
-          name: deploy_dev_newcluster
-          env: "dev"
-          slack_notification: true
-          slack_channel_name: "interventions-dev-notifications"
-          helm_additional_args: "--set ingress.migration.colour=green"
-          filters:
-            branches:
-              only:
-                - main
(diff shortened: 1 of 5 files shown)


sldblog:
[❗️Need to migrate dev/research env to EKS cluster in Cloud Platform](https://trello.com/c/UQgzJJr6/122-%E2%9D%97%EF%B8%8Fneed-to-migrate-dev-research-env-to-eks-cluster-in-cloud-platform)

.circleci/config.yml
```diff
@@ -287,25 +287,6 @@ workflows:
             - vulnerability_scan_and_monitor
           context:
             - hmpps-common-vars
-      - hmpps/deploy_env:
```
mahalrmoj, line 290: Why is dev going from the ci config?

sldblog: I duplicated it in a previous PR thinking we would do a gradual switchover

This PR removes deploy_dev_newcluster and keeps deploy_dev as one deploy to the new cluster instead

mahalrmoj (approved):

PR #720, opened 2021-12-01
Access Delivery Session Appointments

andrewburrowsmoj:
This PR is to facilitate the ability to show all appointments across all sessions on referral progress screen.

Currently we only show the latest appointment for a session and the existing endpoints only consider the latest; furthermore the endpoints tie a session to an action plan rather than referral.

Also need the ability to get an appointment based on appointment id rather than session id since session id only makes sense if you want the latest appointment.

commit 811aaa179025cbb368e77428853554adb50573e9

Add new endpoint /referral/{referralId}/delivery-session-appointment/{appointmentId}

Allows for access to an appointment based on referral and delivery session appointment id rather than action plan.
Returns a DeliverySessionAppointment object which is different from a DeliverySession in that the id is the appointment.

The DeliverySession object contains the latest appointment for the session which can be a bit misleading as the id is the session id rather than the appointment id.

commit f109c44e41798e6fb4d645f995ba068e7df987c3
New endpoint GET /referral/{referralId}/delivery-session-appointments

This returns all appointments for all sessions of a referral; this includes appointments that have been rescheduled for the same session.
Existing endpoints connected with delivery session would only return the latest appointment for a session.

commit de13714de363169a6ec7e535e51407387f7b2dc9 
Added two new endpoints:

POST /referral/{referralId}/delivery-session-appointment - allows to scheduling a new appointment for a session provided that the existing appointments (if any) have feedback submitted
PATCH /referral/{referralId}/delivery-session-appointment/{appointmentId} - allows to reschedule an existing appointment provided that no feedback has been submitted

Deprecated the existing endpoint PATCH /action-plan/{id}/appointment/{sessionNumber}. This endpoint performs both scheduling and rescheduling based on action plan id and session number.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/controller/DeliverySessionController.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/controller/DeliverySessionController.kt
@@ -6,6 +6,7 @@ import org.springframework.web.bind.annotation.GetMapping
 import org.springframework.web.bind.annotation.PatchMapping
 import org.springframework.web.bind.annotation.PathVariable
 import org.springframework.web.bind.annotation.PostMapping
+import org.springframework.web.bind.annotation.PutMapping
 import org.springframework.web.bind.annotation.RequestBody
 import org.springframework.web.bind.annotation.RestController
 import org.springframework.web.server.ResponseStatusException
@@ -13,6 +14,7 @@ import uk.gov.justice.digital.hmpps.hmppsinterventionsservice.authorization.Refe
 import uk.gov.justice.digital.hmpps.hmppsinterventionsservice.authorization.UserMapper
 import uk.gov.justice.digital.hmpps.hmppsinterventionsservice.component.LocationMapper
 import uk.gov.justice.digital.hmpps.hmppsinterventionsservice.dto.DeliverySessionAppointmentDTO
+import uk.gov.justice.digital.hmpps.hmppsinterventionsservice.dto.DeliverySessionAppointmentScheduleDetailsDTO
 import uk.gov.justice.digital.hmpps.hmppsinterventionsservice.dto.DeliverySessionDTO
 import uk.gov.justice.digital.hmpps.hmppsinterventionsservice.dto.RecordAppointmentBehaviourDTO
 import uk.gov.justice.digital.hmpps.hmppsinterventionsservice.dto.UpdateAppointmentAttendanceDTO
@@ -34,6 +36,8 @@ class DeliverySessionController(
   val referralAccessChecker: ReferralAccessChecker,
   val referralService: ReferralService,
 ) {
+
+  @Deprecated("superseded by scheduleNewDeliverySessionAppointment and rescheduleDeliverySessionAppointment")
   @PatchMapping("/action-plan/{id}/appointment/{sessionNumber}")
   fun updateSessionAppointment(
     @PathVariable(name = "id") actionPlanId: UUID,
@@ -160,5 +164,57 @@ class DeliverySessionController(
     referralAccessChecker.forUser(referral, user)
     return deliverySessionService.getSessions(referralId)
       .flatMap { session -> session.appointments.map { appointment -> DeliverySessionAppointmentDTO.from(session.sessionNumber, appointment) } }
+  }
+
+  @PostMapping("/referral/{referralId}/delivery-session-appointments")
+  fun scheduleNewDeliverySessionAppointment(
+    @PathVariable(name = "referralId") referralId: UUID,
+    @RequestBody newDeliverySessionAppointmentRequest: DeliverySessionAppointmentScheduleDetailsDTO,
+    authentication: JwtAuthenticationToken,
+  ): DeliverySessionAppointmentDTO {
+    val user = userMapper.fromToken(authentication)
+    val referral = referralService.getSentReferral(referralId) ?: throw ResponseStatusException(
+      HttpStatus.BAD_REQUEST, "sent referral not found [referralId=$referralId]"
+    )
+    referralAccessChecker.forUser(referral, user)
+    val deliverySession = deliverySessionService.scheduleNewDeliverySessionAppointment(
+      referralId,
+      newDeliverySessionAppointmentRequest.sessionId,
+      newDeliverySessionAppointmentRequest.appointmentTime,
+      newDeliverySessionAppointmentRequest.durationInMinutes,
+      user,
+      newDeliverySessionAppointmentRequest.appointmentDeliveryType,
+      newDeliverySessionAppointmentRequest.sessionType,
+      newDeliverySessionAppointmentRequest.appointmentDeliveryAddress,
+      newDeliverySessionAppointmentRequest.npsOfficeCode,
+    )
+    return DeliverySessionAppointmentDTO.from(deliverySession.sessionNumber, deliverySession.currentAppointment!!)
+  }
+
+  @PutMapping("/referral/{referralId}/delivery-session-appointments/{appointmentId}")
+  fun rescheduleDeliverySessionAppointment(
+    @PathVariable(name = "referralId") referralId: UUID,
+    @PathVariable(name = "appointmentId") appointmentId: UUID,
+    @RequestBody newDeliverySessionAppointmentRequest: DeliverySessionAppointmentScheduleDetailsDTO,
+    authentication: JwtAuthenticationToken,
+  ): DeliverySessionAppointmentDTO {
+    val user = userMapper.fromToken(authentication)
+    val referral = referralService.getSentReferral(referralId) ?: throw ResponseStatusException(
+      HttpStatus.BAD_REQUEST, "sent referral not found [referralId=$referralId]"
+    )
+    referralAccessChecker.forUser(referral, user)
+    val deliverySession = deliverySessionService.rescheduleDeliverySessionAppointment(
+      referralId,
+      newDeliverySessionAppointmentRequest.sessionId,
+      appointmentId,
+      newDeliverySessionAppointmentRequest.appointmentTime,
+      newDeliverySessionAppointmentRequest.durationInMinutes,
+      user,
+      newDeliverySessionAppointmentRequest.appointmentDeliveryType,
+      newDeliverySessionAppointmentRequest.sessionType,
+      newDeliverySessionAppointmentRequest.appointmentDeliveryAddress,
+      newDeliverySessionAppointmentRequest.npsOfficeCode,
+    )
+    return DeliverySessionAppointmentDTO.from(deliverySession.sessionNumber, deliverySession.currentAppointment!!)
   }
 }
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/controller/DeliverySessionController.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/controller/DeliverySessionController.kt
@@ -1,5 +1,6 @@
 package uk.gov.justice.digital.hmpps.hmppsinterventionsservice.controller
 
+import org.springframework.http.HttpStatus
 import org.springframework.security.oauth2.server.resource.authentication.JwtAuthenticationToken
 import org.springframework.web.bind.annotation.GetMapping
 import org.springframework.web.bind.annotation.PatchMapping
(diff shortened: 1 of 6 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/controller/DeliverySessionController.kt
```diff
@@ -118,4 +126,22 @@ class DeliverySessionController(
     val user = userMapper.fromToken(authentication)
     return DeliverySessionDTO.from(deliverySessionService.submitSessionFeedback(actionPlanId, sessionNumber, user))
   }
+
+  @GetMapping("/referral/{referralId}/delivery-session-appointment/{appointmentId}")
```
andrewburrowsmoj, line 136: make this resource noun plural

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/controller/DeliverySessionController.kt
```diff
@@ -161,4 +164,56 @@ class DeliverySessionController(
...
+      user,
+      newDeliverySessionAppointmentRequest.appointmentDeliveryType,
+      newDeliverySessionAppointmentRequest.sessionType,
+      newDeliverySessionAppointmentRequest.appointmentDeliveryAddress,
+      newDeliverySessionAppointmentRequest.npsOfficeCode,
+    )
+    return DeliverySessionAppointmentDTO.from(deliverySession.sessionNumber, deliverySession.currentAppointment!!)
+  }
+
+  @PatchMapping("/referral/{referralId}/delivery-session-appointment/{appointmentId}")
```
andrewburrowsmoj, line 193: this should be PUT

andrewburrowsmoj: make plural

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/controller/DeliverySessionController.kt
```diff
@@ -161,4 +164,56 @@ class DeliverySessionController(
     return deliverySessionService.getSessions(referralId)
       .flatMap { session -> session.appointments.map { appointment -> DeliverySessionAppointmentDTO.from(session.sessionNumber, appointment) } }
   }
+
+  @PostMapping("/referral/{referralId}/delivery-session-appointment")
```
andrewburrowsmoj, line 168: make plural

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/DeliverySessionService.kt
```diff
@@ -62,6 +62,84 @@ class DeliverySessionService(
...
+    appointmentDeliveryType: AppointmentDeliveryType,
+    appointmentSessionType: AppointmentSessionType,
+    appointmentDeliveryAddress: AddressDTO? = null,
+    npsOfficeCode: String? = null,
+  ): DeliverySession {
+    val session = getDeliverySession(referralId, sessionNumber) ?: throw EntityNotFoundException("Session not found for referral [referralId=$referralId, sessionNumber=$sessionNumber]")
+    val existingAppointment = session.currentAppointment
+
+    existingAppointment?.let {
+      if (it.appointmentTime.isAfter(appointmentTime)) {
```
andrewburrowsmoj, line 80: should this be in reschedule instead?

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/controller/DeliverySessionController.kt
andrewburrowsmoj: done

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/controller/DeliverySessionController.kt
andrewburrowsmoj: done

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/controller/DeliverySessionController.kt
andrewburrowsmoj: done

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/controller/DeliverySessionController.kt
```diff
@@ -118,4 +133,88 @@ class DeliverySessionController(
...
+    @PathVariable(name = "referralId") referralId: UUID,
+    @PathVariable(name = "appointmentId") appointmentId: UUID,
+    authentication: JwtAuthenticationToken,
+  ): DeliverySessionAppointmentDTO {
+    val user = userMapper.fromToken(authentication)
+    val referral = referralService.getSentReferral(referralId) ?: throw ResponseStatusException(
+      HttpStatus.BAD_REQUEST, "sent referral not found [referralId=$referralId]"
+    )
+    referralAccessChecker.forUser(referral, user)
+    val matchingAppointment = deliverySessionService.getSessions(referralId)
```
mahalrmoj, line 148: Is this not model/business related logic and hence reusable as a service method?

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/controller/DeliverySessionController.kt
```diff
@@ -118,4 +133,88 @@ class DeliverySessionController(
...
+    val referral = referralService.getSentReferral(referralId) ?: throw ResponseStatusException(
+      HttpStatus.BAD_REQUEST, "sent referral not found [referralId=$referralId]"
+    )
+    referralAccessChecker.forUser(referral, user)
+    return deliverySessionService.getSessions(referralId)
+      .flatMap { session -> session.appointments.map { appointment -> DeliverySessionAppointmentDTO.from(session.sessionNumber, appointment) } }
+  }
+
+  @PostMapping("/referral/{referralId}/delivery-session-appointments")
+  fun scheduleNewDeliverySessionAppointment(
```
mahalrmoj, line 170: Would the word "initial" be clearer as "new" appointments are event created when rescheduling

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/dto/DeliverySessionAppointmentDTO.kt
```diff
@@ -0,0 +1,55 @@
...
+  val npsOfficeCode: String?,
+  val appointmentDeliveryAddress: AddressDTO?,
+  val sessionFeedback: SessionFeedbackDTO,
+) {
+  companion object {
+    fun from(sessionNumber: Int, appointment: Appointment): DeliverySessionAppointmentDTO {
+      val appointmentDelivery = appointment.appointmentDelivery
+      val address = when (appointmentDelivery?.appointmentDeliveryType) {
+        AppointmentDeliveryType.IN_PERSON_MEETING_OTHER -> {
+          if (appointmentDelivery?.appointmentDeliveryAddress !== null) {
```
mahalrmoj, line 25: should we be using ?.let syntax?

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/DeliverySessionService.kt
```diff
@@ -61,6 +62,84 @@ class DeliverySessionService(
     return deliverySessionRepository.save(session)
   }
 
+  fun scheduleNewDeliverySessionAppointment(
```
mahalrmoj, line 65: Again "initial" over "new"?

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/DeliverySessionService.kt
```diff
@@ -61,6 +62,84 @@ class DeliverySessionService(
...
+    appointmentDeliveryType: AppointmentDeliveryType,
+    appointmentSessionType: AppointmentSessionType,
+    appointmentDeliveryAddress: AddressDTO? = null,
+    npsOfficeCode: String? = null,
+  ): DeliverySession {
+    val session = getDeliverySession(referralId, sessionNumber) ?: throw EntityNotFoundException("Session not found for referral [referralId=$referralId, sessionNumber=$sessionNumber]")
+    val existingAppointment = session.currentAppointment
+
+    existingAppointment?.let {
+      if (it.appointmentTime.isAfter(appointmentTime)) {
```
mahalrmoj, line 80: Why is this necessary? Isn't it covered by the check that the latest appointment has feedback

mahalrmoj (approved):

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/controller/DeliverySessionController.kt
andrewburrowsmoj: This is either the first (delivery session) appointment or any additional appointment.
If additional then it checks that the existing appointment is before this new one and that the feedback was submitted.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/DeliverySessionService.kt
andrewburrowsmoj: No because the time given for this new appointment could be before the existing one even though feedback was submitted.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/controller/DeliverySessionController.kt
andrewburrowsmoj: In this case we are transforming the `Session`s' list of appointments into a list of appointment and sessionNumber pairs.

It's not possible to delegate this functionality into the service method since the sessionNumber is not retrievable from an `Appointment` object. We can't know which `Session` the retrieved `Appointment` belongs to without using some type of wrapper object; an `Appointment` has no concept of whether it's an initial assessment or a delivery session appointment.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/dto/DeliverySessionAppointmentDTO.kt
andrewburrowsmoj: Thanks. I've used this syntax now.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/controller/DeliverySessionController.kt
```diff
@@ -118,4 +133,88 @@ class DeliverySessionController(
...
+    authentication: JwtAuthenticationToken,
+  ): DeliverySessionAppointmentDTO {
+    val user = userMapper.fromToken(authentication)
+    val referral = referralService.getSentReferral(referralId) ?: throw ResponseStatusException(
+      HttpStatus.BAD_REQUEST, "sent referral not found [referralId=$referralId]"
+    )
+    referralAccessChecker.forUser(referral, user)
+    val matchingAppointment = deliverySessionService.getSessions(referralId)
+      .flatMap { session -> session.appointments.map { appointment -> Pair(session.sessionNumber, appointment) } }
+      .firstOrNull { appointment -> appointment.second.id == appointmentId }
```
sldblog, line 150: @andrewburrowsmoj @mahalrmoj In non-Boot environments, I've seen this pattern dealt by using custom queries in the service, which feels like a more efficient approach.

I don't quite understand the need for this endpoint though -- `/referral/{id}/sessions/{sessionNumber}` seems ok and doesn't use action plans. Is this endpoint necessary?

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/controller/DeliverySessionController.kt
```diff
@@ -118,4 +133,88 @@ class DeliverySessionController(
...
+    @PathVariable(name = "referralId") referralId: UUID,
+    authentication: JwtAuthenticationToken,
+  ): List<DeliverySessionAppointmentDTO> {
+    val user = userMapper.fromToken(authentication)
+    val referral = referralService.getSentReferral(referralId) ?: throw ResponseStatusException(
+      HttpStatus.BAD_REQUEST, "sent referral not found [referralId=$referralId]"
+    )
+    referralAccessChecker.forUser(referral, user)
+    return deliverySessionService.getSessions(referralId)
+      .flatMap { session -> session.appointments.map { appointment -> DeliverySessionAppointmentDTO.from(session.sessionNumber, appointment) } }
```
sldblog, line 166: `deliverySessionService.getSessionAppointments(referralId)` sounds like something that wouldn't force `deliverySessionService` users to transform on their own -- is this a smell that our entity/data model isn't fit for purpose?

PR #721, opened 2021-12-01
Making call to ARN before CommunityAPI when sending referral as ARN is more likely to fail via a timeout

mahalrmoj:
https://trello.com/c/KwbF414e/183-send-referral-submitadditionalriskinfo-before-calling-community-api-delius-api

Making call to ARN before CommunityAPI when sending referral as ARN is more likely to fail via a timeout

This avoids creating duplicate NSIs in nDelius on user retry.


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/ReferralService.kt
```diff
@@ -210,13 +210,13 @@ class ReferralService(
     referral.sentBy = authUserRepository.save(user)
     referral.referenceNumber = generateReferenceNumber(referral)
 
+    submitAdditionalRiskInformation(referral, user)
```
tomsmyers, line 213: would be good to have a comment about why the order is important

tomsmyers (approved):
👍

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/ReferralService.kt
```diff
@@ -214,9 +214,13 @@ class ReferralService(
...
      * and Delius/ARN. Once the asynchronous link is implemented, this feature can be turned off, and instead
      * Delius/ARN will be notified via the normal asynchronous route
+     *
+     * Order is important: ARN is more likely to fail via a timeout and thus when repeated by the user a
+     * duplicate NSI gets created in Delius as the Delius call is not idempotent. This order avoids creating
+     * duplicate NSIs in nDelius on user retry.
      */
-    communityAPIReferralService.send(referral)
     submitAdditionalRiskInformation(referral, user)
+    communityAPIReferralService.send(referral)
```
sldblog, line 223: might be useful to leverage APIs in Kotlin to communicate the order:
```kt
submitAdditionalRiskInformation(referral, user)
  .also { communityAPIReferralService.send(referral) }
```

PR #723, opened 2021-12-02
Change action plan errors into client errors

sldblog:

Use our `ValidationError` instead of a library `ValidationException`

`ValidationException` bubbles up as a 500 Internal Server Error, while
`ValidationError` is handled as a 400 Bad Request

It is the latter that we want, as the user can refresh and retry

Sentry issue: [INTERVENTIONS-SERVICE-PROD-7M](https://sentry.io/organizations/ministryofjustice/issues/2808132370/?referrer=github_integration)

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/ActionPlanService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/ActionPlanService.kt
@@ -3,6 +3,7 @@ package uk.gov.justice.digital.hmpps.hmppsinterventionsservice.service
 import org.springframework.data.repository.findByIdOrNull
 import org.springframework.stereotype.Service
 import uk.gov.justice.digital.hmpps.hmppsinterventionsservice.component.ActionPlanValidator
+import uk.gov.justice.digital.hmpps.hmppsinterventionsservice.config.ValidationError
 import uk.gov.justice.digital.hmpps.hmppsinterventionsservice.events.ActionPlanEventPublisher
 import uk.gov.justice.digital.hmpps.hmppsinterventionsservice.jpa.entity.ActionPlan
 import uk.gov.justice.digital.hmpps.hmppsinterventionsservice.jpa.entity.ActionPlanActivity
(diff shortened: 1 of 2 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/ActionPlanService.kt
```diff
@@ -106,13 +106,13 @@ class ActionPlanService(
 
   private fun verifySafeForApproval(actionPlan: ActionPlan) {
     if (actionPlan.approvedAt != null) {
-      throw ValidationException("Action plan has already been approved. [id=${actionPlan.id}]")
+      throw ValidationError("Action plan has already been approved", listOf())
```
andrewburrowsmoj, line 109: would it not be useful to still include the actionplan id? 
Or are we certain that it's easy to obtain?

andrewburrowsmoj (approved):

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/ActionPlanService.kt
sldblog: Ah right, I forgot to mention the reasons behind this. Thanks for reminding me

I was thinking...

- In this particular case, we will know the context of the error is `/action-plan/{id}/approve`, which already gives us enough context
- This error is mapped into a client error, so it'll be coming back on the wire as `{"status": 400, "summary": "validation error", "message": "Action plan has already been approved", "userMessage": null, "validationErrors": [], "accessErrors": null}` which I assumed would be weird if it contained IDs -- would "message" be displayed?

sldblog: @andrewburrowsmoj ^

andrewburrowsmoj: I understand - you're right it makes it more user friendly this way and like you say they can infer it.
I was thinking this was more for our purposes for monitoring rather than api response.

PR #727, opened 2021-12-08
Split supplier assessment into schedule/reschedule

andrewburrowsmoj:
The purpose of this PR is to unify the supplier assessment appointment endpoints with the delivery session appointment endpoints.

This splits the existing scheduleOrRescheduleSupplierAssessmentAppointment into scheduleSupplierAssessmentAppointment and rescheduleSupplierAssessmentAppointment.

This makes the api more consistent and less confusing; but mainly it also provides greater control on validation of user expectations.

commit 9863aaf9122823231cb7369a57261af8fa0307e2

new endpoint POST /referral/{referralId}/supplier-assessment which schedules a completely new supplier assessment appointment.

The existing endpoint PUT /supplier-assessment/{id}/schedule-appointment does both schedule and reschedule depending on the state of the existing appointment.

The reasons for splitting is so the caller's intention are known explicitly and further validation can be performed on the request.

commit 699488dc2a110ef93acafb9d706f6ad7a8c46e50 

Added new endpoint PUT /referral/{referralId}/supplier-assessment/{appointmentId} which rescheuldes the existing appointment based on appointmentId.
This is split out from the existing endpoint PUT /supplier-assessment/{id}/schedule-appointment which reschedules the latest appointment if it exists or creates a new appointment if it does not exist.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/controller/SupplierAssessmentController.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/controller/SupplierAssessmentController.kt
@@ -30,6 +30,7 @@ class SupplierAssessmentController(
   private val appointmentValidator: AppointmentValidator,
   private val appointmentService: AppointmentService,
 ) {
+
   @PutMapping("/supplier-assessment/{id}/schedule-appointment")
   fun updateSupplierAssessmentAppointment(
     @PathVariable id: UUID,
@@ -51,6 +52,26 @@ class SupplierAssessmentController(
         updateAppointmentDTO.npsOfficeCode,
       )
     )
+  }
+
+  @PostMapping("/referral/{referralId}/supplier-assessment")
+  fun scheduleSupplierAssessmentAppointment(
+    @PathVariable referralId: UUID,
+    @RequestBody updateAppointmentDTO: UpdateAppointmentDTO,
+    authentication: JwtAuthenticationToken,
+  ): AppointmentDTO {
+    val user = userMapper.fromToken(authentication)
+    appointmentValidator.validateUpdateAppointment(updateAppointmentDTO)
+    val appointment = supplierAssessmentService.scheduleNewSupplierAssessmentAppointment(
+      referralId,
+      updateAppointmentDTO.durationInMinutes,
+      updateAppointmentDTO.appointmentTime,
+      user,
+      updateAppointmentDTO.appointmentDeliveryType,
+      updateAppointmentDTO.appointmentDeliveryAddress,
+      updateAppointmentDTO.npsOfficeCode
+    )
+    return AppointmentDTO.from(appointment)
   }
 
   @PutMapping("/referral/{referralId}/supplier-assessment/record-behaviour")
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/controller/SupplierAssessmentController.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/controller/SupplierAssessmentController.kt
@@ -31,6 +31,7 @@ class SupplierAssessmentController(
   private val appointmentService: AppointmentService,
 ) {
 
+  @Deprecated("superseded by POST /referral/{referralId}/supplier-assessment and PUT /referral/{referralId}/supplier-assessment/{appointmentId}")
   @PutMapping("/supplier-assessment/{id}/schedule-appointment")
   fun updateSupplierAssessmentAppointment(
     @PathVariable id: UUID,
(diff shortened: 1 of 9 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/controller/SupplierAssessmentController.kt
```diff
@@ -53,6 +55,48 @@ class SupplierAssessmentController(
...
+      updateAppointmentDTO.appointmentTime,
+      user,
+      updateAppointmentDTO.appointmentDeliveryType,
+      updateAppointmentDTO.appointmentDeliveryAddress,
+      updateAppointmentDTO.npsOfficeCode
+    )
+    return AppointmentDTO.from(appointment)
+  }
+
+  @PostMapping("/referral/{referralId}/supplier-assessment/{appointmentId}")
```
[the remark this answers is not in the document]
andrewburrowsmoj: Good spot - I meant to have it as PUT originally (it's even in the git commit comments).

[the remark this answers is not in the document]
andrewburrowsmoj: updated

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/AppointmentService.kt
```diff
@@ -290,4 +293,59 @@ class AppointmentService(
...
+    appointmentType: AppointmentType,
+    durationInMinutes: Int,
+    appointmentTime: OffsetDateTime,
+    appointmentDeliveryType: AppointmentDeliveryType,
+    appointmentSessionType: AppointmentSessionType,
+    appointmentDeliveryAddress: AddressDTO? = null,
+    npsOfficeCode: String? = null,
+  ): Appointment {
+    val appointment = this.appointmentRepository.findById(appointmentId).orElseThrow { EntityNotFoundException("Appointment not found [appointmentId=$appointmentId]") }
+    appointment.appointmentFeedbackSubmittedAt ?.let { throw ResponseStatusException(HttpStatus.BAD_REQUEST, "Appointment has already been delivered [appointmentId=$appointmentId]") }
```
sldblog, line 340: 🙋‍♂️ Throwing HTTP specific response codes in a `Service` feels like overreaching -- if we wanted to use this method in other business logic, those non-controller functions end up dealing with http-related unchecked exceptions -- might be a relevant here to use checked exceptions?

(Also wondering if this fits better "The server cannot or will not process the request due to something that is perceived to be a client error" ([400 Bad Request](https://httpstatuses.com/400)) or "The request could not be completed due to a conflict with the current state of the target resource. This code is used in situations where the user might be able to resolve the conflict and resubmit the request" ([409 Conflict](https://httpstatuses.com/409)))

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/SupplierAssessmentService.kt
```diff
@@ -63,6 +65,80 @@ class SupplierAssessmentService(
...
+    }
+    val appointment = appointmentService.scheduleNewAppointment(
+      sentReferral.id,
+      SUPPLIER_ASSESSMENT,
+      durationInMinutes,
+      appointmentTime,
+      createdByUser,
+      appointmentDeliveryType,
+      // Supplier assessment is always ONE_TO_ONE
+      AppointmentSessionType.ONE_TO_ONE,
```
sldblog, line 100: (Thinking about how to make the comment enforceable in code/regressions)

Feels like this is a good thing to encode into the preconditions of `scheduleNewAppointment` with `if appointmentType == SUPPLIER_ASSESSMENT and appointmentDeliveryType != ONE_TO_ONE then Error` and create a test for it

sldblog (approved):
Sorry, I should've reviewed fully before. Couple of questions -- to be honest, I'm OK with this to be merged and iterated separately

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/controller/SupplierAssessmentController.kt
```diff
@@ -30,6 +30,8 @@ class SupplierAssessmentController(
   private val appointmentValidator: AppointmentValidator,
   private val appointmentService: AppointmentService,
 ) {
+
+  @Deprecated("superseded by scheduleSupplierAssessmentAppointment and rescheduleSupplierAssessmentAppointment")
```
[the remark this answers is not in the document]
andrewburrowsmoj: yeah, I agree that endpoints would be better than function names. I've ammeded the commit.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/AppointmentService.kt
andrewburrowsmoj: Agreed. Now using EntityAlreadyExistsException and ValidationError for 409 and 400 respectively.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/SupplierAssessmentService.kt
andrewburrowsmoj: Agreed. I've added this validation into SupplierAsessmentService

PR #728, opened 2021-12-08
propagate appinsights trace context in events

tomsmyers:
the internal event processing happens outside the request context, so we first have to propagate the trace through the event publisher/listener mechanism

`TracePropagatingEventPublisher` handles this, and all the changes to classes like `ReferralEventPublisher` are just changing to inherit from this new class

on the other side `TracePropagatingEventListener` extracts the trace context, ands runs the existing listener code in a new trace span (`context.makeCurrent().use {`)

all the changes in `SNSService` are just moving existing listener code into new methods so we can utilize a generic `TraceableEvent` listener for all of them

next, `SNSPublisher` includes the opentracing headers as string message attributes. the headers looks like this:

```
"traceparent": {"Type": "String", "Value": "00-52908ba66f5742caa2856e9707ce6428-50ed11e4dbdbd1c9-01"}, "tracestate": {"Type": "String", "Value": "ai-internal-sp=100"
```

lastly, app insights falls over itself and doesn't work properly if you use the agent and the old v2 web instrumention. the config changes turn off the latter.

all this means, the listener gets those message attributes in the message. i've yet to fully get it working end to end, because again, app insights doesn't do what it says it does. but i'm confident we can make it work. when i got it working using the opentelemetry agent it auto-parsed the newly included message attributes. app insights in theory uses the same instrumentation, but it's not working OOTB.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -43,6 +43,7 @@ dependencies {
   implementation("io.sentry:sentry-spring-boot-starter:5.4.3")
   implementation("io.sentry:sentry-logback:5.4.3")
   implementation("io.github.microutils:kotlin-logging-jvm:2.1.15")
+  implementation("io.opentelemetry:opentelemetry-api:1.6.0")
 
   // openapi
   implementation("org.springdoc:springdoc-openapi-ui:1.5.13")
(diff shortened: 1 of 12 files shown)


tomsmyers:
closing for now - i never fully got this working e2e with the downstream instrumentation and i've let it get stale.

PR #734, opened 2021-12-10
Replace snyk scannning with trivy scanning

sldblog:

Replace snyk scannning with trivy scanning

As an added bonus, now with Slack alerts :)

📝 Names have been retained for minimal impact on branch "required" checks

Snyk has been replaced with Trivy as it's faster, smaller, and better integrated.


mahalrmoj (approved):
LGTM - the more the better!

PR #736, opened 2021-12-10
CI: Replace custom veracode policy scan code with orb

sldblog:

Replace custom veracode policy scan code with orb

To simplify the CI configuration file and rely on similar behaviour
across all projects


mahalrmoj (approved):
LGTM

PR #754, opened 2021-12-16
check attended field is included for past appointments

rgforsyth:

Ensures the 'attended' field is included for past appointments.

To prevent past appointments being created without feedback.


andrewburrowsmoj (approved):
lgtm

PR #758, opened 2021-12-17
bug fix: appointment in the past

andrewburrowsmoj:
Fixed bug with update Delivery session appointment where past appointment feedback details are ignored if there already is a scheduled appointment.


tomsmyers (approved):
LGTM - we should revisit `setAttendanceAndBehaviourIfHistoricAppointment` to actually do something different if this is not a historic appointment

PR #768, opened 2021-12-20
Decouple transaction from asynch events for appointment feedback event

andrewburrowsmoj:
There is a bug when creating a new delivery session appointment along with session feedback.

The asynchronous events that are triggered on feedback sumission are unable to obtain the current appointment because it's not been persisted at the time of execution as it's an asychronous event.

We could persist the appointment before sending the event - thereby it would be able to extract the appointment; however we are sending the delivery session entity to the event handler.

There is a possibility that the transaction may have been closed when handling the event, so any lazy loading would fail with Hibernate exceptions.

A way around this is to transform any necessary entities into DTOs - this has a significant knock on effect, so all code changes in this commit are connected to this specific goal.


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/dto/AppointmentDTO.kt
```diff
@@ -29,14 +29,7 @@ data class AppointmentDTO(
...
         durationInMinutes = appointment.durationInMinutes,
-        sessionFeedback = SessionFeedbackDTO.from(
-          appointment.attended,
-          appointment.additionalAttendanceInformation,
-          appointment.attendanceBehaviour,
-          appointment.notifyPPOfAttendanceBehaviour,
-          appointment.appointmentFeedbackSubmittedAt != null,
-          appointment.appointmentFeedbackSubmittedBy,
-        ),
+        sessionFeedback = SessionFeedbackDTO.from(appointment),
```
tomsmyers, line 32: this is great - thanks

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/CommunityAPIService.kt
```diff
@@ -29,8 +31,15 @@ interface CommunityAPIService {
...
 
+  fun getNotes(referral: SentReferralDTO, url: String, description: String, contractTypeName: String, primeProviderName: String): String {
+    return "$description for $contractTypeName Referral ${referral.referenceNumber} with Prime Provider $primeProviderName\n$url"
+  }
+
   fun setNotifyPPIfRequired(appointment: Appointment) =
     NO == appointment.attended!! || appointment.notifyPPOfAttendanceBehaviour ?: false
+
+  fun setNotifyPPIfRequired(deliverySession: DeliverySessionDTO) =
+    NO == deliverySession.sessionFeedback.attendance.attended!! || deliverySession.sessionFeedback.behaviour.notifyProbationPractitioner ?: false
```
tomsmyers, line 42: i'm not sure what the elvis operator is doing here. this should be a plain boolean expression.

```
NO == deliverySession.sessionFeedback.attendance.attended!! || deliverySession.sessionFeedback.behaviour.notifyProbationPractitioner == true
```

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/CommunityAPIService.kt
```diff
@@ -29,8 +31,15 @@ interface CommunityAPIService {
...
   }
 
+  fun getNotes(referral: SentReferralDTO, url: String, description: String, contractTypeName: String, primeProviderName: String): String {
+    return "$description for $contractTypeName Referral ${referral.referenceNumber} with Prime Provider $primeProviderName\n$url"
+  }
+
   fun setNotifyPPIfRequired(appointment: Appointment) =
     NO == appointment.attended!! || appointment.notifyPPOfAttendanceBehaviour ?: false
+
+  fun setNotifyPPIfRequired(deliverySession: DeliverySessionDTO) =
```
tomsmyers, line 41: we can get rid of the null checking here and simplify to plain boolean expressions

```
  fun setNotifyPPIfRequired(appointment: Appointment): Boolean =
    appointment.attended == NO || appointment.notifyPPOfAttendanceBehaviour == true
 ```
 
 ```
  fun setNotifyPPIfRequired(deliverySession: DeliverySessionDTO): Boolean =
    deliverySession.sessionFeedback.attendance.attended == NO || deliverySession.sessionFeedback.behaviour.notifyProbationPractitioner == true
```

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/DeliverySessionService.kt
```diff
@@ -279,13 +279,13 @@ class DeliverySessionService(
     appointment.appointmentFeedbackSubmittedBy = authUserRepository.save(submitter)
     deliverySessionRepository.save(session)
 
-    actionPlanAppointmentEventPublisher.attendanceRecordedEvent(session, appointment.attended!! == Attended.NO)
+    actionPlanAppointmentEventPublisher.attendanceRecordedEvent(session)
```
tomsmyers, line 282: this is way cleaner without the notify pp logic.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/notifications/CaseNotesNotificationsService.kt
```diff
@@ -43,7 +43,7 @@ class CaseNotesNotificationsService(
     val senderIsResponsibleOfficer =
       userTypeChecker.isProbationPractitionerUser(sender) &&
         (
-          (responsibleOfficer.authUser?.let { it == sender } ?: false) ||
+          (responsibleOfficer.authUser?.let { it.authSource == sender.authSource && it.username == sender.userName } ?: false) ||
```
tomsmyers, line 46: this shouyld be handled in the equals method of `AuthUser`. we shouldn't duplicate it here.

tomsmyers (approved):
great work - thank you. we should do a full e2e test of all affected code paths in dev before pushing live

PR #771, opened 2021-12-22
Bugfix: Nullpointer on NotifyActionPlanAppointmentService

andrewburrowsmoj:
This is a bug fix of an issue whereby a null pointer is thrown from NotifyActionPlanAppointmentService because the event contains a null session feedback. The reason it is null is because the DeliverySession is unable to reference the latest appointment as it is not persisted at the time of publishing.

To fix this bug we rearrange the order of calls to ensure that setting session feedback is last.

_Required._

_Required._

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/DeliverySessionService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/DeliverySessionService.kt
@@ -170,29 +170,28 @@ class DeliverySessionService(
       attended,
       notifyProbationPractitioner,
     )
-    if (existingAppointment == null) {
-      val appointment = Appointment(
-        id = UUID.randomUUID(),
-        createdBy = authUserRepository.save(updatedBy),
-        createdAt = OffsetDateTime.now(),
-        appointmentTime = appointmentTime,
-        durationInMinutes = durationInMinutes,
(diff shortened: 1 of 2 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/DeliverySessionService.kt
```diff
@@ -170,29 +170,27 @@ class DeliverySessionService(
...
+      durationInMinutes = durationInMinutes,
+      deliusAppointmentId = deliusAppointmentId,
+      referral = session.referral,
+    )
+    appointmentService.createOrUpdateAppointmentDeliveryDetails(appointment, appointmentDeliveryType, appointmentSessionType, appointmentDeliveryAddress, npsOfficeCode)
+    appointmentRepository.saveAndFlush(appointment)
+    session.appointments.add(appointment)
+    deliverySessionRepository.save(session)
+    return deliverySessionRepository.save(session).also {
       setAttendanceAndBehaviourIfHistoricAppointment(session, appointment, appointmentTime, attended, additionalAttendanceInformation, behaviourDescription, notifyProbationPractitioner, updatedBy)
```
tomsmyers, line 192: i think this would benefit from a comment which explains why it's laid out like this

tomsmyers (approved):

PR #773, opened 2021-12-22
Refactor the code in `submitAdditionalRiskInformation`

tomsmyers:
there are two main reasons

1. it wasn't really clear exactly what was going on in the two different branches. this
   is now cleanly split with a single condition.
2. there was actually a bug in the time we used for `riskSubmittedAt`; for the new oasys
   risk information we were still using the updated time from the old additional risk form.

_Required._

_Required._


andrewburrowsmoj (approved):
LGTM

PR #774, opened 2021-12-22
Seed new contract types for Greater Manchester

sldblog:

Adds Mentoring, Dependency and recovery and Women's support services as
those services are not currently running in Commissioned Rehabilitative
Services.

There is a slight deviation in desired outcomes for "Family and
significant others" in Greater Manchester services: a new outcome has to
be added.

To facilitate this with the least changes, we decided to copy the
"Family and significant others" service category as a new variation and
add the extra outcome on there.

**Consequences for data analysis**
Outcomes that are shared between the two categories will have the same
description but different IDs. This means aggregation on "Family and
significant others" outcomes becomes more tedious.

This is the first step to onboard Greater Manchester services.


tomsmyers (approved):
LGTM

PR #775, opened 2021-12-22
Gracefully handle 409s from arn

tomsmyers:

If a risk already exists on ARN, check that the createdDate matches the 'last updated' timestamp on the risk information we are trying to send. if it's the same, we don't need to do anything. if it's different, we have no way to proceed, so we throw an error just like we used to.

fix referrals that are 'stuck' because ARN is returning 409s.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/RisksAndNeedsService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/RisksAndNeedsService.kt
@@ -49,6 +49,7 @@ data class RedactedRisk(
 
 data class SupplementaryRiskResponse(
   val supplementaryRiskId: UUID,
+  val createdDate: OffsetDateTime,
 )
 
 @Service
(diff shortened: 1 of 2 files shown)


andrewburrowsmoj (approved):
ah, I assume this happens because the old referrals have already posted to arn but the new way it does it at the very end.

LGTM

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/RisksAndNeedsService.kt
```diff
@@ -92,22 +93,26 @@ class RisksAndNeedsService(
...
+      if (response.body.createdDate != riskCreatedAt) {
+        logger.error(
+          "attempted to update an existing supplementary risk with new data {} {} {} {}",
+          kv("crn", crn),
+          kv("referralId", referralId),
+          kv("riskId", response.body.supplementaryRiskId),
+          kv("riskCreatedAt", riskCreatedAt)
+        )
+
+        throw WebClientResponseException(409, "Conflict from ARN 'createSupplementaryRisk'", null, null, null)
```
sldblog, line 112: Do we expect the user to be able to resolve the conflict and resubmit the request in this case?

tomsmyers: no - there's nothing they can do. which is the situation we are in now, regardless of whether or not they have changed the information they are trying to submit.

tomsmyers: i think it's unlikely the risk information will have changed for most cases. so hoping that this unblocks most people with referrals stuck in this state.

tomsmyers: we will now get sentry alerts for these cases too

PR #790, opened 2022-01-05
Only require feedback for appointments that happened on a previous day.

tomsmyers:

Only require feedback for appointments that happened on a previous day.

This replicates the current behaviour of the API, allowing the existing UI
to continue to function as normal without any changes.

Once the UI is allowing appointments to be created in the past, we can revert
this change to ensure that any past appointment, even if it was on the current
day, requires that attendance information be passed along side the scheduling
information.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/validator/AppointmentValidator.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/validator/AppointmentValidator.kt
@@ -9,6 +9,7 @@ import uk.gov.justice.digital.hmpps.hmppsinterventionsservice.dto.UpdateAppointm
 import uk.gov.justice.digital.hmpps.hmppsinterventionsservice.jpa.entity.AppointmentDeliveryType
 import uk.gov.justice.digital.hmpps.hmppsinterventionsservice.jpa.entity.Attended
 import uk.gov.justice.digital.hmpps.hmppsinterventionsservice.jpa.entity.Attended.NO
+import java.time.LocalTime
 import java.time.OffsetDateTime
 
 @Component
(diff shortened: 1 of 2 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/validator/AppointmentValidator.kt
```diff
@@ -63,8 +64,11 @@ class AppointmentValidator {
     behaviourDescription: String?,
     errors: MutableList<FieldError>
   ) {
-    if (appointmentTime.isAfter(OffsetDateTime.now()))
+    // if the appointment occurred today or on a date in the future, no attendance validation is required
+    if (appointmentTime.isAfter(OffsetDateTime.now().with(LocalTime.MIN))) {
```
sldblog, line 68: Just musing: later perhaps we could also introduce a return value as currently both error and non-error cases use an empty `return`, making understanding the side-effect a little bit more effort to follow

sldblog (approved):
It feels like this should have a regression in case we're hijacked by priorities

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/validator/AppointmentValidator.kt
tomsmyers: on the regression thing - i'm not super bothered if this stays around. worst case is the SP is allowed to enter a session that happens earlier today and they're not forced to enter feedback right away. they are still contractually obliged to enter feedback on the same day, it's just a few more clicks.

but yeh i take your point that having a reminder would be beneficial. i think just tracking in trello is sufficient though.

tomsmyers: as a side note though - it sucks no unit tests started failing when i made this change.

tomsmyers: added unit test

PR #793, opened 2022-01-06
Revert "Gracefully handle 409s from arn"

tomsmyers:
Reverts ministryofjustice/hmpps-interventions-service#775


sldblog (approved):
👍 until ISO8601 compatibility is debugged

PR #794, opened 2022-01-06
🐛 Use datetimes without timezones from ARN

sldblog:

6a2b29a10c37b3244eeb6c5006f0e32378c5b739 introduced these changes but we got errors:

`com.fasterxml.jackson.databind.exc.InvalidFormatException: Cannot deserialize value of type 'java.time.OffsetDateTime' from String "2022-01-06T11:33:18.666979": Failed to deserialize java.time.OffsetDateTime: (java.time.format.DateTimeParseException) Text '2022-01-06T11:33:18.666979' could not be parsed at index 26
 at [Source: (io.netty.buffer.ByteBufInputStream); line: 1, column: 239] (through reference chain: uk.gov.justice.digital.hmpps.hmppsinterventionsservice.service.SupplementaryRiskResponse["createdDate"])`

This was due to ARN response having datetimes without time zones.

This change was reverted in #793 and this PR reinstates it with working time types.

Make conflict resolution in ARN work with the current payloads.


tomsmyers (approved):
thanks david

PR #800, opened 2022-01-12
Remove shared services from Manchester's women's services

sldblog:

- Remove shared services from Manchester's women's services
- Add Manchester as a PCC region to make it show up in the filter

All other Manchester services can now accept both male and female participants, meaning these offerings are no longer necessary under "Women's Support Services (GM)"

*Bonus* We were missing a Manchester PCC region, which meant the UI cannot filter for our new services.


rgforsyth (approved):

sldblog:
Thank you!

PR #832, opened 2022-02-09
log the number of referrals for both 'sent-referrals' endpoints (as well as some supplementary information)

tomsmyers:

log the number of referrals for both 'sent-referrals' endpoints (as well as some supplementary information)

so we can actually see how many referrals are being returned in average SP/PP dashboard lists


sldblog (approved):
LGTM

💭 We could join this up with `InterventionsAuthorizedProvider` custom events in AppInsights if these would also go into custom events, to draw a correlation between size + providers

sldblog:
I think the vulnerability problem is fixed on top of `main`

sldblog:
Merged -- the flagged vulnerability is already fixed on main, so in the interest of moving forward, I bypassed that requirement

PR #833, opened 2022-02-09
add paging params for GET /sent-referrals endpoint

tomsmyers:

adds a new endpoint `/sent-referrals/paged` that takes page query params (page, size, sort) and returns page objects of referrals instead of long lists.

allow UI to call the paged endpoint for large case lists


sldblog (approved):

sldblog:
_random thought_
I'd be kinda keen to separate the linter from the unit tests, the delay is a bit annoying

PR #834, opened 2022-02-09
Switched the extract table python to shell script

PriyaBasker23:

The table extraction functionality is switched to shell script from python 

Sort the memory issue caused by large data load.

### DIFF
--- a/helm_deploy/hmpps-interventions-service/templates/cronjob-data-extractor-analytics.yaml
+++ b/helm_deploy/hmpps-interventions-service/templates/cronjob-data-extractor-analytics.yaml
@@ -13,7 +13,7 @@ spec:
           - name: data-extractor-analytics
             image: ministryofjustice/data-engineering-data-extractor:v1
             imagePullPolicy: Always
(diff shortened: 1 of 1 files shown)


helm_deploy/hmpps-interventions-service/templates/cronjob-data-extractor-analytics.yaml
```diff
@@ -13,7 +13,7 @@ spec:
           - name: data-extractor-analytics
             image: ministryofjustice/data-engineering-data-extractor:v1
             imagePullPolicy: Always
-            args: ["extract_table_names.py && extract_pg_jsonl_snapshot.py && transfer_local_to_s3.sh"]
+            args: ["extract_psql_all_tables_to_jsonl.sh && extract_pg_jsonl_snapshot.py && transfer_local_to_s3.sh"]
```
sldblog, line 16: @PriyaBasker23 should `extract_pg_jsonl_snapshot.py` be removed as well?

PriyaBasker23: Thanks for spotting. removed it now. We dont need it

sldblog (approved):

PR #841, opened 2022-02-14
Deploy our Grafana dashboard to prod namespace

sldblog:

Deploy our Grafana dashboard to prod namespace

This makes sure it never gets lost and we can iterate on it (it's currently not in our deployment)


sldblog:
Fails `ci/circleci: validate_helm` with

`Error: parse error at (hmpps-interventions-service/templates/grafana-dashboard.yml:110): function "pod" not defined` -- this is due to Grafana defining labels on graphs with the same template syntax as helm:

```yaml
"legendFormat": "{{ pod }}",
```

sldblog:
Removed the file

PR #844, opened 2022-02-15
Log reference identifiers on referral submission

sldblog:

Log reference identifiers on referral submission

To support looking up an ID from any other ID without having to peek into the database directly

(Finding out the referral ID from a reference number is a typical task when dealing with support questions)

**Hypothesis**: dev time spent on support goes (slightly) 📉 down


src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/LogReferralIdsListenerTest.kt
```diff
@@ -0,0 +1,50 @@
...
+  fun `sent referral event logs all identifiers so that we can look up other IDs by searching the logs`() {
+    val mockReferral = mock<Referral>()
+    val sentEvent = ReferralEvent(
+      this,
+      ReferralEventType.SENT,
+      mockReferral,
+      "irrelevant-for-this-test"
+    )
+    LogReferralIdsListener().onApplicationEvent(sentEvent)
+    assertThat(memoryAppender.logEvents).isNotEmpty
```
sldblog, line 48: Hmm, I wasn't quite pedantic with this test...

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/LogReferralIdsListener.kt
```diff
@@ -0,0 +1,27 @@
...
+
+  override fun onApplicationEvent(event: ReferralEvent) {
+    if (event.type != ReferralEventType.SENT) {
+      return
+    }
+    logger.info(
+      "new referral sent, reference numbers:",
+      StructuredArguments.kv("id", event.referral.id),
+      StructuredArguments.kv("reference", event.referral.referenceNumber),
+      StructuredArguments.kv("crn", event.referral.serviceUserCRN)
```
sldblog, line 24: could also log the NSI ID or something

tomsmyers (approved):

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/LogReferralIdsListenerTest.kt
sldblog: Fixed in 3535477f33b4d556481f14babb30392ed18ab846

PR #848, opened 2022-02-18
Remove social inclusion from mentoring contract type

sldblog:

Remove social inclusion from mentoring contract type

In the versions of the onboarding spreadsheets, social inclusion was
mentioned as part of mentoring, but now it seems it does not have to be

As far as I can see, ongoing mentoring referrals with social inclusion outcomes will be
⛔ **blocked** if this goes live, as we validate that the categories given must
be part of the contract type ([`validateDraftReferralUpdate` in `ReferralService.kt`](https://github.com/ministryofjustice/hmpps-interventions-service/blob/2d74d12ac52529c01f4de99199f663384d5d0648/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/ReferralService.kt#L341-L347))

@tomsmyers @rgforsyth do you agree? Am I misinterpreting?

### DIFF
--- a/src/main/resources/db/migration/V1_93_3__remove_social_inclusion_from_mentoring.sql
+++ b/src/main/resources/db/migration/V1_93_3__remove_social_inclusion_from_mentoring.sql
@@ -1,0 +1,5 @@
+DELETE
+FROM contract_type_service_category
+-- Mentoring
+WHERE contract_type_id = 'bc65434d-089b-4c37-80c2-50efebb76933'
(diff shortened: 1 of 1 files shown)


sldblog:
Should we instead opt for end-dating these links? (New referrals can't use it, existing ones can)

tomsmyers:
i don't see any draft referrals for this contract with Social Inclusion selected so this is fine. 

Note: this is a draft referral for this contract with no service cats yet selected (as of 11am this morning). since i'm getting this data from preprod this could have changed.

tomsmyers (approved):

PR #853, opened 2022-03-03
update service provider performance report to relfect latest agreed changes

tomsmyers:

update service provider performance report to relfect latest agreed changes

https://docs.google.com/spreadsheets/d/16pfAf6xhpYAn10DBGNXspnA78cz6SCS8/edit#gid=282321295

who knows


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/reporting/serviceprovider/performance/PerformanceReportProcessor.kt
```diff
@@ -44,6 +43,12 @@ class PerformanceReportProcessor(
       endRequestedReason = referral.endRequestedReason?.description,
       eosrSubmittedAt = referral.endOfServiceReport?.submittedAt,
       concludedAt = referral.concludedAt,
+      referralEndState = referral.endState,
+      dateSupplierAssessmentBooked = saa?.firstAppointment?.createdAt,
+      dateSupplierAssessmentFirstScheduledFor = saa?.firstAppointment?.appointmentTime,
+      dateSupplierAssessmentFirstNotAttended = saa?.firstAppointmentWithNonAttendance?.appointmentTime,
+      dateSupplierAssessmentFirstAttended = saa?.firstAttendedAppointment?.appointmentTime,
+      supplierAssessmentAttendedOnTime = saa?.firstAttendedAppointment?.attended == Attended.YES,
```
sldblog, line 51: This would be true if someone did not attend -- in that case we should use `null` here

Can we add a test for that?

sldblog: Oh... This is `firstAttended...` so this won't happen. Never mind me!

tomsmyers: no actually i think you're right. `saa?.firstAttendedAppointment?.attended` resolves to `null` and `null == ...` is false. good spot.

tomsmyers:
> The reordering of columns might create a bit of headache for people who currently rely on ordering
> 
> That said, we never formally defined this schema so we can still consider it v0/under development so I think that's OK
> 
> 👍

yeh - i figured the removal of `initialAssessmentAttendedAt` constitues a reordering anyway so what harm is there grouping the existing field with new fields.

PR #864, opened 2022-03-10
Revert "update service provider performance report to relfect latest agreed changes"

tomsmyers:
Reverts ministryofjustice/hmpps-interventions-service#853


sldblog (approved):
👍 To unblock other work on main until this change is verified by user acceptance tests

PR #870, opened 2022-03-14
Enable selecting all API pods with a label

sldblog:

Enable selecting all API pods with a single label

In the `ServiceMonitor` resource we can tell prometheus to scrape `/prometheus` endpoint on all pods matching the label `app: hmpps-interventions-service`

Currently, the alternative would be to define matching labels for all `app`s

Note: `release: hmpps-interventions-service` does not work as the data dictionary is currently labelled with that too

### DIFF
--- a/helm_deploy/hmpps-interventions-service/templates/_helpers.tpl
+++ b/helm_deploy/hmpps-interventions-service/templates/_helpers.tpl
@@ -57,14 +57,17 @@ Selector labels
 */}}
 {{- define "app.selectorLabels" -}}
 app: {{ include "app.name" . }}
+section: api
 release: {{ .Release.Name }}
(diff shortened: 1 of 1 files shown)


sldblog:
`ci/circleci: validate_helm` failing with `for: "STDIN": Deployment.apps "hmpps-interventions-service" is invalid: spec.selector: Invalid value: v1.LabelSelector{MatchLabels:map[string]string{"app":"hmpps-interventions-service", "release":"hmpps-interventions-service", "section":"api"}, MatchExpressions:[]v1.LabelSelectorRequirement(nil)}: field is immutable`

`matchLabels` is immutable in the Kubernetes API.

This means we would need to recreate the `Deployment` resources to make this change.

PR #875, opened 2022-03-16
Patch CVE-2022-0778

sldblog:
## What does this pull request do?

Patch CVE-2022-0778 (upgrade openssl) by forcing `apk upgrade` to always run

## What is the intent behind these changes?

https://nvd.nist.gov/vuln/detail/CVE-2022-0778

PR #876, opened 2022-03-16
Make use of docker layer caching

sldblog:

Restructures the `Dockerfile` to make use of our docker layer caching.

In an event of a full cache hit (no dependency and no `src/main/` changes) the entire build step now finishes in a minute
image

In the event of a partial cache hit (no dependency updates), the entire build step would take about ~3 minutes I believe

## How?

`ADD . .` copies all files, including `.git/` directory

This forces `./gradlew clean assemble` to always be a fresh build on all commits

Since we have docker layer caching on, we can split this into 3 steps:

1. download gradle and most dependencies
2. compile the runtime classes
3. parse .git information and create the JAR file

Each step only copies the relevant files, making it possible to cache the `./gradlew` layer.

The `.git` folder changes on every single file change, which means we have to ignore `generateGitProperties` task. We only copy git files when it is the least costly thing to do.

To improve docker build times.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -24,6 +24,7 @@ RUN ./gradlew assemble
 FROM adoptopenjdk/openjdk11:alpine-jre
 LABEL maintainer="HMPPS Digital Studio <info@digital.justice.gov.uk>"
 
+# force a rebuild of `apk upgrade` below by invalidating the BUILD_NUMBER env variable on every commit
 ARG BUILD_NUMBER
 ENV BUILD_NUMBER ${BUILD_NUMBER:-1_0_0}
 


Dockerfile
```diff
@@ -28,7 +43,7 @@ RUN addgroup --gid 2000 --system appgroup && \
 WORKDIR /app
 COPY --from=builder --chown=appuser:appgroup /app/build/libs/hmpps-interventions-service*.jar /app/app.jar
 COPY --from=builder --chown=appuser:appgroup /app/build/libs/applicationinsights-agent*.jar /app/agent.jar
-COPY --from=builder --chown=appuser:appgroup /app/applicationinsights.json /app
+COPY --chown=appuser:appgroup applicationinsights.json /app
```
sldblog, line 46: It's OK to copy this file from the filesystem directly, and perhaps cleaner as well

Dockerfile
```diff
@@ -1,15 +1,30 @@
...
+# assemble extracts information from .git, this layer changes for all commits
+COPY . .
+RUN ./gradlew assemble
+
+
+# ---
 FROM adoptopenjdk/openjdk11:alpine-jre
 LABEL maintainer="HMPPS Digital Studio <info@digital.justice.gov.uk>"
 
+# force a rebuild of `apk upgrade` below by invalidating the BUILD_NUMBER env variable on every commit
```
sldblog, line 27: @tomsmyers ^ Related to https://github.com/ministryofjustice/hmpps-interventions-service/pull/875#pullrequestreview-911867663

tomsmyers (approved):

PR #879, opened 2022-03-17
Do not alert InvalidAssumptionError into Sentry

sldblog:

Removes the `InvalidAssumptionError` exception.

We have to monitor the occurrence of this longer-term, which is possible through AppInsights via [`customEvents | where cloud_RoleName == 'interventions-service' and name == 'InterventionsInvalidAssumption'`](https://portal.azure.com#@747381f4-e81f-4a43-bf68-ced6a1e14edf/blade/Microsoft_Azure_Monitoring_Logs/LogsBlade/resourceId/%2Fsubscriptions%2Fa5ddf257-3b21-4ba9-a28c-ab30f751b383%2FresourceGroups%2Fnomisapi-prod-rg%2Fproviders%2Fmicrosoft.insights%2Fcomponents%2Fnomisapi-prod/source/LogsBlade.AnalyticsShareLinkToQuery/q/H4sIAAAAAAAAA0suLS7Jz3UtS80rKVaoUSjPSC1KVUjOyS9NiQ%252FKz0n1S8xNVbC1VVDPzCtJLQKpyszPK9YtBrIzk1PVFRLzUhTyYGo8kdV45pUl5mSmOBYXl%252BYWgETUASZPoUZrAAAA/timespan/P1D)

🎯 Sprint goal: "Reduce noise in #interventions-alerts channel". This error is ignored, but occured 121 times over 14 days.

image

The `InvalidAssumptionError` exception is not an "unexpected situtation" and does not need an alert because we cannot immediately action it.

This creates noise in our exception monitoring: https://sentry.io/organizations/ministryofjustice/issues/2914226502/?project=5807819&referrer=slack

Removing the exception but keeping the telemetry will reduce noise.

## How?

`TelemetryService.reportInvalidAssumption(..., recoverable = false)` is used as a control flow; the exception is later caught again and used to branch how the data is fetched.

The change here removes the exception control flow, leading to

- No more sentry exceptions for this occurrence, while keeping the telemetry about it
- `CommunityAPIOffenderService.getResponsibleOfficer` can now return without an officer (also, 2 warnings fixed), indicating "no matching officer found"
- `getResponsibleProbationPractitioner` explicitly branches between `getContactableResponsibleOfficer` and `getContactableReferringOfficer` as a fallback
- added a new test to explicitly define what happens when the hmpps-auth call fails during `getContactableReferringOfficer` (we bubble up)

Amended test cases:

`getResponsibleProbationPractitioner`:
- uses responsible officer if they have an email
- uses the referring officer details if the community-api call fails
- uses the referring officer details if the responsible officer has no email
- uses the creator officer details if the responsible officer has no email and the referral is a draft
- aborts if it cannot retrieve responsible and referring officer details

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/config/ErrorConfiguration.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/config/ErrorConfiguration.kt
@@ -41,8 +41,6 @@ class ValidationError(override val message: String, val errors: List<FieldError>
 
 class AccessError(val user: AuthUser, override val message: String, val errors: List<String>) : RuntimeException(message)
 
-class InvalidAssumptionError(assumption: String) : RuntimeException("assumption proved invalid: $assumption")
-
 @JsonInclude(JsonInclude.Include.NON_NULL)
 data class ErrorResponse(
   val status: Int,
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/CommunityAPIOffenderService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/CommunityAPIOffenderService.kt
@@ -109,7 +109,7 @@ class CommunityAPIOffenderService(
       ?.staffIdentifier
   }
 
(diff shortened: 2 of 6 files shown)


sldblog:
Idea: Perhaps make it more visible for probation staff that responsible officer can't be found or contacted

sldblog:
As I am leaving HMPPS today, sadly, I don't have the capacity to tidy this up. Closing

PR #885, opened 2022-03-22
Building entity graph to reduce number of queries getting executed wh…

kth13:
…en the dashboard is loaded

improve the performance of "/sent-referrals/paged" service.

Reduces the number of queries executed when the dashboard page is landed

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/jpa/entity/DynamicFrameworkContract.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/jpa/entity/DynamicFrameworkContract.kt
@@ -8,8 +8,36 @@ import javax.persistence.JoinColumn
 import javax.persistence.JoinTable
 import javax.persistence.ManyToMany
 import javax.persistence.ManyToOne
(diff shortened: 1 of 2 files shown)


tomsmyers (approved):
this is a great approach - thank you!

tomsmyers:
do you think this may have a noticable effect on db performance? if so it may be worthwhile monitoring the stats as we roll this out to production.

you can see db CPU on this dashboard https://grafana.live.cloud-platform.service.justice.gov.uk/d/PyQ91ARnk/hmpps-refer-and-monitor-an-intervention?orgId=1&from=now-6h&to=now&var-namespace=hmpps-interventions-prod

but all RDS stats are on here [(but you have to filter by instance id - ours is ](https://grafana.live-1.cloud-platform.service.justice.gov.uk/d/VR46pmwWk/aws-rds?orgId=1&var-datasource=Cloudwatch&var-region=default&var-dbinstanceidentifier=cloud-platform-e4573fa1c83bbbb9&from=now-3d&to=now)

kth13:
@tomsmyers This is a first step towards fixing the performance of our new service. I believe the further steps along with this will definitely can improve the performance. 
 I will keep an eye on the performance too

PR #891, opened 2022-03-28
Write referral details to the new update-able table alongside the existing referral table

tomsmyers:

Write referral details to the new update-able table alongside the existing referral table

This change is step one in the roll out of referral details that are allowed to change
after a referral is sent. Currently the new table is write-only, but that means
that once we backfill the table using data from existing referrals, it will be then fully
up-to-date with no need to backfill anything again.


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/jpa/entity/ReferralDetails.kt
```diff
@@ -0,0 +1,24 @@
...
+import java.util.UUID
+import javax.persistence.Column
+import javax.persistence.Entity
+import javax.persistence.Id
+import javax.persistence.OneToOne
+
+@Entity
+data class ReferralDetails (
+  @Id val id: UUID,
+  @OneToOne var supersededBy: ReferralDetails?,
```
sldblog, line 14: _performance question_
Would loading this field cause recursive resolution of the entire chain? As in, the first detail would load the next, then the next, etc...? Asking as we don't need that behaviour in typical use (we might be able to achieve updates to this by a specific repository update function)

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/ReferralService.kt
```diff
@@ -345,9 +348,39 @@ class ReferralService(
     }
   }
 
-  fun updateDraftReferral(referral: Referral, update: DraftReferralDTO): Referral {
-    validateDraftReferralUpdate(referral, update)
+  private fun updateContainsReferralDetails(update: DraftReferralDTO): Boolean {
+    return update.completionDeadline != null || update.furtherInformation != null || update.maximumEnforceableDays != null
```
sldblog, line 352: _not a blocker, semantics_
`DraftReferralDTO` might support this by having a `details` field or function. Currently it's fairly flat, but we're trying to create a few more granular 'sections', so it could also introduce those sections

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/ReferralService.kt
```diff
@@ -345,9 +348,39 @@ class ReferralService(
...
+    )
 
+    referralDetailsRepository.saveAndFlush(newDetails)
+
+    existingDetails?.let {
+      it.supersededBy = newDetails
+      referralDetailsRepository.saveAndFlush(it)
+    }
+
+    // TODO: currently we are duplicating these fields in both the referral, and
```
sldblog, line 381: _nit, not a blocker_
Perhaps we could move these to a function that is `@Deprecated` so we can be warned on build?

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/ReferralService.kt
```diff
@@ -345,9 +348,39 @@ class ReferralService(
...
+      referral.id,
+      OffsetDateTime.now(),
+      actor.id,
+      reason,
+      update.completionDeadline ?: existingDetails?.completionDeadline,
+      update.furtherInformation ?: existingDetails?.furtherInformation,
+      update.maximumEnforceableDays ?: existingDetails?.maximumEnforceableDays,
+    )
 
+    referralDetailsRepository.saveAndFlush(newDetails)
```
sldblog, line 374: _testability question_
Could we add a test for adding a new version of details, please?

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/jpa/entity/ReferralDetails.kt
tomsmyers: dealt with in cd25dfe3a6467f6f42d7b5402c2189f9fca3e18d

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/ReferralService.kt
tomsmyers: agreed - i think i'll save this for later though to seperate out any API changes from this work.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/ReferralService.kt
tomsmyers: 👍

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/ReferralService.kt
tomsmyers: yes - incoming

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/ReferralServiceUnitTest.kt
```diff
@@ -795,4 +800,86 @@ class ReferralServiceUnitTest {
...
+        draftReferral.createdAt,
+        draftReferral.createdBy.id,
+        "initial version",
+        LocalDate.of(2022, 3, 28),
+        null,
+        null,
+      )
+
+      whenever(referralDetailsRepository.findLatestByReferralId(draftReferral.id)).thenReturn(existingDetails)
+      referralService.updateDraftReferral(draftReferral, DraftReferralDTO(maximumEnforceableDays = 22))
```
sldblog, line 875: Is it worth testing all fields that are changeable here to ensure that all of `completionDeadline`, `furtherInformation` and `maximumEnforceableDays` are persisted?

And also, perhaps, that if they are missing (`null`) the existing values are never overwritten?

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/ReferralServiceUnitTest.kt
```diff
@@ -795,4 +800,86 @@ class ReferralServiceUnitTest {
...
+        draftReferral.createdAt,
+        draftReferral.createdBy.id,
+        "initial version",
+        LocalDate.of(2022, 3, 28),
+        null,
+        null,
+      )
+
+      whenever(referralDetailsRepository.findLatestByReferralId(draftReferral.id)).thenReturn(existingDetails)
+      referralService.updateDraftReferral(draftReferral, DraftReferralDTO(furtherInformation = "nothing to see here"))
```
sldblog, line 848: This `furtherInformation` field isn't used below, do we need it?

tomsmyers: the update has to have something in it otherwise it's not applied

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/ReferralServiceUnitTest.kt
tomsmyers: the latter part is tested here by checking that the completion deadline is unchanged in the new version. 

i can also add furtherInformation to ensure it is also persisted.

tomsmyers: but i can write a new test with explicit expectations

sldblog (approved):

PR #901, opened 2022-03-29
backfill referral_details table with existing referral details (which don't already have data in the new table)

tomsmyers:

backfill referral_details table with existing referral details (which don't already have data in the new table)

ensure that the new table is fully up to date with data from all existing and in progress referrals. this corresponds to step 2 in this plan https://github.com/ministryofjustice/hmpps-interventions-service/discussions/898

### DIFF
--- a/src/main/resources/db/migration/V1_95_4__referral_details_backfill_with_existing_referrals.sql
+++ b/src/main/resources/db/migration/V1_95_4__referral_details_backfill_with_existing_referrals.sql
@@ -1,0 +1,26 @@
+insert into referral_details (
+    id,
+    superseded_by_id,
+    created_at,
+    created_by,
+    reason_for_change,
+    referral_id,
+    completion_deadline,
+    further_information,
+    maximum_enforceable_days
+)
+select
+    gen_random_uuid(),
+    null,
+    ref.created_at,
+    ref.created_by_id,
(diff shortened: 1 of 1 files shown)


src/main/resources/db/migration/V1_95_4__referral_details_backfill_with_existing_referrals.sql
```diff
@@ -0,0 +1,26 @@
...
+    completion_deadline,
+    further_information,
+    maximum_enforceable_days
+)
+select
+    gen_random_uuid(),
+    null,
+    ref.created_at,
+    ref.created_by_id,
+    'initial referral details',
```
tomsmyers, line 17: should we make this field have some indication that these came from existing referrals?


it's content has no impact on anything really.

tomsmyers, line 17: should we make this field have some indication that these came from existing referrals?


it's content has no impact on anything really.

sldblog: Perhaps "backfilled from old referral details"?

sldblog (approved):
LGTM 👍 (one comment about default "reason for change")

PR #904, opened 2022-03-30
delete all referral details entities in between pact test runs to avoid foreign key constraint errors when deleting referrals

tomsmyers:

delete all referral details entities in between pact test runs to avoid foreign key constraint errors when delete referrals

fix the failing pact tests in the `main` branch pipeline


src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/integration/SetupAssistant.kt
```diff
@@ -114,6 +116,7 @@ class SetupAssistant(
     appointmentDeliveryRepository.deleteAll()
     appointmentRepository.deleteAll()
 
+    referralDetailsRepository.deleteAll()
```
sldblog, line 119: _future-proofing, not a blocker_
To avoid remembering cleaning individual tables, I wonder if we can do a checkpoint/snapshot in the beginning and rollback to it at the end, or do a DB reset?

sldblog (approved):

PR #906, opened 2022-03-30
Improve performance of dashboard lookup queries 2

kth13:

- Created a new endpoint sentreferrals/summaries which returns the payload lighter than the sentreferrals/paged end point.
- Created a separate entity view for sentreferrals/summaries which would only query and pull the necessary fields for the landing dashboard in contrary to pulling in lots of fields from the database.
- This reduces the number of queries executed when the landing page is loaded.

- The intent behind these changes is there was performance issue when the landing page was loaded. whilst investigating we found out the API is pulling in lots of fields which is not required.


tomsmyers:
✏️ would be great to do some commit squashing here to tidy up the history

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/jpa/entity/ReferralForDashboard.kt
```diff
@@ -0,0 +1,79 @@
...
+      attributeNodes = [
+        NamedAttributeNode("disabilities"),
+        NamedAttributeNode("referral"),
+      ]
+    )
+  ]
+)
+@Entity(name = "referral")
+@Table(name = "referral", indexes = arrayOf(Index(columnList = "created_by_id")))
+class ReferralForDashboard(
```
tomsmyers, line 58: nit on the naming of this entity: this service doesn't have the concept of dashboards, so i think we should keep the name generic

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/controller/ReferralController.kt
```diff
@@ -33,14 +33,15 @@ import uk.gov.justice.digital.hmpps.hmppsinterventionsservice.dto.EndReferralReq
 import uk.gov.justice.digital.hmpps.hmppsinterventionsservice.dto.ReferralAssignmentDTO
 import uk.gov.justice.digital.hmpps.hmppsinterventionsservice.dto.SelectedDesiredOutcomesDTO
 import uk.gov.justice.digital.hmpps.hmppsinterventionsservice.dto.SentReferralDTO
-import uk.gov.justice.digital.hmpps.hmppsinterventionsservice.dto.SentReferralSummaryDTO
```
tomsmyers, line 36: can we mark this old DTO @Deprecated? once the API endpoints using it are no longer in use we can remove them

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/controller/ReferralController.kt
```diff
@@ -121,16 +122,16 @@ class ReferralController(
     @Nullable @RequestParam(name = "cancelled", required = false) cancelled: Boolean?,
     @Nullable @RequestParam(name = "unassigned", required = false) unassigned: Boolean?,
     @Nullable @RequestParam(name = "assignedTo", required = false) assignedToUserId: String?,
-  ): List<SentReferralSummaryDTO> {
+  ): List<SentReferralSummaryForDashboardDTO> {
```
tomsmyers, line 125: we can't really change the response types of these existing endpoints, since the new responses are not backwards compatible with the old responses.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/controller/ReferralController.kt
```diff
@@ -142,9 +143,29 @@ class ReferralController(
...
+    return (referralService.getSentReferralsForUser(user, concluded, cancelled, unassigned, assignedToUserId, page) as Page<ReferralForDashboard>).map { SentReferralSummaryForDashboardDTO.from(it) }.also {
+      telemetryClient.trackEvent(
+        "PagedDashboardRequest",
+        null,
+        mutableMapOf("totalNumberOfReferrals" to it.totalElements.toDouble(), "pageSize" to page.pageSize.toDouble())
+      )
+    }
+  }
+
+  @JsonView(Views.SentReferral::class)
```
tomsmyers, line 157: i don't think we need the JsonView annotation on this method. 

This annotation will only affect responses which contain DraftReferralDTOs.

if you think it's better safe than sorry, we can include this here too. i don't mind.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/dto/SentReferralSummaryForDashboardDTO.kt
```diff
@@ -0,0 +1,35 @@
...
+
+data class SentReferralSummaryForDashboardDTO(
+  val id: UUID,
+  val sentAt: OffsetDateTime,
+  val sentBy: AuthUserDTO,
+  val referenceNumber: String,
+  val assignedTo: AuthUserDTO?,
+  val serviceUser: ServiceUserDTO?,
+  val serviceProvider: ServiceProviderDTO?,
+  val supplementaryRiskId: UUID,
```
tomsmyers, line 15: why does this class need the risk Id?

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/dto/SentReferralSummaryForDashboardDTO.kt
```diff
@@ -0,0 +1,35 @@
...
+  val interventionTitle: String,
+  val concludedAt: OffsetDateTime?
+) {
+  companion object {
+    fun from(referral: ReferralForDashboard): SentReferralSummaryForDashboardDTO {
+      return SentReferralSummaryForDashboardDTO(
+        id = referral.id,
+        sentAt = referral.sentAt!!,
```
tomsmyers, line 23: it would be cool if this new entity projection has these fields as non-nullable. 

the only reason they are nullable in the existing entity is that it is shared by referrals that in both draft and sent states. all referrals here will be 'sent' - i think !

tomsmyers (commented):
great work kumar - a few comments to look at before we can merge.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/jpa/entity/ReferralForDashboard.kt
kth13: I agree Tom but naming a code is one of the complex jobs for a developer 😢 , Can I name it as ReferralView?

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/controller/ReferralController.kt
kth13: Sure I will make a point on this and action it

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/controller/ReferralController.kt
kth13: I really was worried about that too but the reason why I changed this DTO is because the UI code which was calling this endpoint got changed to the new endpoint and there is no one calling the old endpoint. All the UI related code like the dashboardPresenter got changed to cater the new endpoint. If we have to keep both the api's with different models, we have to maintain two different service methods(one for fetching the the model and other for the old one). That urged me to make this change but I am 100% ok to revert this change.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/dto/SentReferralSummaryForDashboardDTO.kt
kth13: Might have come in by mistake, I will remove it

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/dto/SentReferralSummaryForDashboardDTO.kt
kth13: So in the landing page, we won't show the Draft Referrals?

kth13:
> ✏️ would be great to do some commit squashing here to tidy up the history

Sure will do.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/dto/SentReferralSummariesDTO.kt
```diff
@@ -0,0 +1,33 @@
...
+import java.time.OffsetDateTime
+import java.util.UUID
+
+data class SentReferralSummariesDTO(
+  val id: UUID,
+  val sentAt: OffsetDateTime,
+  val sentBy: AuthUserDTO,
+  val referenceNumber: String,
+  val assignedTo: AuthUserDTO?,
+  val serviceUser: ServiceUserDTO?,
```
tomsmyers, line 13: nit: i think this should be non-nullable

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/dto/SentReferralSummariesDTO.kt
```diff
@@ -0,0 +1,33 @@
...
+  val serviceProvider: ServiceProviderDTO?,
```
tomsmyers, line 14: same with this

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/jpa/entity/ReferralSummary.kt
```diff
@@ -0,0 +1,78 @@
...
+      attributeNodes = [
+        NamedAttributeNode("disabilities"),
+        NamedAttributeNode("referral"),
+      ]
+    )
+  ]
+)
+@Entity(name = "referral")
+@Table(name = "referral", indexes = arrayOf(Index(columnList = "created_by_id")))
+class ReferralSummary(
```
tomsmyers, line 58: nit: this entity probably reflects 'SentReferralSummary' due to non-nullable fields like referenceNumber and sentAt.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/jpa/specification/ReferralSpecifications.kt
```diff
@@ -39,30 +39,30 @@ class ReferralSpecifications {
...
      */
-    fun cancelled(): Specification<Referral> {
-      return Specification<Referral> { root, query, cb ->
+    fun <T> cancelled(): Specification<T> {
+      return Specification<T> { root, query, cb ->
         cb.and(
           cb.isNotNull(root.get<OffsetDateTime>("endRequestedAt")),
           cb.isNotNull(root.get<OffsetDateTime>("concludedAt")),
-          root.join<Referral, EndOfServiceReport>("endOfServiceReport", JoinType.LEFT).isNull
+          root.join<ReferralSummary, EndOfServiceReport>("endOfServiceReport", JoinType.LEFT).isNull
```
tomsmyers, line 47: should this be `T`?

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/jpa/specification/ReferralSpecifications.kt
```diff
@@ -75,15 +75,15 @@ class ReferralSpecifications {
...
         // INNER JOIN
-        val referralAssignmentJoin = root.join<Referral, ReferralAssignment>("assignments", JoinType.INNER)
+        val referralAssignmentJoin = root.join<T, ReferralAssignment>("assignments", JoinType.INNER)
 
         // WHERE assigned_at = max(assigned_at)
         val subquery = query.subquery(OffsetDateTime::class.java)
-        val subQueryReferral = subquery.from(Referral::class.java)
-        val subQueryReferralAssignmentJoin = subQueryReferral.join<Referral, ReferralAssignment>("assignments", JoinType.INNER)
+        val subQueryReferral = subquery.from(T::class.java)
+        val subQueryReferralAssignmentJoin = subQueryReferral.join<ReferralSummary, ReferralAssignment>("assignments", JoinType.INNER)
```
tomsmyers, line 86: should this be `T`?

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/ReferralService.kt
```diff
@@ -124,13 +127,34 @@ class ReferralService(
...
   }
 
-  private fun applyOptionalConjunction(existingSpec: Specification<Referral>, predicate: Boolean?, specToJoin: Specification<Referral>): Specification<Referral> {
+  private fun <T> applyOptionalConjunction(existingSpec: Specification<T>, predicate: Boolean?, specToJoin: Specification<T>): Specification<T> {
     if (predicate == null) return existingSpec
     return existingSpec.and(if (predicate) specToJoin else not(specToJoin))
   }
 
+  fun getSentReferralSummaryForUser(user: AuthUser, concluded: Boolean?, cancelled: Boolean?, unassigned: Boolean?, assignedToUserId: String?, page: Pageable?): Iterable<ReferralSummary> {
+    var findSentReferralsSpec = ReferralSpecifications.sent<ReferralSummary>()
```
tomsmyers, line 136: would be good to deduplicate the code here with the code below if possible? perhaps a private parameterized method in this class?

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/ReferralService.kt
```diff
@@ -156,6 +180,13 @@ class ReferralService(
     throw AccessError(user, "unsupported user type", listOf("logins from ${user.authSource} are not supported"))
   }
 
+  private fun getSentReferralSummaryForServiceProviderUser(user: AuthUser, sentReferralFilterSpecification: Specification<ReferralSummary>, page: Pageable?): Iterable<ReferralSummary> {
+    // todo: query for referrals where the service provider has been granted nominated access only
+    val filteredSpec = referralAccessFilter.serviceProviderReferrals(sentReferralFilterSpecification, user)
+    return if (page == null) referralSummariesRepository.findAll(filteredSpec).sortedBy { it.sentAt } else referralSummariesRepository.findAll(filteredSpec, page)
```
tomsmyers, line 186: in this method i don't think page should be nullable? this would simplify the code on line 186.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/ReferralService.kt
```diff
@@ -168,8 +199,29 @@ class ReferralService(
     return referralAccessFilter.serviceProviderReferralSummaries(referralSummaries, user)
   }
 
+  private fun getSentReferralSummaryForProbationPractitionerUser(user: AuthUser, sentReferralFilterSpecification: Specification<ReferralSummary>, page: Pageable?): Iterable<ReferralSummary> {
```
tomsmyers, line 202: same goes for here - these methods should always be called with pageable params

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/ReferralService.kt
```diff
@@ -168,8 +199,29 @@ class ReferralService(
...
+    var referralsForPPUser = ReferralSpecifications.createdBy<ReferralSummary>(user)
+    try {
```
tomsmyers, line 204: it would also be good to do some de-duplication here if possible. not a blocker though.

tomsmyers: i know the plan is to remove the deprecated code in future, but sometimes these things have a habit of sticking around for longer than planned... :)

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/jpa/entity/SampleData.kt
```diff
@@ -94,6 +94,37 @@ class SampleData {
       )
     }
 
+    fun sampleReferralSummary(
```
tomsmyers, line 97: sorry - this isn't clear in the code, but this class is deprecated in favour of the newer factory classes. we shouldn't be adding new code to it.

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/util/ReferralSummariesFactory.kt
```diff
@@ -0,0 +1,247 @@
...
+import uk.gov.justice.digital.hmpps.hmppsinterventionsservice.jpa.entity.ReferralSummary
+import uk.gov.justice.digital.hmpps.hmppsinterventionsservice.jpa.entity.SelectedDesiredOutcomesMapping
+import uk.gov.justice.digital.hmpps.hmppsinterventionsservice.jpa.entity.ServiceCategory
+import uk.gov.justice.digital.hmpps.hmppsinterventionsservice.jpa.entity.ServiceUserData
+import uk.gov.justice.digital.hmpps.hmppsinterventionsservice.jpa.entity.SupplierAssessment
+import java.time.LocalDate
+import java.time.OffsetDateTime
+import java.util.UUID
+
+class ReferralSummariesFactory(em: TestEntityManager? = null) : EntityFactory(em) {
```
tomsmyers, line 21: thanks for adding this! is there anyway to reuse the code in the existing factory class in here ? that code will not be getting removed in the future, and maintiaining two sets of referral data could be a pain. perhaps we could extrac the default values themselves into some data structure used by both factories?

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/dto/SentReferralSummariesDTO.kt
kth13: I was thinking about unassigned cases and thought it will not have serviceUser or serviceProvider?

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/dto/SentReferralSummariesDTO.kt
kth13: I was thinking about unassigned cases and thought it will not have serviceUser or serviceProvider?

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/jpa/entity/ReferralSummary.kt
kth13: Sure will rename this SentReferralSummary.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/ReferralService.kt
kth13: Sure Tom, the only reason why I didn't do like that way is because its going to be deleted any ways. I will change it to be consistent

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/ReferralService.kt
kth13: Sure, got your point

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/ReferralService.kt
kth13: creating the specification can be refactored to a common method and deduplicated.

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/util/ReferralSummariesFactory.kt
kth13: Created a BaseReferralFactory and moved in the saving of referral to the class and made both the factory class subclass the Base class

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/dto/SentReferralSummariesDTO.kt
tomsmyers: service user is the person on probation this referral was made for. so every referral has a service user associated with it. thanks for making the change.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/dto/SentReferralSummariesDTO.kt
tomsmyers: service provider is the organization that runs the intervention. so again every referral will have an associated service provider.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/ReferralService.kt
tomsmyers: awesome - thanks!

tomsmyers (approved):

PR #910, opened 2022-03-31
use the new referral details table to populate referral DTOs

tomsmyers:

use the new referral details table to populate referral DTOs

expose updates to the referral details table to API callers


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/dto/DraftReferralDTO.kt
```diff
@@ -52,19 +52,19 @@ data class DraftReferralDTO(
       return DraftReferralDTO(
         id = referral.id,
         createdAt = referral.createdAt,
-        completionDeadline = referral.completionDeadline,
+        completionDeadline = referral.referralDetails?.completionDeadline,
         complexityLevels = referral.complexityLevelIds?.ifEmpty { null }
           ?.map { ReferralComplexityLevel(it.key, it.value) }
           ?.sortedBy { it.serviceCategoryId },
-        furtherInformation = referral.furtherInformation,
+        furtherInformation = referral.referralDetails?.furtherInformation,
```
kth13, line 59: I guess the reason why you made the referraleDetails as referralDetails?(nullable) is there can be an instances where the referrralDetails may not be present for a referral?

tomsmyers: yeh that's right - e.g. in a newly created draft referral

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/integration/SetupAssistant.kt
```diff
@@ -539,31 +539,32 @@ class SetupAssistant(
...
-        null,
-        referral.id,
-        referral.createdAt,
-        referral.createdBy.id,
-        "initial referral details",
-        completionDeadline,
-        furtherInformation,
-        maximumEnforceableDays,
+    return referralRepository.save(referral).also {
+      val details = referralDetailsRepository.save(
```
sldblog, line 550: Given this is a setup tool, do we need to care about these executing in the same transaction?

sldblog (approved):

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/integration/SetupAssistant.kt
tomsmyers: yeh unfortunately the order in which these things gets done is important to avoid sql constraint violations

PR #918, opened 2022-04-04
ensure referral details are properly populated and cleaned using the new entity structure

tomsmyers:

ensure referral details are properly populated and cleaned using the new entity structure

allows us to run contract tests that alter referral details


src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/integration/SetupAssistant.kt
```diff
@@ -539,6 +549,20 @@ class SetupAssistant(
...
     referral.whenUnavailable = whenUnavailable
 
+    referralDetailsRepository.save(
+      ReferralDetails(
+        UUID.randomUUID(),
+        null,
+        referral.id,
+        referral.createdAt,
+        referral.createdBy.id,
+        "initial referral details",
```
kth13, line 559: Why we need give a default value here?

kth13: Ok, that's good then.

kth13 (approved):
Looks Good to me

PR #920, opened 2022-04-05
🐛 Build with full version number

sldblog:

🐛 Build with full version number

Fixes "losing" the git SHA from the application version:
image

2c7d77a34e94932f64126094759ca2e07c06aafa introduced a regression in the
application version -- the built version is `YYYY-MM-DD` now instead of
`YYYY-MM-DD.buildNumber.gitSHA`

This is because `ARG BUILD_NUMBER` was removed, which is used by the
gradle task `bootBuildInfo` to pass it in

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -15,7 +15,9 @@ RUN ./gradlew classes --exclude-task=generateGitProperties
 COPY src/main/ src/main/
 RUN ./gradlew classes --exclude-task=generateGitProperties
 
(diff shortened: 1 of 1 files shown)


sldblog:
[🐛 Application version is no longer unique](https://trello.com/c/tVSJelmF/232-%F0%9F%90%9B-application-version-is-no-longer-unique)

tomsmyers (approved):

PR #922, opened 2022-04-06
include 'lastName' field in 'ContactiblePerson' interface and populate it from API responses

tomsmyers:

include 'lastName' field in 'ContactiblePerson' interface and populate it from API responses

we require the PP lastname for emails about changes to the referral.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/HMPPSAuthService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/HMPPSAuthService.kt
@@ -23,6 +23,7 @@ class UnverifiedEmailException : RuntimeException()
 data class UserDetail(
   override val firstName: String,
   override val email: String,
+  override val lastName: String,
 ) : ContactablePerson
 
 private const val AuthServiceProviderGroupPrefix = "INT_SP_"
@@ -36,6 +37,7 @@ private data class AuthUserDetailResponse(
   val firstName: String,
   val email: String,
   val verified: Boolean,
+  val lastName: String,
 )
 
 private data class UserEmailResponse(
@@ -87,7 +89,7 @@ class HMPPSAuthService(
           if (!it.verified) {
             throw UnverifiedEmailException()
           }
-          UserDetail(it.firstName, it.email)
+          UserDetail(it.firstName, it.email, it.lastName)
         }
         .block()
     } else {
@@ -98,7 +100,7 @@ class HMPPSAuthService(
           .retrieve()
           .bodyToMono(UserDetailResponse::class.java)
           .withRetryPolicy()
-          .map { it.name.substringBefore(' ') },
+          .map { Pair(it.name.substringBefore(' '), it.name.substringAfterLast(' ')) },
         hmppsAuthApiClient.get(emailUrl)
           .retrieve()
           .onStatus({ it.equals(HttpStatus.NO_CONTENT) }, { Mono.error(UnverifiedEmailException()) })
@@ -106,7 +108,7 @@ class HMPPSAuthService(
           .withRetryPolicy()
           .map { it.email }
       )
-        .map { UserDetail(it.t1, it.t2) }
+        .map { UserDetail(it.t1.first, it.t2, it.t1.second) }
         .block()
     }
   }
--- a/src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/CaseNotesNotificationsServiceTest.kt
+++ b/src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/CaseNotesNotificationsServiceTest.kt
@@ -44,9 +44,9 @@ internal class CaseNotesNotificationsServiceTest {
   @Test
   fun `both PPs (responsible officer) and SPs (assignee) get notifications for a sent case note as long as they didn't send it`() {
 
-    whenever(hmppsAuthService.getUserDetail(any<AuthUser>())).thenReturn(UserDetail("sp", "sp@provider.co.uk"))
+    whenever(hmppsAuthService.getUserDetail(any<AuthUser>())).thenReturn(UserDetail("sp", "sp@provider.co.uk", "last"))
     whenever(referralService.getResponsibleProbationPractitioner(any())).thenReturn(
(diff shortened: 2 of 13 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/HMPPSAuthService.kt
```diff
@@ -98,15 +100,15 @@ class HMPPSAuthService(
           .retrieve()
           .bodyToMono(UserDetailResponse::class.java)
           .withRetryPolicy()
-          .map { it.name.substringBefore(' ') },
+          .map { Pair(it.name.substringBefore(' '), it.name.substringAfterLast(' ')) },
```
sldblog, line 103: _nit_
Would using `it.name.split` with a limit make this simpler?

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/HMPPSAuthService.kt
```diff
@@ -23,6 +23,7 @@ class UnverifiedEmailException : RuntimeException()
 data class UserDetail(
   override val firstName: String,
   override val email: String,
+  override val lastName: String,
 ) : ContactablePerson
```
sldblog, line 27: _observation_
Given the following interface and classes:

```kt
interface ContactablePerson {
  val firstName: String
  val email: String
  val lastName: String
}

data class UserDetail(
  override val firstName: String,
  override val email: String,
  override val lastName: String,
) : ContactablePerson

data class ResponsibleOfficer(
  val firstName: String?,
  val email: String?,
  val staffId: Long?,
  val lastName: String?,
)

data class ResponsibleProbationPractitioner(
  override val firstName: String,
  override val email: String,
  val deliusStaffId: Long?,
  val authUser: AuthUser?,
  override val lastName: String,
) : ContactablePerson
```

Should `ResponsibleOfficer` and `ResponsibleProbationPractitioner` be merged?

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/CaseNotesNotificationsServiceTest.kt
```diff
@@ -44,9 +44,9 @@ internal class CaseNotesNotificationsServiceTest {
   @Test
   fun `both PPs (responsible officer) and SPs (assignee) get notifications for a sent case note as long as they didn't send it`() {
 
-    whenever(hmppsAuthService.getUserDetail(any<AuthUser>())).thenReturn(UserDetail("sp", "sp@provider.co.uk"))
+    whenever(hmppsAuthService.getUserDetail(any<AuthUser>())).thenReturn(UserDetail("sp", "sp@provider.co.uk", "last"))
```
sldblog, line 47: _impact of change in other files_

This class changed but its production code equivalent did not -- should we consider extracting these test objects into a constant/factory?

This applies to all tests that only changed in `thenReturn` data and no production code changes

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/CommunityAPIOffenderServiceTest.kt
```diff
@@ -107,8 +107,8 @@ internal class CommunityAPIOffenderServiceTest {
 
   @Test
   fun `getResponsibleOfficer returns first when there are multiple responsible officers`() {
-    val offenderService = offenderServiceFactory(createMockedRestClient(MockedResponse(offenderManagersLocation, HttpStatus.OK, "[{\"isResponsibleOfficer\": true, \"staff\": {\"forenames\": \"tom\", \"email\": \"tom@tom.tom\"}, \"staffId\": 123}, {\"isResponsibleOfficer\": true}]")))
+    val offenderService = offenderServiceFactory(createMockedRestClient(MockedResponse(offenderManagersLocation, HttpStatus.OK, "[{\"isResponsibleOfficer\": true, \"staff\": {\"forenames\": \"tom\", \"email\": \"tom@tom.tom\", \"surname\": \"jones\"}, \"staffId\": 123}, {\"isResponsibleOfficer\": true}]")))
```
sldblog, line 110: Given this is an optional field, should we also add a test it maps to `null` if `staff.surname` or `staff` doesn't exist?

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/HMPPSAuthService.kt
tomsmyers: possibly - names could be any number of words e.g. tom david lantos myers, this code just plucks the first and last words out. the alternative would be `it.name.split(' ').first` ... `it.name.split(' ').last` or something - whcih i don't think is any neater. do you?

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/CaseNotesNotificationsServiceTest.kt
tomsmyers: yeh this is a fair point. we do do this elsewhere with our factory class pattern. i guess in the past it was thought we didn't need a factory for such a skinny object, but as you say, it does make sense for this reason.

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/CommunityAPIOffenderServiceTest.kt
tomsmyers: yes it would be a fair test. i will add it now.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/HMPPSAuthService.kt
sldblog: 👍 yeah, this is fine, thanks

sldblog (approved):
LGTM. Is it worth adding the factory changes as part of this PR?

PR #923, opened 2022-04-06
Send email to sp when completion date is changed

tomsmyers:
sends an email to the assigned caseworker when the referral completion deadline is update by the PP

if the referral is unassigned, no email is sent. 

notify SPs when things change on the referral.


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/events/ReferralEventPublisher.kt
```diff
@@ -38,6 +47,19 @@ class ReferralEventPublisher(
     applicationEventPublisher.publishEvent(ReferralEvent(this, eventType, referral, getSentReferralURL(referral)))
   }
 
+  fun referralDetailsChangedEvent(referral: Referral, newDetails: ReferralDetails, previousDetails: ReferralDetails) {
+    applicationEventPublisher.publishEvent(
+      ReferralEvent(
+        this, ReferralEventType.DETAILS_CHANGED, referral, getSentReferralURL(referral),
+        mapOf(
```
tomsmyers, line 54: using DTOs here to avoid having to do any entity related ops in the async listener - avoids 'failed to lazily initialize a collection of role' errors

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/notifications/ReferralNotificationService.kt
```diff
@@ -56,6 +65,36 @@ class ReferralNotificationService(
...
+      }
+    }
+  }
+
+  private fun handleDetailsChangedEvent(event: ReferralEvent) {
+    val newDetails = event.data["newDetails"] as ReferralDetailsDTO
+    val previousDetails = event.data["previousDetails"] as ReferralDetailsDTO
+
+    val recipient = hmppsAuthService.getUserDetail(event.data["recipient"] as AuthUserDTO)
+    val updater = hmppsAuthService.getUserDetail(authUserRepository.findByIdOrNull(newDetails.createdById)!!)
```
tomsmyers, line 80: i don't like having to make the repository call here - any thoughts?

kth13: Is it possible to populate that in ReferralDetailsDTO itself?

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/events/ReferralEventPublisher.kt
```diff
@@ -6,13 +6,22 @@ import org.springframework.context.ApplicationEventPublisher
...
 import uk.gov.justice.digital.hmpps.hmppsinterventionsservice.component.LocationMapper
 import uk.gov.justice.digital.hmpps.hmppsinterventionsservice.controller.ReferralController
+import uk.gov.justice.digital.hmpps.hmppsinterventionsservice.dto.AuthUserDTO
+import uk.gov.justice.digital.hmpps.hmppsinterventionsservice.dto.ReferralDetailsDTO
 import uk.gov.justice.digital.hmpps.hmppsinterventionsservice.jpa.entity.Referral
+import uk.gov.justice.digital.hmpps.hmppsinterventionsservice.jpa.entity.ReferralDetails
 
 enum class ReferralEventType {
-  SENT, ASSIGNED, CANCELLED, PREMATURELY_ENDED, COMPLETED
+  SENT, ASSIGNED, CANCELLED, PREMATURELY_ENDED, COMPLETED, DETAILS_CHANGED
```
sldblog, line 15: _language_
Nuanced change: what do you think about `DETAILS_AMENDED`?

"Changed" can mean 'actioned' and 'observing a change', "amended" for me implies an action. (I am often wrong!)

> **change**: make (someone or something) different; alter or modify // replace (something) with something else, especially something of the same kind that is newer or better; substitute one thing for (another)

> **amend**: make minor changes to (a text, piece of legislation, etc.) in order to make it fairer or more accurate, or to reflect changing circumstances

I like the "to reflect changing circumstances" focus on the latter

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/notifications/ReferralNotificationService.kt
```diff
@@ -0,0 +1,101 @@
+package uk.gov.justice.digital.hmpps.hmppsinterventionsservice.service.notifications
```
sldblog, line 1: _suggestion_
Would it make sense to put all instances of `ApplicationListener` to a new `listeners` package? (So it's simpler to track event flow)

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/notifications/ReferralNotificationService.kt
sldblog: To me, making lookups based on an absolute value (primary key) is fine and reasonable

I would probably have concerns about looking up something that may be a "dirty read" -- eg. a publish that happened during a transaction and the listener code kicks off before the transaction commits

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/NotifyReferralServiceTest.kt
```diff
@@ -100,4 +110,52 @@ class NotifyReferralServiceTest {
...
+            referral.createdBy,
+            completionDeadline = LocalDate.of(2022, 4, 13)
+          )
+        ),
+        "currentAssignee" to AuthUserDTO.from(referral.currentAssignee!!)
+      )
+    )
+
+    @Test
+    fun `referral details changed event sends email when completion deadline has changed`() {
```
sldblog, line 146: Can we add a test that 'sensitive' fields are not included in the email by accident in the future, please?

sldblog: Shall we also add a test about the assertion in the PR description?

> if the referral is unassigned, no email is sent.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/events/ReferralEventPublisher.kt
tomsmyers: good idea - thanks!

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/NotifyReferralServiceTest.kt
tomsmyers: yep - good idea.

tomsmyers: found a bug relating to unassigned cases, so thanks for noting this.

sldblog (approved):

PR #931, opened 2022-04-08
Time delay between referrals for same person/service

sldblog:

Time delay between referrals for same person/service

A key business metric we are trying to improve is "waste", characterised
by probation staff having to re-refer someone to the same service.

Our goal as a team is to make sure the "delays" between referrals are
kept long -- which indicate actual needs; and eliminate the need to
"quickly" re-refer.

This metric will be published to Prometheus and AppInsights since both
are wired into micrometer.


sldblog:
📝 Updated the description with the re-referral metrics after I've done a duplicate referral locally (which apparently took 74 seconds)

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/listeners/ReferralMetricsListener.kt
```diff
@@ -0,0 +1,59 @@
...
+      .register(registry)
+
+    val elapsedTime = Duration.between(previous.sentAt, current.sentAt)
+    timer.record(elapsedTime)
+  }
+
+  private fun findPreviousReferral(current: Referral): Referral? {
+    return repository.findAll(
+      otherReferrals(current.id, current.serviceUserCRN, current.intervention),
+      PageRequest.ofSize(1)
```
tomsmyers, line 45: wonder if better to use a limit here rather than page?

tomsmyers (approved):
LGTM - easy to extend to add additional info like whether or not there is a re-referral on referral cancellation

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/listeners/ReferralMetricsListener.kt
sldblog: after a brief online discussion, we've settled on using `findFirstBy`...change coming

sldblog: it seems like `findFirstBy` on its own isn't exposed -- so I'm going to leave it like this for now

PR #934, opened 2022-04-11
Define API for developer portal

sldblog:

- Define API for developer portal
- Define CircleCI and Sentry links for the developer portal
- Define the data dictionary link for everyone to see

The developer portal now accepts OpenAPI definitions from default cloud-platform hostnames (it was blocked before)

This allows us to define the API exposed by this service and use `consumesApis`/`providersApis` in other components

### DIFF
--- a/catalog-info.yaml
+++ b/catalog-info.yaml
@@ -12,6 +12,9 @@ spec:
 apiVersion: backstage.io/v1alpha1
 kind: Component
 metadata:
+  annotations:
+    circleci.com/project-slug: github/ministryofjustice/hmpps-interventions-service
+    sentry.io/project-slug: interventions-service-prod
   name: hmpps-interventions-service
   title: Intervention service
   description: Domain API for tracking the lifecycle of CRS (Commissioned Rehabilitative Services) interventions and services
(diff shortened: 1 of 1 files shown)


sldblog:
(Marking this as draft as the URL read doesn't quite work yet)

catalog-info.yaml
```diff
@@ -23,3 +29,20 @@ spec:
...
+  providesApis:
+    - api:hmpps-interventions
+
+---
+apiVersion: backstage.io/v1alpha1
+kind: API
+metadata:
+  name: hmpps-interventions
+  title: Intervention API
+  description: Domain API for tracking the lifecycle of CRS (Commissioned Rehabilitative Services) interventions and services
```
kth13, line 41: The description is same as the component's description. Is that intentional

catalog-info.yaml
```diff
@@ -23,3 +29,20 @@ spec:
...
+spec:
+  type: openapi
+  lifecycle: production
+  owner: group:hmpps-interventions-dev
+  system: system:hmpps-interventions
+  definition:
+    $text: https://hmpps-interventions-service-dev.apps.live-1.cloud-platform.service.justice.gov.uk/v3/api-docs
```
kth13, line 48: I see dependsOn: tag in the Ui file, don't we need add it here?

sldblog: No, backstage will automatically create the inverse relationship

kth13: Ok

kth13 (approved):
Looks good to me

sldblog:
Thank you! 🙇‍♂️

PR #950, opened 2022-04-19
🧼 Serve dashboard load from generic API pods

sldblog:

Serve dashboard load from generic API pods -- remove "api-suspect" pods, service and ingress route.

This was discussed during review of https://github.com/ministryofjustice/hmpps-interventions-service/pull/882#issuecomment-1085755423

Introduced in 0e7038b4f50f7d72f846bc44556ffe18ea058ea3, we separated
memory-intensive provider dashboard requests from generic API requests
to isolate their impact from 'business-as-usual' operations

Since then, we have introduced pagination (albeit it's not enabled for
users yet) and improved the memory and computation requirements for
dashboards for non-paginated requests

So currently, it makes sense to undo the change and keep operating the
service simpler


Froom2 (approved):

tomsmyers (approved):
thanks david

PR #951, opened 2022-04-20
⚗️  Shorten new report by removing region dimension

sldblog:

⚗️  Shorten new report by removing region dimension

The report today spilled over multiple Slack messages (the webhook has
shorter character limits than my own posts, it seems)

To get around this, drop the region dimension so the report fits into
one message

| Now | After this PR |
| --- | --- |
|  image | image |

### DIFF
--- a/helm_deploy/hmpps-interventions-service/reports/team_kpi_details_ages.sql
+++ b/helm_deploy/hmpps-interventions-service/reports/team_kpi_details_ages.sql
@@ -1,6 +1,5 @@
 -- referral ages by contract type and NPS region
 SELECT ct.name                                                                                          AS contract_type
-     , nr.name                                                                                          AS probation_region
      , TO_CHAR(PERCENTILE_CONT(0.5) WITHIN GROUP ( ORDER BY DATE_TRUNC('day', NOW()) - r.created_at)
                FILTER ( WHERE r.sent_at ISNULL ), 'DD" days" HH24" hours"')                             AS draft_age_median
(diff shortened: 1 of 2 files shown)


helm_deploy/hmpps-interventions-service/reports/team_kpi_details_ages.sql
```diff
@@ -11,7 +10,5 @@ FROM referral r
          JOIN intervention i ON r.intervention_id = i.id
          JOIN dynamic_framework_contract c ON i.dynamic_framework_contract_id = c.id
          JOIN contract_type ct ON c.contract_type_id = ct.id
-         LEFT JOIN pcc_region pr ON c.pcc_region_id = pr.id
-         LEFT JOIN nps_region nr ON (c.nps_region_id = nr.id OR pr.nps_region_id = nr.id)
-GROUP BY ct.name, nr.name
-ORDER BY ct.name, nr.name;
+GROUP BY 1
```
tomsmyers, line 13: do these actually do anything? is the behaviour the same without them?

tomsmyers (approved):

helm_deploy/hmpps-interventions-service/reports/team_kpi_details_ages.sql
sldblog: yes, we need to specify group by as percentiles are aggregate functions and `ct.name` is the 'fact' it aggregates around

PR #952, opened 2022-04-20
Email ro on referral completion deadline changes

kth13:

- Email the responsible officer if there is a change in deadline and only email if the responsible officer is different to the updater.

Emailing the service provider if there is a change in deadline is already in production but we realised even the responsible officer also needs to be intimated. There is a reason behind this change

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/notifications/CaseNotesNotificationsService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/notifications/CaseNotesNotificationsService.kt
@@ -10,7 +10,6 @@ import uk.gov.justice.digital.hmpps.hmppsinterventionsservice.events.CreateCaseN
 import uk.gov.justice.digital.hmpps.hmppsinterventionsservice.exception.AsyncEventExceptionHandling
 import uk.gov.justice.digital.hmpps.hmppsinterventionsservice.jpa.entity.AuthUser
 import uk.gov.justice.digital.hmpps.hmppsinterventionsservice.jpa.entity.Referral
-import uk.gov.justice.digital.hmpps.hmppsinterventionsservice.service.ProbationPractitionerRole.RESPONSIBLE_OFFICER
 import java.util.UUID
 import javax.transaction.Transactional
 
@@ -40,10 +39,11 @@ class CaseNotesNotificationsService(
   }
 
   private fun emailResponsibleProbationPractitioner(referral: Referral, sender: AuthUser, caseNoteId: UUID) {
-    val responsibleOfficer = referralService.getResponsibleProbationPractitioner(referral, sender)
+    val responsibleOfficer = referralService.getResponsibleProbationPractitioner(referral)
+    val senderIsResponsibleOfficer = referralService.isUserTheResponsibleOfficer(responsibleOfficer, sender)
 
     // if the case note was sent by someone other than the RO, email the RO
-    if (responsibleOfficer.role != RESPONSIBLE_OFFICER) {
+    if (!senderIsResponsibleOfficer) {
       emailSender.sendEmail(
         sentTemplate,
         responsibleOfficer.email,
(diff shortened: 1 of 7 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/notifications/CaseNotesNotificationsService.kt
```diff
@@ -40,15 +40,7 @@ class CaseNotesNotificationsService(
...
-    val senderIsResponsibleOfficer =
-      userTypeChecker.isProbationPractitionerUser(sender) &&
-        (
-          (responsibleOfficer.authUser?.let { it == sender } ?: false) ||
-            (
-              responsibleOfficer.deliusStaffId?.let { it == communityAPIOffenderService.getStaffIdentifier(sender) }
-                ?: false // if the RO doesn't have a staff ID we cannot determine if they are the sender, so assume not
-              )
-          )
+    val senderIsResponsibleOfficer = referralService.isUserTheResponsibleOfficer(responsibleOfficer, sender)
```
sldblog, line 43: I like this refactor, it highlights we are effectively asking how the `getResponsibleProbationPractitioner` algorithm decided

```kt
val ro = referralService.getResponsibleProbationPractitioner(...)
referralService.isUserTheResponsibleOfficer(ro, ...)
```

Since it's working backwards, what do you think about exposing this information from the returned object?

Maybe so that we can check `responsibleOfficer.role == RESPONSIBLE_OFFICER` or `responsibleOfficer.role == SENDER`, or similar? We can encode these decisions in the `getResponsibleProbationPractitioner` call.

This would also remove the need for all the new mocking below

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/notifications/ReferralNotificationService.kt
```diff
@@ -109,6 +128,26 @@ class ReferralNotificationService(
     }
   }
 
+  private fun sendEmail(
```
sldblog, line 131: Since this email is specific to `completionDeadlineUpdatedTemplateID`, what do you think about using a specific function name?

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/notifications/CaseNotesNotificationsService.kt
kth13: It is a very good point so that we can encapsulate the role in ResponsibleProbationPractitioner. I will make those changes.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/notifications/ReferralNotificationService.kt
kth13: Agreed @sldblog , I will make those changes

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/notifications/CaseNotesNotificationsService.kt
kth13: @sldblog , while refactoring this method, what is realised, if we want to find out something like this `responsibleOfficer.role == RESPONSIBLE_OFFICER or responsibleOfficer.role == SENDER`, we need to pass the other `user(sender)` as a parameter to the function. Then I feel the method is doing more than one thing like

- get the ProbabationPractioner
- Check whether the user passed is same as the probabationpractioner
- set the role.
I feel it's better off to leave it like that even if we want to do extra mocking.

sldblog: yes, agree. my mistake for not realising that we check if the delius entity is same as the sender

sldblog (approved):

sldblog:
To tidy up the commits, do you want to rebase from main?

kth13:
> To tidy up the commits, should we rebase from main?

@sldblog Sure will do before pushing

PR #953, opened 2022-04-20
🧼 Remove "Dashboard request durations" panel

sldblog:

🧼 Remove "Dashboard request durations" panel from the [dashboard](https://grafana.live.cloud-platform.service.justice.gov.uk/d/PyQ91ARnk/hmpps-refer-and-monitor-an-intervention?orgId=1&from=now-3h&to=now&var-namespace=hmpps-interventions-prod)

Exported by deleting the panel and saving the new JSON directly (hence weird changes)

The load went from those pods to the "service requests":
image

Since e83ace787521b5aa5905a511c4156360bda31b3c it has no data as `exported_service="api-suspect"` no longer exists


kth13 (approved):
Looks good

PR #954, opened 2022-04-21
⚗️  Track follow-up referrals for same person

sldblog:

Introduce tracking "follow-up" referrals along "repeat" referrals.

Definitions:

- "repeat" referral: new referral to same person and intervention
- "follow-up" referral: new referral to same person

The metric deliberately does **not** track the intervention IDs: at
first, we want to track aggregate counts, not transition patterns.

We received user research feedback on the previous metric:

> yes, but just note that sometimes re-referrals happen because they
> have selected the wrong intervention type…e.g. Social inclusion for
> those who are not in custody
>
> that is a big one
>
> that will be one of the bigger reasons for a re-referral in the first
> 7 days i think

This metric implements the above need (measuring re-referrals on
different interventions)


kth13 (approved):
Looks good

PR #955, opened 2022-04-22
⚗️  Merge repeated and redirected metrics

sldblog:

Merge (simplify) repeated and redirected metrics into one **new** metric

| Before | After |
| --- | --- |
| `intervention.referral.repeat_time` | `intervention.referral.repeated_time`<br/>with `intervention="same"` tag |
|  `intervention.referral.redirect_time` | `intervention.referral.repeated_time`<br/>with `intervention="different"` tag|

In AppInsights, I am always selecting both metrics and even aggregate them as "either repeated or redirected"

This makes me think they are semantically the same and would be better to filter them based on a tag instead

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/listeners/ReferralMetricsListener.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/listeners/ReferralMetricsListener.kt
@@ -34,16 +34,14 @@ class ReferralMetricsListener(
   }
 
   private fun timer(previous: Referral, current: Referral): Timer {
-    return if (current.intervention.id == previous.intervention.id) {
-      Timer.builder("intervention.referral.repeat_time")
-        .tag("crn", current.serviceUserCRN)
-        .tag("interventionId", current.intervention.id.toString())
-        .register(registry)
-    } else {
-      Timer.builder("intervention.referral.redirect_time")
-        .tag("crn", current.serviceUserCRN)
-        .register(registry)
-    }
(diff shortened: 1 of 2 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/listeners/ReferralMetricsListener.kt
```diff
@@ -34,16 +34,14 @@ class ReferralMetricsListener(
...
-        .tag("crn", current.serviceUserCRN)
-        .register(registry)
-    }
+    val interventionKind =
+      if (current.intervention.id == previous.intervention.id) "same"
+      else "different"
+
+    return Timer.builder("intervention.referral.repeated_time")
+      .tag("crn", current.serviceUserCRN)
+      .tag("intervention", interventionKind)
```
kth13, line 43: You removed the interventionId, will that not be useful to capture that as well?

sldblog: I realised I am not using it in any of the graphs, only the fact that something is "same" or "different" intervention. So it's a kind of YAGNI

I figured if we want to dig into what exactly happened, we could look up a CRN's referral history

Does that make sense? Do you agree?

kth13: @sldblog I agree with your point.

kth13 (approved):
Looks good to me

PR #964, opened 2022-04-26
Replace repository metric with longest API metric

sldblog:

Replace repository metric with longest API metric

Middle graph:

image

Getting a sense of which are the slowest API calls helps us better than
slowest repository calls


NeilLuptonMoJ (approved):

sldblog:
Thanks! 🙇‍♂️

PR #966, opened 2022-04-26
end of service report is now only required if at least one delivery session is attended, rather than attempted

rgforsyth:

Updates the response so that an end of service report (endOfServiceReportCreationRequired in dto) is only required if at least one delivery session has been attended. 

The current logic is incorrect. Currently and end of service report is required if at least one session is attempted, regardless if it was attended or not.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/jpa/repository/ActionPlanRepository.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/jpa/repository/ActionPlanRepository.kt
@@ -15,4 +15,11 @@ interface ActionPlanRepository : JpaRepository<ActionPlan, UUID> {
       "and appt.appointmentFeedbackSubmittedAt is not null"
   )
   fun countNumberOfAttemptedSessions(referralId: UUID): Int
+
+  @Query(
+    "select count(sesh) from DeliverySession sesh join sesh.appointments appt " +
+      "where sesh.referral.id = :referralId and appt.attended in ('YES', 'LATE') " +
+      "and appt.appointmentFeedbackSubmittedAt is not null"
+  )
+  fun countNumberOfAttendedSessions(referralId: UUID): Int
 }
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/ReferralConcluder.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/ReferralConcluder.kt
@@ -48,8 +48,8 @@ class ReferralConcluder(
     if (allSessionsAttempted)
       return true
 
-    val atLeastOneSessionAttempted = numberOfSessionsAttempted > 0
-    if (atLeastOneSessionAttempted && referral.endRequestedAt != null)
+    val deliveredFirstSubstantiveAppointment = numberOfSessionsAttempted > 0
+    if (deliveredFirstSubstantiveAppointment && referral.endRequestedAt != null)
       return true
 
     return false
@@ -85,13 +85,11 @@ class ReferralConcluder(
 
   private fun countSessionsAttempted(referral: Referral): Int {
     return referral.currentActionPlan?.let {
-      actionPlanRepository.countNumberOfAttemptedSessions(referral.id)
+      return actionPlanRepository.countNumberOfAttemptedSessions(referral.id)
     } ?: 0
   }
 
   private fun countSessionsAttended(referral: Referral): Int {
-    return referral.currentActionPlan?.let {
-      actionPlanRepository.countNumberOfAttendedSessions(referral.id)
-    } ?: 0
+    return actionPlanRepository.countNumberOfAttendedSessions(referral.id)
   }
 }
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/ReferralConcluder.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/ReferralConcluder.kt
@@ -60,7 +60,7 @@ class ReferralConcluder(
     val hasActionPlan = nonNull(referral.currentActionPlan)
 
     val numberOfAttendedSessions = countSessionsAttended(referral)
-    val hasAttemptedNoSessions = numberOfAttendedSessions == 0
+    val hasAttendedNoSessions = numberOfAttendedSessions == 0
 
     val totalNumberOfSessions = referral.currentActionPlan?.numberOfSessions ?: 0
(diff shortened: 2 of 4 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/ReferralConcluder.kt
```diff
@@ -88,4 +88,10 @@ class ReferralConcluder(
       actionPlanRepository.countNumberOfAttemptedSessions(referral.id)
     } ?: 0
   }
+
+  private fun countSessionsAttended(referral: Referral): Int {
+    return referral.currentActionPlan?.let {
+      actionPlanRepository.countNumberOfAttendedSessions(referral.id)
```
sldblog, line 94: Do we need the `currentActionPlan`?  It looks like `actionPlanRepository.countNumberOfAttendedSessions`  returns 0 if there's no approved action plan anyway; is that correct?

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/jpa/repository/ActionPlanRepository.kt
```diff
@@ -15,4 +15,11 @@ interface ActionPlanRepository : JpaRepository<ActionPlan, UUID> {
       "and appt.appointmentFeedbackSubmittedAt is not null"
   )
   fun countNumberOfAttemptedSessions(referralId: UUID): Int
+
+  @Query(
+    "select count(sesh) from DeliverySession sesh join sesh.appointments appt " +
+      "where sesh.referral.id = :referralId and appt.attended in ('YES', 'LATE') " +
+      "and appt.appointmentFeedbackSubmittedAt is not null"
+  )
+  fun countNumberOfAttendedSessions(referralId: UUID): Int
```
sldblog, line 24: 👏 nice

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/ReferralConcluderTest.kt
```diff
@@ -190,11 +190,11 @@ internal class ReferralConcluderTest {
   }
 
   @Test
-  fun `should not flag end of service report as required when no sessions have been attempted`() {
```
sldblog, line 193: should we keep 1 test case for regression that merely an "attempted" session will **not** count towards 'delivery' status? (and update the below maps)

image

sldblog:
I _think_ there's logic about this in the reports as well

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/ReferralConcluder.kt
rgforsyth: Hmm the ticket didn't mention anything to do with the current action plan. I'm not sure, I just implemented it the way it was but only checking for an actual attendance of an appointment.

rgforsyth: The appointments would be moved from any previous action plans anyway wouldn't they?

sldblog:
[Only ask for EOSR where IAA / ICA has occurred](https://trello.com/c/L2AU1Lda/159-only-ask-for-eosr-where-iaa-ica-has-occurred)

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/ReferralConcluderTest.kt
rgforsyth: We aren't calling that repository method in this flow anymore. So we can't really still test it in that way? We have a test that returns '0' from the repository, simulating that no sessions were attended if that's what you mean?

sldblog: Was thinking of explicitly testing the "1 delivery session with 'no' attendance -> must not flag EoSR as required" -- but I might be missing something

sldblog: Perhaps a resolution could be then to add a test for `countNumberOfAttendedSessions`?

rgforsyth: added some repository tests to ensure that only the attended values (YES/LATE) are included in the count. As well as scheduled sessions with no attendance not being included in the count

sldblog:
I think yes, as now

- `requiresEndOfServiceReportCreation` operates on the bases on attendance (yes, late)
- all other places assume "end of service report is missing" when determining cancelled (ie [here](https://github.com/ministryofjustice/hmpps-interventions-service/blob/dda05ebfc8723666a4247afc59f97a9d254729f3/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/jpa/specification/ReferralSpecifications.kt#L44-L46), [here](https://github.com/ministryofjustice/hmpps-interventions-service/blob/dda05ebfc8723666a4247afc59f97a9d254729f3/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/jpa/repository/ReferralSummaryRepositoryImpl.kt#L59) and [here](https://github.com/ministryofjustice/hmpps-interventions-service/blob/dda05ebfc8723666a4247afc59f97a9d254729f3/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/jpa/entity/Referral.kt#L118))
- therefore, a referral with all sessions in "did not attend" would constitute a "cancellation" with this change

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/ReferralConcluder.kt
```diff
@@ -59,12 +59,12 @@ class ReferralConcluder(
 
     val hasActionPlan = nonNull(referral.currentActionPlan)
 
-    val numberOfAttemptedSessions = countSessionsAttempted(referral)
-    val hasAttemptedNoSessions = numberOfAttemptedSessions == 0
+    val numberOfAttendedSessions = countSessionsAttended(referral)
+    val hasAttemptedNoSessions = numberOfAttendedSessions == 0
```
sldblog, line 63: 🙋‍♂️ left side says "attempted", right side says "attended" (similarly below)

kth13 (approved):
LGTM

sldblog:
[Only ask for EOSR where ICA has occurred](https://trello.com/c/L2AU1Lda/159-only-ask-for-eosr-where-ica-has-occurred)

PR #967, opened 2022-04-29
Updated access error messages to be user friendly

JakeGoodwin22:

Changes to the error messages passed through to the front end to be more user friendly, allowing users to know what they need to do in order to resolve the problem.

Currently the content of the errors don't tell the users of the service what they need to do in order to resolve the problem. This PR passes through HTML which the front end will parse, allowing us to display more user friendly messages.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/authorization/ServiceProviderAccessScopeMapper.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/authorization/ServiceProviderAccessScopeMapper.kt
@@ -34,28 +34,70 @@ class ServiceProviderAccessScopeMapper(
 ) {
   private val serviceProviderGroupPrefix = "INT_SP_"
   private val contractGroupPrefix = "INT_CR_"
-  private val errorMessage = "could not map service provider user to access scope"
+  private val noAccessToPageHeader = "You do not have permission to view this page"
+  private val noAccessToServiceHeader = "You do not have permission to view this service"
 
   fun fromUser(user: AuthUser): ServiceProviderAccessScope {
     if (!userTypeChecker.isServiceProviderUser(user)) {
-      throw AccessError(user, errorMessage, listOf("user is not a service provider"))
+      throw AccessError(
+        user,
+        noAccessToPageHeader,
+        listOf("Your account is not set up correctly. Ask an admin user in your organisation to add the ‘CRS provider’ role in HMPPS Digital Services.")
+      )
     }
 
-    val groups = hmppsAuthService.getUserGroups(user)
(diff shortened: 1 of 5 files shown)


sldblog (changes requested):
More verbose user messages are great, but they don't seem to fit in here

These advices are a communication/representation concern, as such should be worded in the UI -- needing backend changes for a href change would be a smell I think

It seems like what we are missing is communicating the _reason_ for the error; can we instead extend the JSON representation with a `reason` field with well-known keys, and do the English translation from those `reason`s in the UI?

Froom2:
> More verbose user messages are great, but they don't seem to fit in here
> 
> These advices are a communication/representation concern, as such should be worded in the UI -- needing backend changes for a href change would be a smell I think
> 
> It seems like what we are missing is communicating the _reason_ for the error; can we instead extend the JSON representation with a `reason` field with well-known keys, and do the English translation from those `reason`s in the UI?
> 
> (This isn't a hard block by the way, I might be missing something, please push back if I'm out of the picture)

Agree with this, Doesn't make a lot of sense for UI things to be present in the backend, and this would also cause us trouble if we ever had to translate the service.

JakeGoodwin22:
> More verbose user messages are great, but they don't seem to fit in here
> 
> These advices are a communication/representation concern, as such should be worded in the UI -- needing backend changes for a href change would be a smell I think
> 
> It seems like what we are missing is communicating the _reason_ for the error; can we instead extend the JSON representation with a `reason` field with well-known keys, and do the English translation from those `reason`s in the UI?
> 
> (This isn't a hard block by the way, I might be missing something, please push back if I'm out of the picture)

I also agree. We should be returning a status code from the API and interpreting that with a message in the front end. Are we OK to go ahead with this change and discuss how we can improve error codes going forward? Possibly in the Dev chat. @sldblog

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/authorization/ServiceProviderAccessScopeMapper.kt
```diff
@@ -34,28 +34,70 @@ class ServiceProviderAccessScopeMapper(
...
+      )
+    }
 
+    blockUsersWithoutContracts(workingScope)
     if (workingScope.errors.isNotEmpty()) {
-      throw AccessError(user, errorMessage, workingScope.errors)
+      throw AccessError(
+        user,
+        noAccessToServiceHeader,
+        listOf("You do not have any contract groups on your account. Ask an admin in your organisation to set this up in HMPPS Digital Services.")
```
kth13, line 99: @JakeGoodwin22 I believe this scenario needs to be covered in test?

JakeGoodwin22:
> 

Sounds good to me. Will look at doing this!

PR #973, opened 2022-05-04
Mark all older assignments superseded

sldblog:

Mark all older assignments superseded

The code makes sure newer assignments get correctly superseded, but I forgot to finish this data migration

We should have an invariant for assignments that all older assignments have been superseded
In other words: there is only **zero or one** assignment with `superseded = FALSE`

This invariant will simplify querying for the typical "latest assignment"

Finishes migration started in f1f884b7d1c23da8d367d452971573550b2061c1


kth13 (approved):
LGTM

PR #976, opened 2022-05-06
Optimise the unpaginated provider dashboard

sldblog:

On preprod, this optimises the query used by `/sent-referrals/summary/service-provider` by removing the need to do a `WindowAgg` and a disk `Sort` on the result set, saving about 35% of query execution time

- Before: https://explain.depesz.com/s/B7jX
- After: https://explain.depesz.com/s/Vvai

`row_number() over(partition by r.id order by ra.assigned_at desc) as assigned_at_desc_seq` is no longer necessary to pick the latest assignment; we can use `ra.superseded = false` instead

Even though we plan to deprecate the endpoint using this query in two weeks, I think this is a simple enough change to warrant the extra savings

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/jpa/repository/ReferralSummaryRepositoryImpl.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/jpa/repository/ReferralSummaryRepositoryImpl.kt
@@ -15,19 +15,7 @@ class ReferralSummaryRepositoryImpl : ReferralSummaryRepository {
 
   private fun summariesQuery(customCriteria: String?): String {
     val dashboardRestrictionCriteria = customCriteria?.let { " $customCriteria " } ?: ""
-    return """select 
-      referralId, 
-      sentAt,
-			referenceNumber,
-			interventionTitle,
-      dynamicFrameworkContractId,
-			assignedToUserName,
-			serviceUserFirstName,
-			serviceUserLastName,
-      endOfServiceReportId,
-      endOfServiceReportSubmittedAt,
-      concludedAt from (	
-	select
+    return """select
 			cast(r.id as varchar) AS referralId,
 			cast(r.sent_at as TIMESTAMP WITH TIME ZONE) as sentAt,
 			r.reference_number as referenceNumber,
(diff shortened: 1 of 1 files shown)


sldblog:
There is also another ~20% improvement if we introduce joining through `supplier_assessment` -- as it reduces the need to scan the entire appointments table by referrals

```diff
-         LEFT OUTER JOIN appointment app
-           LEFT JOIN supplier_assessment_appointment saa ON saa.appointment_id = app.id
-           ON app.referral_id = r.id
+         LEFT JOIN supplier_assessment sa ON sa.referral_id = r.id
+         LEFT JOIN supplier_assessment_appointment saa ON saa.supplier_assessment_id = sa.id
+         LEFT JOIN appointment app ON saa.appointment_id = app.id
```

I will defer that to another PR

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/jpa/repository/ReferralRepositoryTest.kt
```diff
@@ -412,23 +412,19 @@ class ReferralRepositoryTest @Autowired constructor(
...
-      )
-    } ?: run {
-      referralFactory.createSent(
-        intervention = createIntervention(asPrime),
-        assignments = assignedUsers.map { ReferralAssignment(OffsetDateTime.now(), assignedBy = it, assignedTo = it) }
-      )
+    val assignedUsers = mutableListOf<ReferralAssignment>().apply {
+      repeat(numberOfAssignedUsers) {
+        val user = authUserFactory.create(random(6), random(5), random(12))
+        this.add(ReferralAssignment(OffsetDateTime.now(), assignedBy = user, assignedTo = user, superseded = true))
```
sldblog, line 418: Would be much better if we could use `repeat(...) { assignmentService.assign(referral, user, user) }` or similar

kth13:
@sldblog when we switch to 100% of pagination, we won't be requiring this API. Are we planning to keep this API for some more time?

sldblog:
@kth13 I think it's worth keeping for a week after 100%. The test setup changes are valuable on their own though -- they were setting up an invalid state.

kth13:
@sldblog Sure Ok.

sldblog:
Todo:

- [ ] Remove the unpaginated dashboard instead -- it's used by 0% of users

PR #982, opened 2022-05-10
Updated Postgres to version 14.2

JakeGoodwin22:

Updates the version of postgres to 14.2 to prevent our old version of the image becoming deprecated and unsupported.

Prevents us from having an unsupported postgres image in the future.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -17,6 +17,15 @@ _db_docker_config: &db_docker_config
   - image: cimg/openjdk:17.0
     environment:
       POSTGRES_DB: interventions_ci
+  - image: cimg/postgres:14.2
+    environment:
+      POSTGRES_PASSWORD: password
+      POSTGRES_DB: interventions_ci
+
+_db_docker_config_postgres10: &db_docker_config_postgres10
+  - image: cimg/openjdk:17.0
+    environment:
+      POSTGRES_DB: interventions_ci
   - image: circleci/postgres:10-alpine-ram
     environment:
       POSTGRES_PASSWORD: password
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -17,7 +17,7 @@ _db_docker_config: &db_docker_config
   - image: cimg/openjdk:17.0
     environment:
       POSTGRES_DB: interventions_ci
-  - image: circleci/postgres:10-alpine-ram
+  - image: postgres:14.2
```
sldblog, line 20: 🙋‍♂️ `cimg/postgres:14.2` replaced the older `circleci/postgres` images

JakeGoodwin22: Yeah, what confused me is when updating it to `image: cimg/postgres:14.2` circleci fails because the docker pull command has changed from what was "docker pull circleci/postgres:10-alpine-ram" to "docker pull postgres:14.2" so by using the "cimg/" it was pulling an image which didn't exist within docker hub. Confusing because the docs even say to use cimg/postgres:14.2 but it didn't work

sldblog (commented):
Since this change would introduce a difference between the production database version and the tested version, could we create duplicates of these jobs, one for v10 and one for v14? 

That way we can be confident we don't accidentally introduce incompatible migrations or queries until we migrated the production version, but also enable the UI etc to use v14 as a backing DB

sldblog:
> I suspect it needs to be merged before the jobs are picked up?

It should be picked up by the branch workflow -- these jobs have to be added to the `workflow.jobs` section in the `.circleci/config.yml`

JakeGoodwin22:
> > I suspect it needs to be merged before the jobs are picked up?
> 
> It should be picked up by the branch workflow -- these jobs have to be added to the `workflows.build_test_and_deploy.jobs` section in the `.circleci/config.yml`
> 
> In the CircleCI model, jobs declare the tasks and workflows declare the orchestration and dependency between tasks

Thank you! Added these now

NeilLuptonMoJ (approved):

PR #983, opened 2022-05-11
decommission the deprecated referral dashboard api

kth13:

updates the pact test for the new API `/sent-referrals/summaries` and remove the pact test for the deprecated API `/sent-referrals`

We have published a new API `/sent-referrals/summaries`, so it requires the pact test to be updated to test the new API. Also, we have deprecated the old API `/sent-referrals`, so it requires the pact test to be removed for that API.


src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/ReferralServiceTest.kt
```diff
@@ -635,8 +634,8 @@ class ReferralServiceTest @Autowired constructor(
       whenever(serviceProviderAccessScopeMapper.fromUser(user))
         .thenReturn(ServiceProviderAccessScope(setOf(serviceProviderFactory.create()), validContracts.toSet()))
 
-      val filteredReferrals = referralService.getSentReferralsForUser(user, null, null, null, null, null)
-      assertThat((filteredReferrals as List<Views.SentReferral>).size).isEqualTo(3)
+      val filteredReferrals = referralService.getSentReferralSummaryForUser(user, null, null, null, null, PageRequest.of(0, 10))
+      assertThat((filteredReferrals as Page<SentReferralSummary>).size).isEqualTo(10)
```
sldblog, line 638: Feels like this was testing that invalid contracts are properly filtered out (we get 3 referrals instead of 5), whereas the new test seems to test the page size -- is this correct?

sldblog (commented):
LGTM :) Great to see stuff deleted

One comment about a test

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/ReferralServiceTest.kt
kth13: @sldblog , we noticed and updated this already

sldblog (approved):
🎉 

(pending on pact tests)

PR #990, opened 2022-05-17
Add contract type name to draft referral

Froom2:

Adds the Contract Type name to the draft referral DTO to be passed to the frontend.

It was decided to add this information in the backend rather than making additional calls from the frontend.


JakeGoodwin22 (approved):

JakeGoodwin22:
Looks good - I've approved changes in their current state but of course they can't be approved until the pipelines are successful

PR #1010, opened 2022-05-30
Fix "dependency call times" graph

sldblog:

Fix "dependency call times" graph

The tag name changed from `clientName` to `client_name`, probably due to some dependency update

To make sure we can still tell which dependency is slower

| Before | After |
| --- | --- |
| image | image |


kth13 (approved):
LGTM

PR #1013, opened 2022-06-01
Fix "Previous deployment not found" message in the deployment notification

sldblog:

🐛 This fixes the "Previous deployment not found, showing current commit only." messages during deployment.

image

https://mojdt.slack.com/archives/CLA97UR7D/p1654094487087869

The changelog makes the assumption that there is a resource in Kubernetes called `deployment/PROJECT_NAME`

Since 4ca388769c5721fa2b514679d42370853a2df0d7, there is no `deployment/hmpps-interventions-service`; the closest replacement is the `deployment/hmpps-interventions-service-api`.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -298,7 +298,6 @@ workflows:
       - hmpps/deploy_env:
           name: deploy_preprod
           env: "preprod"
-          show_changelog: true
           slack_notification: true
           slack_channel_name: "interventions-dev-notifications"
           requires: [deploy_dev]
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -298,9 +299,9 @@ workflows:
       - hmpps/deploy_env:
           name: deploy_preprod
           env: "preprod"
-          show_changelog: true
```
sldblog, line 301: please see 64739481ed44fae435d5650fe5e8726078834b22 for reason for removal

NeilLuptonMoJ (approved):

PR #1019, opened 2022-06-08
Split referrals to draft and live referrals

kth13:

- It introduces a new table draft referral
- All draft referrals will be residing in this table instead of referral table
- once the draft referrals is sent, it is moved to referral table
- draft referral data will not be deleted as part of the first release
- The existing draft referral in the referral table will be deleted after the migration is done

- referral table was growing and we saw draft referrals holds 30% of data
- By segregating draft referrals , we can fetch only the data that we need. This improves the performance of the queries

## Release plan

Splitting Live and Draft Referrals initial implementation details

- Draft referrals table will acts as primary table holding the foreign referential integrity for the following
    - selectedServiceCategories,
    - selectedDesiredOutcomes,
    - complexityLevelIds,
    - draft_oasys_risk_information,
    - referral_details
- Since this PR requires adding new constraints and removing existing constraints, it would be better to do this in out of office hours, queries like below requires insert and update lock
   - ```add constraint fk_referral_details_referral_id foreign key (referral_id) references draft_referral;```
   - ```CREATE TABLE draft_referral AS SELECT * FROM referral;```
- The existing referral is copied into draft referral
- The existing draft referrals present in the referral is deleted once the migration is completed
- The draft and live referral holds the same id
- When the draft referral is moved to the live referral (referral table), the draft referral will not be deleted at the moment due to their references to the other table
- Once the migration is done, we will work out a plan to delete data from draft referral

## Validation

- `SELECT COUNT(*) FROM draft_referral` must equal to `SELECT COUNT(*) FROM referral`

## Rollback scripts if there is any issues with production deployment

- `alter table referral_service_user_data DROP CONSTRAINT fk_referral_id`
- `alter table referral_desired_outcome DROP CONSTRAINT fk__referral_desired_outcome__referral`
- `alter table referral_complexity_level_ids DROP CONSTRAINT fk_referral_complexity_level_ids_referral_id`
- `alter table draft_oasys_risk_information DROP CONSTRAINT fk_draft_oasys_risk_information_referral_id`
- `alter table referral_details DROP CONSTRAINT fk_referral_details_referral_id`
- `alter table referral_selected_service_category DROP CONSTRAINT fk_referral_id`
- `alter table referral_service_user_data add constraint fk_referral_id foreign key (referral_id) references referral`
- `alter table referral_desired_outcome add constraint fk__referral_desired_outcome__referral foreign key (referral_id) references referral`
- `alter table referral_complexity_level_ids add constraint fk_referral_complexity_level_ids_referral_id foreign key (referral_id) references referral`
- `alter table draft_oasys_risk_information add constraint fk_draft_oasys_risk_information_referral_id foreign key (referral_id) references referral`
- `alter table referral_details add constraint fk_referral_details_referral_id foreign key (referral_id) references referral`
- `alter table referral_selected_service_category add constraint fk_referral_id foreign key (referral_id) references referral`
- `drop table draft_referral`
- `delete from metadata where table_name='draft_referral'`


src/main/resources/db/migration/V1_97__create_draft_referral_update_referral_constraints.sql
```diff
@@ -0,0 +1,42 @@
+INSERT INTO metadata (table_name, column_name, sensitive, domain_data)
+  select 'draft_referral',column_name,sensitive,domain_data from metadata  where table_name='referral';
+
+CREATE TABLE draft_referral AS SELECT * FROM referral;
```
sldblog, line 4: During this transaction, old pods still using the `referral` table for drafts can attempt to insert; is that ok?

src/main/resources/db/migration/V1_97__create_draft_referral_update_referral_constraints.sql
```diff
@@ -0,0 +1,42 @@
...
+DROP CONSTRAINT fk_referral_id;
+
+alter table referral_service_user_data
+add constraint fk_referral_id foreign key (referral_id) references draft_referral;
+
+alter table referral_desired_outcome
+add constraint fk__referral_desired_outcome__referral foreign key (referral_id) references draft_referral;
+
+alter table referral_complexity_level_ids
+add constraint fk_referral_complexity_level_ids_referral_id foreign key (referral_id) references draft_referral;
```
sldblog, line 33: Do we need a constraint? We want to be able to clear the draft_referral table periodically, but these tables would need to keep data for live referrals

src/main/resources/db/migration/V1_97__create_draft_referral_update_referral_constraints.sql
kth13: - For the first point, probably we can mitigate that by copying the referrals where sentAt is null again to draft referrals manually

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/ReferralService.kt
```diff
@@ -239,10 +244,42 @@ class ReferralService(
...
+    eventPublisher.referralSentEvent(referral)
     supplierAssessmentService.createSupplierAssessment(referral)
-    return sentReferral
+    // This is done to not break the UI.
+    setDraftReferralFields(draftReferral, referral)
+    return referral
+  }
+
+  private fun createReferral(draftReferral: DraftReferral): Referral {
+    return Referral(
```
sldblog, line 256: We discussed moving this into `Referral` entity class but decided against it because a business transition has better place here

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/ReferralService.kt
```diff
@@ -239,10 +244,42 @@ class ReferralService(
     submitAdditionalRiskInformation(referral, user)
     communityAPIReferralService.send(referral)
 
-    val sentReferral = referralRepository.save(referral)
-    eventPublisher.referralSentEvent(sentReferral)
+    referralRepository.save(referral)
+    eventPublisher.referralSentEvent(referral)
```
sldblog, line 248: I don't remember why but I think there was a reason for saving the result of the `save` and passing that through. Can we use that approach here too?

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/jpa/repository/DraftReferralRepository.kt
```diff
@@ -0,0 +1,22 @@
...
+import org.springframework.data.jpa.repository.JpaRepository
+import org.springframework.data.jpa.repository.Query
+import uk.gov.justice.digital.hmpps.hmppsinterventionsservice.jpa.entity.DraftReferral
+import java.util.Optional
+import java.util.UUID
+
+interface DraftReferralRepository : JpaRepository<DraftReferral, UUID> {
+
+  // queries for draft referrals
+  override fun findById(id: UUID): Optional<DraftReferral>
```
sldblog, line 12: Do we have to define this, if this is `override`?

src/main/resources/db/migration/V1_98__create_draft_referral_update_referral_constraints.sql
```diff
@@ -0,0 +1,42 @@
+INSERT INTO metadata (table_name, column_name, sensitive, domain_data)
+  select 'draft_referral',column_name,sensitive,domain_data from metadata  where table_name='referral';
+
+CREATE TABLE draft_referral AS SELECT * FROM referral;
```
sldblog, line 4: Can we find out what locking mechanism would this use during the migration?

- If it locks `referral` for inserts, we cannot deploy this during working hours
- If it does not lock, we have a potential issue of missing data and we need to do a follow-up copy after the deploy

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/dto/DraftReferralDTO.kt
```diff
@@ -48,6 +49,43 @@ data class DraftReferralDTO(
...
+        serviceProvider = ServiceProviderDTO.from(contract.primeProvider),
+        relevantSentenceId = referral.relevantSentenceId,
+        serviceCategoryIds = referral.selectedServiceCategories?.ifEmpty { null }
+          ?.map { it.id }
+          ?.sorted(),
+        interventionId = referral.intervention.id,
+        contractTypeName = referral.intervention.dynamicFrameworkContract.contractType.name,
+      )
+    }
     fun from(referral: Referral): DraftReferralDTO {
```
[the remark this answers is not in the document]
kth13: This method is an already existing method and it is used in ```SentReferralDTO``` which is in turn used in ```ReferralController```

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/ReferralService.kt
```diff
@@ -102,13 +105,13 @@ class ReferralService(
     return referralRepository.findByIdAndSentAtIsNotNull(id)
   }
 
-  fun getDraftReferralForUser(id: UUID, user: AuthUser): Referral? {
+  fun getDraftReferralForUser(id: UUID, user: AuthUser): Optional<DraftReferral> {
```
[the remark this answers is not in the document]
kth13: updated the method to return DraftReferral?

src/main/resources/db/migration/V1_98__create_draft_referral_update_referral_constraints.sql
```diff
@@ -0,0 +1,42 @@
...
+add constraint fk__referral_desired_outcome__referral foreign key (referral_id) references draft_referral;
+
+alter table referral_complexity_level_ids
+add constraint fk_referral_complexity_level_ids_referral_id foreign key (referral_id) references draft_referral;
+
+alter table draft_oasys_risk_information
+add constraint fk_draft_oasys_risk_information_referral_id foreign key (referral_id) references draft_referral;
+
+alter table referral_details
+add constraint fk_referral_details_referral_id foreign key (referral_id) references draft_referral;
```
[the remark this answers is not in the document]
kth13: Can We do the migration out of hours if we feel that it interferes with a potential update at that time

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/jpa/repository/DraftReferralRepository.kt
```diff
@@ -0,0 +1,22 @@
...
+
+  @Query(
+    value =
+    """select * from draft_referral dr where dr.created_by_id = :userId 
+      and NOT EXISTS 
+      (select from referral where id = dr.id)""",
+    nativeQuery = true
+  )
+  fun findByCreatedById(userId: String): List<DraftReferral>
```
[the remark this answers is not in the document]
kth13: I have checked this query and thought we don't need a native query. So updated the query to JPA query

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/ReferralService.kt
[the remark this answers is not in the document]
kth13: updated it

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/jpa/repository/DraftReferralRepository.kt
kth13: we don't need the override, removed it

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/ReferralService.kt
kth13: using the saved variable is passing it through is not working  in the new approach as when we try to set ``` referral.complexityLevelIds = draftReferral.complexityLevelIds```, it was throwing Found shared references to a ElementCollection exception.

src/main/resources/db/migration/V1_98__create_draft_referral_update_referral_constraints.sql
kth13: This query uses ACCESS SHARE LOCK and being cautious as it is a big change, it would be good during quiet hours

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/ReferralService.kt
kth13: Resolved.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/dto/DraftReferralDTO.kt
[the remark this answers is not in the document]
kth13: Done

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/ReferralService.kt
```diff
@@ -243,10 +247,42 @@ class ReferralService(
...
     communityAPIReferralService.send(referral)
 
-    val sentReferral = referralRepository.save(referral)
-    eventPublisher.referralSentEvent(sentReferral)
+    referralRepository.save(referral)
+    eventPublisher.referralSentEvent(referral)
     supplierAssessmentService.createSupplierAssessment(referral)
-    return sentReferral
+    // This is done to not break the UI.
+    setDraftReferralFields(draftReferral, referral)
```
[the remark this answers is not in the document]
kth13: Removed and added in the controller side

sldblog (approved):
Let's prepare an "undo" migration script just in case, otherwise LGTM

PR #1030, opened 2022-06-13
Add heap tools to production image

sldblog:

Add/enable JVM heap-related tools to the production Docker image.

### Using jhsdb and jmap

Our images run as non-root on the hosting environment, so adding tooling on the fly is not possible.

We are having out-of-memory errors on one instance (ndmis-performane-report containers) so let's include tooling to
investigate the heap (part of JDK).

### Enable saving heap dumps on out-of-memory

All applications run with `-XX:+HeapDumpOnOutOfMemoryError -XX:HeapDumpPath=/dumps`, but this folder is not writeable for the NDMIS cronjob, resulting in

    Unable to create /dumps: Permission denied

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -23,7 +23,7 @@ RUN ./gradlew assemble
 
 
 # ---
-FROM alpine:3.15 AS final
+FROM alpine:3.16 AS final
 LABEL maintainer="HMPPS Digital Studio <info@digital.justice.gov.uk>"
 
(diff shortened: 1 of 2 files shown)


kth13 (approved):
LGTM

sldblog:
[NDMIS report does not generate any more](https://trello.com/c/aZOvTInu/394-ndmis-report-does-not-generate-any-more)

PR #1032, opened 2022-06-15
Update spring boot dependency to fix OWASP vulnerabilities

JakeGoodwin22:

Updates spring boot version to 4.2.3 to fix OWASP vulnerabilities

Fix dependency vulnerabilities within the service.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -1,5 +1,5 @@
 plugins {
-  id("uk.gov.justice.hmpps.gradle-spring-boot") version "4.2.1"
+  id("uk.gov.justice.hmpps.gradle-spring-boot") version "4.2.3"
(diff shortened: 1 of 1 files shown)


build.gradle.kts
```diff
@@ -62,6 +62,8 @@ tasks {
 }
 
 dependencies {
+
+  implementation("org.owasp:dependency-check-gradle:7.1.1")
```
sldblog, line 66: 🙋 This is already included from `uk.gov.justice.hmpps.gradle-spring-boot`, so this change should not be needed. (The nightly job runs `./gradlew dependencyCheckAnalyze` without installing anything)

JakeGoodwin22: Removed 👍

sldblog (approved):
Cheers! 🙇‍♂️

PR #1033, opened 2022-06-16
Remove old deprecated views and tables

sldblog:

Remove old deprecated views and tables.

These views and tables were a stepping stone in extensive migrations and a safety net in case we needed to roll back.

They are no longer referenced or needed.


kth13 (approved):
LGTM

PR #1034, opened 2022-06-16
Remove deprecated fields

sldblog:

Remove deprecated fields from the `referral` table
Remove unused field from the `draft_referral` table

Migration safety:

- `DROP COLUMN` requires an exclusive lock during the DDL change, but does not rewrite the table. So the lock won't be held long. Still, the referral table is used in **everything**, so let's do this migration out of hours (past 9 pm).
- `pg_depends` shows that only the two constraints are referencing these columns and they will be dropped automatically.

`deprecated_assigned_*` fields were a stepping stone in migrating assignments into `referral_assignment`. We left them in case we needed to roll back. They are no longer referenced or needed.

`maximum_enforceable_days`, `completion_deadline` and `further_information` are now in `referral_details` and not needed in either drafts or live referrals.

`reference_number` is not generated for `draft_referral`, so it does not have to exist.


kth13 (approved):
LGTM

sldblog:
Todo: needs to include `draft_referral` table now

.circleci/config.yml
```diff
@@ -54,8 +54,8 @@ jobs:
       - run:
           name: Setup flyway and psql
           command: |
-            wget -qO- https://repo1.maven.org/maven2/org/flywaydb/flyway-commandline/7.7.0/flyway-commandline-7.7.0-linux-x64.tar.gz | \
-              tar xvz && sudo ln -s `pwd`/flyway-7.7.0/flyway /usr/local/bin
+            wget -qO- https://repo1.maven.org/maven2/org/flywaydb/flyway-commandline/8.5.13/flyway-commandline-8.5.13-linux-x64.tar.gz | \
+              tar xvz && sudo ln -s `pwd`/flyway-8.5.13/flyway /usr/local/bin
```
sldblog, line 58: Updates flyway to the version we use now
```
$ ./gradlew dependencies | grep flyway
+--- org.flywaydb:flyway-core -> 8.5.13
```

kth13 (approved):
LGTM

sldblog:
Superseded by #1400
@kth13 Could you re-review that, please? Sorry.

PR #1041, opened 2022-06-20
create a new appointment when rescheduling to store the historic appo…

kth13:
…intment details

- creates a new appointment instead of updating the existing appointment
- By doing that way, we store the historic changes of the appointment

- We realised that when a reschedule of appointment happens, we are not storing the appointment history which seems to be mandatory. So, in the view of storing the appointment history, when a reschedule appointments is requested, we created a new appointment instead of updating the existing one

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/AppointmentService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/AppointmentService.kt
@@ -280,7 +280,7 @@ class AppointmentService(
   }
 
   private fun updateAppointment(
-    oldAppointment: Appointment,
+    appointment: Appointment,
     durationInMinutes: Int,
     appointmentTime: OffsetDateTime,
     deliusAppointmentId: Long?,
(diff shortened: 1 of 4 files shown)


JakeGoodwin22 (approved):
LGTM

sldblog:
@kth13 I couldn't find the context, sorry, why was bbba31bcb5c49d57974cbea0f67b44af891bae57 reverted?

sldblog:
Just a hunch, but could we double-check how this behaves with now duplicate appointments?

As in, would it find the rescheduled appointment?
```
src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/reporting/serviceprovider/performance/PerformanceReportProcessor.kt
42:      firstSessionAttendedAt = approvedActionPlan?.let { actionPlanService.getFirstAttendedAppointment(it)?.appointmentTime },
```

kth13:
> @kth13 Sorry, I couldn't find the context around this; why was [bbba31b](https://github.com/ministryofjustice/hmpps-interventions-service/commit/bbba31bcb5c49d57974cbea0f67b44af891bae57) reverted? (It looks interesting; could the reasons behind these decisions be mentioned in the commit, please?)

We need to come up with a logic of separating between what is really a historical appointment and what is new appointment that was created after it was not attended.

PR #1042, opened 2022-06-23
Book new appointment on did not attend

kth13:

- Creating a convenience field of old appointments for a session in the DTO which will aid the front end to display that data.
- Fixing the small issue while creating the new Appointment for the update Appointment

- There is a requirement to show appointment histories when a PoP did not attend a particular appointment and the appointment being rescheduled. So , to show the appointment histories we are creating a convenience field.


JakeGoodwin22 (approved):
LGTM

PR #1044, opened 2022-06-23
Create new appointment for supplier assessment

kth13:

- When a supplier assessment appointment is rescheduled, a new appointment is created instead of updating the existing appointment
- When the new appointment is created, old appointments is replaced with the new appointment
- PerformanceReportProcessor now gets the appointment data directly from the database instead of supplier assessment

- The supplier assessment was not storing the appointments story when a appointment gets rescheduled, these changes are done to resolve the issue.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/dto/SupplierAssessmentDTO.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/dto/SupplierAssessmentDTO.kt
@@ -13,7 +13,7 @@ data class SupplierAssessmentDTO(
     fun from(supplierAssessment: SupplierAssessment): SupplierAssessmentDTO {
       return SupplierAssessmentDTO(
         id = supplierAssessment.id,
-        appointments = AppointmentDTO.from(supplierAssessment.appointments.filter { !it.superseded }.toMutableSet()),
+        appointments = AppointmentDTO.from(supplierAssessment.appointments.filterNot { it.superseded }.toMutableSet()),
         currentAppointmentId = supplierAssessment.currentAppointment?.id,
         referralId = supplierAssessment.referral.id
       )
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/dto/SupplierAssessmentDTO.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/dto/SupplierAssessmentDTO.kt
@@ -13,7 +13,7 @@ data class SupplierAssessmentDTO(
     fun from(supplierAssessment: SupplierAssessment): SupplierAssessmentDTO {
       return SupplierAssessmentDTO(
         id = supplierAssessment.id,
-        appointments = AppointmentDTO.from(supplierAssessment.appointments),
+        appointments = AppointmentDTO.from(supplierAssessment.appointments.filter { !it.superseded }.toMutableSet()),
         currentAppointmentId = supplierAssessment.currentAppointment?.id,
         referralId = supplierAssessment.referral.id
       )
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/AppointmentService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/AppointmentService.kt
@@ -80,6 +80,7 @@ class AppointmentService(
           communityAPIBookingService.book(referral, appointment, appointmentTime, durationInMinutes, appointmentType, npsOfficeCode)
         updateAppointment(
           durationInMinutes,
+          appointment,
           appointmentTime,
           deliusAppointmentId,
           createdByUser,
@@ -281,6 +282,7 @@ class AppointmentService(
 
   private fun updateAppointment(
     durationInMinutes: Int,
+    oldAppointment: Appointment,
     appointmentTime: OffsetDateTime,
     deliusAppointmentId: Long?,
     createdByUser: AuthUser,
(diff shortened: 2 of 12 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/AppointmentService.kt
```diff
@@ -290,16 +291,25 @@ class AppointmentService(
...
     notifyProbationPractitioner: Boolean?,
     behaviourDescription: String?,
     appointmentType: AppointmentType,
   ): Appointment {
-    appointment.durationInMinutes = durationInMinutes
-    appointment.appointmentTime = appointmentTime
-    appointment.deliusAppointmentId = deliusAppointmentId
-    setAttendanceAndBehaviourIfHistoricAppointment(appointment, appointmentTime, attended, additionalAttendanceInformation, behaviourDescription, notifyProbationPractitioner, createdByUser, appointmentType)
-    val appointment = appointmentRepository.save(appointment)
+    val appointment = Appointment(
```
sldblog, line 300: _Observation_
This change makes `updateAppointment` almost identical to `createAppointment` except for saving the previous appointment with `superseded`.

Should we then instead compose these to avoid duplication?

E.g.
```kt
private fun updateAppointment(...) {
  val newAppointment = createAppointment(...)
  oldAppointment.superseded = true
  appointmentRepository.save(oldAppointment) // do we need saveAndFlush?
  return newAppointment
}
```
or (depending on which order is correct)
```kt
private fun updateAppointment(...) {
  oldAppointment.superseded = true
  appointmentRepository.save(oldAppointment) // do we need saveAndFlush?
  return createAppointment(...)
}
```

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/AppointmentService.kt
```diff
@@ -290,16 +291,25 @@ class AppointmentService(
...
+      id = UUID.randomUUID(),
+      appointmentTime = appointmentTime,
+      durationInMinutes = durationInMinutes,
+      deliusAppointmentId = deliusAppointmentId,
+      createdBy = authUserRepository.save(createdByUser),
+      createdAt = OffsetDateTime.now(),
+      referral = referral,
+    )
+    setAttendanceAndBehaviourIfHistoricAppointment(appointment, attended, additionalAttendanceInformation, behaviourDescription, notifyProbationPractitioner, createdByUser, appointmentType)
+    oldAppointment.superseded = true
```
sldblog, line 310: Do we need this logic in deprecated endpoints as well?

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/dto/SupplierAssessmentDTO.kt
```diff
@@ -13,7 +13,7 @@ data class SupplierAssessmentDTO(
     fun from(supplierAssessment: SupplierAssessment): SupplierAssessmentDTO {
       return SupplierAssessmentDTO(
         id = supplierAssessment.id,
-        appointments = AppointmentDTO.from(supplierAssessment.appointments),
+        appointments = AppointmentDTO.from(supplierAssessment.appointments.filter { !it.superseded }.toMutableSet()),
```
sldblog, line 16: 🙋‍♂️ (nit) Could also be written as `appointments.filterNot { it.superseded }`

sldblog: Could `supplierAssessment.appointments` be filtered on the query level?

sldblog (commented):
Nice work, just a couple of clarifying questions

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/AppointmentService.kt
kth13: Refactored this bit.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/AppointmentService.kt
kth13: This is the deprecated endpoint and we can implement this logic when we migrate to the new API's

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/dto/SupplierAssessmentDTO.kt
kth13: > 🙋‍♂️ (nit) Could also be written as `appointments.filterNot { it.superseded }` (I don't have a preference)

Agree and Done

kth13: > Could `supplierAssessment.appointments` be filtered on the query level?

The application currently queries for the referral and from the referral they retrieve ```sentReferral.supplierAssessment```. I felt this is best place to filter out appointments as they are already extracting there and  no where else

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/AppointmentServiceTest.kt
```diff
@@ -190,14 +190,18 @@ class AppointmentServiceTest {
...
+    whenever(appointmentRepository.saveAndFlush(any())).thenReturn(savedAppointment)
 
     // When
     val updatedAppointment = appointmentService.createOrUpdateAppointment(referral, existingAppointment, durationInMinutes, appointmentTime, SUPPLIER_ASSESSMENT, createdByUser, AppointmentDeliveryType.PHONE_CALL, AppointmentSessionType.ONE_TO_ONE)
 
     // Then
     verifyResponse(updatedAppointment, existingAppointment.id, false, rescheduledDeliusAppointmentId, appointmentTime, durationInMinutes, AppointmentDeliveryType.PHONE_CALL, AppointmentSessionType.ONE_TO_ONE)
     verifySavedAppointment(appointmentTime, durationInMinutes, rescheduledDeliusAppointmentId, AppointmentDeliveryType.PHONE_CALL, AppointmentSessionType.ONE_TO_ONE)
+    val argumentCaptor = argumentCaptor<Appointment>()
+    verify(appointmentRepository, atLeast(1)).save(argumentCaptor.capture())
```
Froom2, line 202: is this passing? the mock has been changed to use saveAndFlush

Froom2 (approved):

PR #1049, opened 2022-06-27
upgrading the uk.gov.justice.hmpps.gradle-spring-boot to 4.3.0

kth13:

- upgrade uk.gov.justice.hmpps.gradle-spring-boot to 4.3.0
- Since kotlin 1.7 comes with the strict checking on the default behaviour of the ```when``` expression
- Link for the actual issue- https://youtrack.jetbrains.com/issue/KT-47709
- Link to the release notes - https://github.com/ministryofjustice/dps-gradle-spring-boot/blob/main/release-notes/4.3.0.md

- upgrade uk.gov.justice.hmpps.gradle-spring-boot to 4.3.0

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -1,5 +1,5 @@
 plugins {
-  id("uk.gov.justice.hmpps.gradle-spring-boot") version "4.2.3"
+  id("uk.gov.justice.hmpps.gradle-spring-boot") version "4.3.0"
   kotlin("plugin.spring") version "1.6.21"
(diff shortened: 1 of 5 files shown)


sldblog (approved):
Love the hard compile error on uncovered enum branches.

I think it strengthens the argument to change `Event(eventEnumType)` into bespoke `SomeEvent`, e.g. `ReferralEvent(ReferralEventType.COMPLETED` to `ReferralCompletedEvent`.

PR #1050, opened 2022-06-28
Filter assignments by superseded flag

sldblog:

Simplifies filtering dashboards for an assigned person. Essentially changes

- from `where ra.assigned_at = (SELECT MAX(assigned_at) FROM referral_assignments WHERE referral_id = ra.referral_id)`
- to `where ra.superseded = false`

## Enabled by

The invariant is now tested on persistence.

I did this because I was struggling with invalid test states and wanted to ensure we could not accidentally test the wrong things.

There is an invariant in the data model that guarantees that all assignments are superseded, except the latest one.

This allows us to remove query and code complexity.

### DIFF
--- a/src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/util/AssignmentsFactory.kt
+++ b/src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/util/AssignmentsFactory.kt
@@ -15,7 +15,12 @@ class AssignmentsFactory(em: TestEntityManager? = null) : EntityFactory(em) {
         this.add(authUserFactory.create(randomAlphanumeric(6), randomAlphanumeric(5), randomAlphanumeric(12)))
       }
     }
-    return assignedUsers.mapIndexed { i, u ->
+    return createInOrder(*assignedUsers.toTypedArray())
+  }
+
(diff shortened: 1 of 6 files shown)


src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/util/AssignmentsFactory.kt
```diff
@@ -0,0 +1,33 @@
...
+  fun create(numberOfAssignments: Int): List<ReferralAssignment> {
+    val assignedUsers = mutableListOf<AuthUser>().apply {
+      repeat(numberOfAssignments) {
+        this.add(authUserFactory.create(randomAlphanumeric(6), randomAlphanumeric(5), randomAlphanumeric(12)))
+      }
+    }
+    return createInOrder(*assignedUsers.toTypedArray())
+  }
+
+  fun createInOrder(vararg users: AuthUser): List<ReferralAssignment> {
```
kth13, line 21: Do we need to name it as createInOrder as we are not maintaining any order and things is the new one superseds the old one

kth13: @sldblog , It was just a clarification. Looking good

kth13 (approved):
LGTM

PR #1051, opened 2022-06-30
Nullified time component from date validation fields to align with UI validation rules

JakeGoodwin22:

Removes the time component from the date validation check to prevent issues where the user see's an error from the API
that isn't caught in the UI validation. We could potentially change this in the future if we want to make the validation stricter
and include the time the referral was sent

Prevent the users (SP) from seeing an API error due to it being stricter than the front end validation.


NeilLuptonMoJ (approved):
👍

sldblog:
> Prevent the users (SP) from seeing an API error due to it being stricter than the front end validation.

@JakeGoodwin22 Is it OK that we have duplicate validation?

PR #1065, opened 2022-07-06
check for cancelled referral were pop did not attend appointment shou…

OmotayoTalabi:

to ignore cancelled unattended referral

check for cancelled referral were pop did not attend appointment should not return


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/jpa/entity/SentReferralSummary.kt
```diff
@@ -29,6 +31,7 @@ import javax.validation.constraints.NotNull
     NamedAttributeNode(value = "serviceUserData", subgraph = "serviceUserData"),
     NamedAttributeNode(value = "intervention", subgraph = "interventions"),
     NamedAttributeNode(value = "endOfServiceReport", subgraph = "endOfServiceReport")
+
```
JakeGoodwin22, line 34: Picky I know, but could this whitespace be removed?

JakeGoodwin22 (approved):
LGTM - one minor comment

PR #1085, opened 2022-07-18
disabling the fix for showing the unattended cancelled referrals in t…

kth13:
…he SP side alone.

- shows cancelled unattended referrals on both sides

- This pull request was intended to disable the fix for not showing the cancelled unattended referrals on the sp side. This has caused an production issue. So, once we fix the issue, we will put the fix back


JakeGoodwin22 (approved):
LGTM

sldblog:
@kth13 Could we add detail about the production issue, please? As-is, people reading this later will not have context on what the effect of the previous code was

PR #1090, opened 2022-07-22
Shrink image by switching back to JRE from JDK

sldblog:

Shrink image by switching back to JRE from JDK

Also switching to `eclipse-temurin:17-jre-alpine` base image as the kotlin template repo is using `eclipse-temurin` too

Included JDK in 9a349a62e1fd591ffc37ac6dc66e1852c5badac7 so we have
command line tools to investigate heap problems

We no longer have heap problems so we can reduce the image size again

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM openjdk:17-slim AS builder
+FROM eclipse-temurin:17-jre-alpine AS builder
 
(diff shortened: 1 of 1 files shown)


JakeGoodwin22 (approved):
LGTM

sldblog:
Thank you!

PR #1091, opened 2022-07-22
Support multiplatform builds

sldblog:

Support x86_64 and arm64 builds out of the box.

Use `BUILDPLATFORM` which is [automatically defined][args].

🚫 Blocked until #1090 is merged (same line change)

Taken from the ["Faster Multi-Platform Builds" guide][guide].

Picked from https://github.com/ministryofjustice/hmpps-template-kotlin/commit/70b9ff6ed5060d00639f135b7339e84ba290f62e

[guide]: https://www.docker.com/blog/faster-multi-platform-builds-dockerfile-cross-compilation-guide/
[args]: https://docs.docker.com/engine/reference/builder/#automatic-platform-args-in-the-global-scope

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -9,7 +9,7 @@ parameters:
     default: main
 
 orbs:
-  hmpps: ministryofjustice/hmpps@3.14
+  hmpps: ministryofjustice/hmpps@5.1
   gradle: circleci/gradle@2.2.0
   mem: circleci/rememborb@0.0.1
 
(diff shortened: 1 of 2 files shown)


sldblog:
Current blocker: we cannot upgrade to use `hmpps/build_multiplatform_docker` if we want to use `hmpps/trivy_pipeline_scan` on `main` as that new job does not persist the image

Either `trivy_pipeline_scan` should pull, `build_multiplatform_docker` should persist the x86_64 version, or we should switch to `trivy_latest_scan` on `main`

PR #1093, opened 2022-07-22
Amend needs and requirement

OmotayoTalabi:

service for amend needs and requirements

to allow pp to amend needs and requirements.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/AmendReferralService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/AmendReferralService.kt
@@ -1,0 +1,48 @@
+package uk.gov.justice.digital.hmpps.hmppsinterventionsservice.service
+
+import org.springframework.stereotype.Service
+import uk.gov.justice.digital.hmpps.hmppsinterventionsservice.dto.AmendComplexityLevelDTO
+import uk.gov.justice.digital.hmpps.hmppsinterventionsservice.dto.ReferralAmendmentDetails
+import uk.gov.justice.digital.hmpps.hmppsinterventionsservice.events.ReferralEventPublisher
+import uk.gov.justice.digital.hmpps.hmppsinterventionsservice.jpa.entity.AuthUser
+import uk.gov.justice.digital.hmpps.hmppsinterventionsservice.jpa.entity.Changelog
+import uk.gov.justice.digital.hmpps.hmppsinterventionsservice.jpa.entity.Referral
+import uk.gov.justice.digital.hmpps.hmppsinterventionsservice.jpa.repository.ChangelogRepository
+import uk.gov.justice.digital.hmpps.hmppsinterventionsservice.jpa.repository.ReferralRepository
+import java.time.OffsetDateTime
+import java.util.UUID
+import javax.transaction.Transactional
+
+@Service
+@Transactional
+class AmendReferralService(
+  private val referralEventPublisher: ReferralEventPublisher,
+  val changelogRepository: ChangelogRepository,
+  val referralRepository: ReferralRepository
+) {
+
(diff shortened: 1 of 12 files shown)


NeilLuptonMoJ:
Needs changes to `ReferralNotificationService` to send emails to notify SP

sldblog:
What is the scope of this ticket? It seems to introduce changes for all these fields
```
  var furtherInformation: String? = null,
  var additionalNeedsInformation: String? = null,
  var accessibilityNeeds: String? = null,
  var needsInterpreter: Boolean? = null,
  var interpreterLanguage: String? = null,
  var hasAdditionalResponsibilities: Boolean? = null,
```
And that seems to be a _lot_.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/AmendReferralService.kt
```diff
@@ -57,4 +58,50 @@ class AmendReferralService(
...
+    referral.hasAdditionalResponsibilities = update.hasAdditionalResponsibilities
+    referral.whenUnavailable = update.whenUnavailable
+    referral.furtherInformation = update.furtherInformation
+
+    val changelog = Changelog(
+      referral.id,
+      UUID.randomUUID(),
+      "NEED_AND_REQUIREMENT",
+      oldValue,
+      newValue,
```
sldblog, line 96: 🙅 _Technical issue_: We cannot serialise DTOs into the JSON fields because it would create a schema dependency

🙅 I think the team decided to serialise the values as arrays, which requires us to store each field change independently, with a separate reason for the change.

E.g. `referral_id, UUID, "additionalNeedsInformation", [old], [new], ...`

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/AmendReferralService.kt
```diff
@@ -57,4 +58,50 @@ class AmendReferralService(
...
+    referral.needsInterpreter = update.needsInterpreter
+    referral.interpreterLanguage = update.interpreterLanguage
+    referral.hasAdditionalResponsibilities = update.hasAdditionalResponsibilities
+    referral.whenUnavailable = update.whenUnavailable
+    referral.furtherInformation = update.furtherInformation
+
+    val changelog = Changelog(
+      referral.id,
+      UUID.randomUUID(),
+      "NEED_AND_REQUIREMENT",
```
sldblog, line 94: The `topic` would probably be good if this were an enum to make sure we're consistent

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/ActionPlanServiceTest.kt
```diff
@@ -58,6 +56,7 @@ internal class ActionPlanServiceTest {
     actionPlanEventPublisher,
     deliverySessionService,
     deliverySessionRepository,
+
```
kth13, line 59: probably this space is not required

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/controller/AmendReferralControllerTest.kt
```diff
@@ -64,5 +65,56 @@ internal class AmendReferralControllerTest {
...
+      )
+      val exception = assertThrows<ResponseStatusException> {
+        amendReferralController.updateNeedsAndRequirements(referral.id, needsAndRequirements, token)
+      }
+      assertThat(exception.status).isEqualTo(HttpStatus.BAD_REQUEST)
+      assertThat(exception.reason).isEqualTo("needs and requirements could not be updated")
+    }
+
+    @Test
+    fun `updateNeedsAndRequirement update needs and requirements with valid referral and fields `() {
```
kth13, line 91: probably more scenarios to check other needs and requirements as well

PR #1103, opened 2022-07-26
Amend desired outcomes

kth13:

- allow the user to amend the desired outcomes for a service category.
- updates the current referral to the new desired outcomes
- captures the audit information in the change log table

- The user needed the ability the amend the referrals after it is been assigned as it helps them to correct their mistakes and avoid re-referring again

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/AmendReferralService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/AmendReferralService.kt
@@ -1,0 +1,139 @@
+package uk.gov.justice.digital.hmpps.hmppsinterventionsservice.service
+
+import org.springframework.data.repository.findByIdOrNull
+import org.springframework.http.HttpStatus
+import org.springframework.security.oauth2.server.resource.authentication.JwtAuthenticationToken
+import org.springframework.stereotype.Service
+import org.springframework.web.server.ResponseStatusException
+import org.springframework.web.server.ServerWebInputException
+import uk.gov.justice.digital.hmpps.hmppsinterventionsservice.authorization.UserMapper
+import uk.gov.justice.digital.hmpps.hmppsinterventionsservice.dto.AmendComplexityLevelDTO
+import uk.gov.justice.digital.hmpps.hmppsinterventionsservice.dto.AmendDesiredOutcomesDTO
+import uk.gov.justice.digital.hmpps.hmppsinterventionsservice.dto.ReferralAmendmentDetails
+import uk.gov.justice.digital.hmpps.hmppsinterventionsservice.events.ReferralEventPublisher
+import uk.gov.justice.digital.hmpps.hmppsinterventionsservice.jpa.entity.Changelog
+import uk.gov.justice.digital.hmpps.hmppsinterventionsservice.jpa.entity.Referral
+import uk.gov.justice.digital.hmpps.hmppsinterventionsservice.jpa.entity.SelectedDesiredOutcomesMapping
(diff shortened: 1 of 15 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/AmendReferralService.kt
```diff
@@ -1,44 +1,59 @@
...
+  private val serviceCategoryRepository: ServiceCategoryRepository,
   private val userMapper: UserMapper,
   private val referralService: ReferralService
 ) {
 
-  fun updateComplexityLevel(referral: Referral, update: AmendComplexityLevelDTO, authentication: JwtAuthenticationToken): AmendComplexityLevelDTO? {
+  fun updateComplexityLevel(referral: Referral, update: AmendComplexityLevelDTO, serviceCategoryId: UUID, authentication: JwtAuthenticationToken): Changelog? {
+
+    if (referral.approvedActionPlan != null) {
+      throw ServerWebInputException("desired outcomes cannot be updated: the action plan is already approved")
```
JakeGoodwin22, line 43: should say 'complexity level cannot be updated'

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/AmendReferralService.kt
```diff
@@ -49,7 +64,70 @@ class AmendReferralService(
...
+      oldValue,
+      newValue,
+      amendDesiredOutcomesDTO.reasonForChange,
+      OffsetDateTime.now(),
+      actor
+    )
+    changelogRepository.save(changelog)
+    val savedReferral = referralRepository.save(referral)
+    referralEventPublisher.referralDesiredOutcomesChangedEvent(savedReferral)
+    return changelog
```
NeilLuptonMoJ, line 130: should we be returning the changelog entry?  The user of the endpoint is trying to update the referral, the changelog entry is created as a side effect of this update, so that isn't what I'd expect to see sent back

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/AmendReferralService.kt
kth13: Its a copy-paste error, will update it

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/AmendReferralService.kt
kth13: we are not returning the changelog back to the user. It just returns back to the controller. We then return this object

NeilLuptonMoJ: I don't need any of the values for the front end, it doesn't use the return values anyway.  
But for a user of this endpoint I don't think that would be an expected response - they are trying to amend a referral, the changelog details are created as a side effect so I wouldn't expect that to be the response (and if they are hitting amend-desired-outcomes why do they need amendTopic in the response?)

kth13: Though its extracting from the changelog, its actually the value that is been updated that has been passed back. I agree with you saying we don't need amendTopic but if that is case, we don't need other values also to be send back. Instead we can simply send a string message "Details Amended" with a 204 status code or else send back the ReferralDTO object.

NeilLuptonMoJ (approved):
👍

PR #1118, opened 2022-08-04
Adding new ingress

Froom2:

This is the first step of updating the ingress as per https://user-guide.cloud-platform.service.justice.gov.uk/documentation/other-topics/Switch-ingress-to-v1-ingress-controller.html

This is a required update.


kth13 (approved):
LGTM

PR #1119, opened 2022-08-04
Removing if statement in ingress

Froom2:

removes an unnecessary if statement in the ingress

the build was failing.


kth13 (approved):
LGTM

PR #1131, opened 2022-08-09
rename the alter changelog table file name to maintain the sequence f…

kth13:
…or flyway migration

rename the alter changelog table file name to maintain the sequence for flyway migration

The dev deployment is broken due to the sequence of the file changed


NeilLuptonMoJ:
not sure we should be adding `outOfOrder` just to get this one through, would it not be better to delete the v101 file (instead of renaming it) and add v100 as a new file?  Dev would have to be manually updated to remove the v101 flyway entry if it's already been created

kth13:
@NeilLuptonMoJ outOfOrder  is a temporary measure to resolve this issue. You mean to say add 100 as an empty file?

kth13:
This pull request is not required anymore as the issue in dev is resolved after we cleanup the row in flyway migration table

PR #1132, opened 2022-08-09
Fixing Performance issue when withdrawn referrals are shown in SP das…

kth13:

- Fixing Hibernate HHH000104 firstResult maxResults warning. 
- Now retrieving the data is a two step process, first we retrieve only the ID's using the Spec query
- Then we retrieve the entity using the Id's which we got as a result of the previous query

- When the fix for hiding withdrawn referrals was introduced in production, the performance issue started happening where the entire service was crashing because Hibernate did lots of pagination in memory. This fix was done as part of recommendation to fix the HHH000104 error.

## Performance testing

- As part of our team discussion, we decided to do performance test for the dashboard to make sure the new application is performing well after this change is introduced.


OmotayoTalabi (approved):

sldblog:
@kth13 Can this be rebased and merged? Is there something preventing this from going live?

PR #1134, opened 2022-08-11
all case search by reference number

kth13:

- introduce all case search by reference number. 

- we had case search by person on probation but we had a requirement where we need provide the capability to the user to search all cases by reference number


JakeGoodwin22 (approved):
LGTM

PR #1135, opened 2022-08-12
End of service report fix

OmotayoTalabi:
to fix end of service report bug 

Unable to complete EoSR. All appointments have an outcome but not giving them the option to complete an EoSR.


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/DeliverySessionService.kt
```diff
@@ -221,7 +221,9 @@ class DeliverySessionService(
       deliusAppointmentId = deliusAppointmentId,
       createdAt = OffsetDateTime.now(),
       createdBy = existingAppointment?.createdBy ?: authUserRepository.save(updatedBy),
-      referral = existingAppointment?.referral ?: session.referral
+      referral = existingAppointment?.referral ?: session.referral,
+      superseded = session.appointments.size> 0
```
kth13, line 225: What if there are two separate appointments(both will have superseded=false), then this case would be true.

kth13 (approved):
LGTM

PR #1155, opened 2022-08-22
Fix for session feedback link not going to correct page

JakeGoodwin22:

Fix for session feedback link not going to correct page when an email is sent

Currently we have a bug where the link in emails sent to PP's are wrong for the feedback


kth13 (approved):
LGTM

PR #1160, opened 2022-08-25
pact test fix

NeilLuptonMoJ:

create referral for previous day to prevent `APPOINTMENT_TIME_BEFORE_REFERRAL_START_DATE` error

Pact test is creating a past appointment (at start of current day) but actionplancontract is setting up referral with start date of now, so creating past appointment fails


kth13 (approved):
LGTM

PR #1161, opened 2022-08-25
End of service report

markmoj:

fixes the issue where superseded was not set correctly in db

to fix the superseded issue


src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/DeliverySessionServiceTest.kt
```diff
@@ -196,7 +205,38 @@ class DeliverySessionServiceTest @Autowired constructor(
...
+        it, ""
+      )
+    }
+
+    val appointment = deliverySession?.appointments?.first()
+
+    assertThat(appointment?.superseded).isFalse
+  }
+  @Test
+  fun `superseded flag set to true when pop did not attended appointment`() {
```
sldblog, line 225: Can we replace the acronym with the expanded version, please? It would help future developers a lot

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/DeliverySessionServiceTest.kt
```diff
@@ -196,7 +205,38 @@ class DeliverySessionServiceTest @Autowired constructor(
       assertThat(error.message).contains("can't schedule new appointment for session; latest appointment has no feedback delivered")
     }
   }
+  @Test
+  fun `superseded flag set to false when pop attended appointment`() {
```
sldblog, line 209: It looks like only `Attended.{YES, NO}` cases are tested. Can we test the `LATE` case?

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/DeliverySessionServiceTest.kt
```diff
@@ -355,18 +395,6 @@ class DeliverySessionServiceTest @Autowired constructor(
...
-    @Test
-    fun `expect failure when appointment is in the future`() {
-      val deliverySession = deliverySessionFactory.createScheduled(appointmentTime = OffsetDateTime.now().plusMonths(2))
-      val referral = deliverySession.referral
-      val appointment = deliverySession.currentAppointment!!
-
-      val error = assertThrows<ResponseStatusException> {
-        deliverySessionService.recordAppointmentAttendance(referral.id, appointment.id, defaultUser, Attended.YES, "additionalInformation")
-      }
-      assertThat(error.message).contains("Cannot submit feedback for a future appointment [id=${appointment.id}]")
```
sldblog, line 368: Is this assertion no longer relevant?

sldblog (commented):
Sorry, I still out of touch a bit, this is only a cursory review -- thought I'd raise the initial set of questions early

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/DeliverySessionService.kt
```diff
@@ -265,9 +263,6 @@ class DeliverySessionService(
     additionalInformation: String?
   ): Pair<DeliverySession, Appointment> {
     var sessionAndAppointment = getDeliverySessionAppointmentOrThrowException(referralId, appointmentId)
-    if (sessionAndAppointment.second.appointmentTime.isAfter(OffsetDateTime.now())) {
```
kth13, line 268: This checks are recently added by Neil for fixing the bug where we should not allow to give session feedback to future appointments. I think we need this

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/DeliverySessionService.kt
```diff
@@ -322,10 +317,6 @@ class DeliverySessionService(
       throw ResponseStatusException(HttpStatus.UNPROCESSABLE_ENTITY, "can't submit session feedback unless attendance has been recorded")
     }
 
-    if (appointment.appointmentTime.isAfter(OffsetDateTime.now())) {
-      throw ResponseStatusException(HttpStatus.BAD_REQUEST, "Cannot submit feedback for a future appointment [id=${appointment.id}]")
```
kth13, line 326: Same as above

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/DeliverySessionService.kt
```diff
@@ -343,9 +334,6 @@ class DeliverySessionService(
   fun submitSessionFeedback(referralId: UUID, appointmentId: UUID, submitter: AuthUser): Pair<DeliverySession, Appointment> {
     var sessionAndAppointment = getDeliverySessionAppointmentOrThrowException(referralId, appointmentId)
     val appointment = sessionAndAppointment.second
-    if (appointment.appointmentTime.isAfter(OffsetDateTime.now())) {
```
kth13, line 346: Same as the above two comments

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/DeliverySessionServiceTest.kt
kth13: As we need this feature, please bring this back

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/DeliverySessionService.kt
NeilLuptonMoJ: yes this a regression, we need to keep this

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/DeliverySessionService.kt
```diff
@@ -109,7 +109,7 @@ class DeliverySessionService(
       appointmentTime = appointmentTime,
       durationInMinutes = durationInMinutes,
       referral = session.referral,
-      superseded = session.appointments.size >= 1
+      superseded = attended == null,
```
NeilLuptonMoJ, line 112: all new appointments are created with attended = null, this is marking them as superseded which isn't correct

NeilLuptonMoJ: Instead of these changes to `DeliverySessionService.kt` would it not be better to let `AppointmentService.kt` handle the superseded flag?

It has existing processing 
```
  private fun updateAppointment(
   ....
    oldAppointment.superseded = true
    appointmentRepository.save(oldAppointment)
   ...
```
which does exactly what we want, but this method isn't called in the current reschedule appointment processing - so either find which method is being used and update it to include the above, or refactor to use updateAppointment as part of the call.  What do you think @kth13 ?

NeilLuptonMoJ: alternatively (as deliverysessionservice already has `// TODO: Some code duplication here with AppointmentService.kt`)
in `DeliverySessionService.kt` line 227 something like 
```diff
    appointmentRepository.saveAndFlush(appointment)
+    existingAppointment.superseded = true
+    appointmentRepository.save(existingAppointment)
```
(repeated anywhere else we are creating new appointments where necessary)

should do it, shouldn't it?

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/DeliverySessionService.kt
markmoj: reverted

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/DeliverySessionService.kt
markmoj: reverted

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/DeliverySessionService.kt
markmoj: reverted

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/DeliverySessionServiceTest.kt
markmoj: reverted

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/DeliverySessionServiceTest.kt
markmoj: Late is the same as attended?

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/DeliverySessionService.kt
markmoj: After discussion with @kth13 this change would require another ticket because its quite a big change however superseded by default has been changed to only when rescheduled. see future commit.

PR #1179, opened 2022-09-06
Pact test for amend

OmotayoTalabi:

added pact test

to fix the pact test for amend referral for interpreter required


kth13 (approved):
LGTM

PR #1180, opened 2022-09-06
added pact test for amend

OmotayoTalabi:

pact test for amend

pact test for amend


kth13 (approved):
LGTM

PR #1206, opened 2022-09-26
♻️ Remove unnecessary draft fields from database

sldblog:

Removes unnecessary draft fields from the database, removing the potential to misuse them.

Queries against the 25 September 2022 snapshot of the database show that these fields were last updated on 25 August 2022: they have not been set since the live/draft split.

Consequently, they are safe to drop.

These columns are not used in the draft data processing.

They are unused and unreferenced but still appear in the data with values until 25 August 2022. This is unsafe as it can lead to misuse. (E.g. the analytical platform could build a [DBT](https://www.getdbt.com/) model on top of these fields.)

## Zero-downtime behaviour

> Dropping a column requires an exclusive lock on the table to update
> the catalog but does not rewrite the table. As long as the column
> isn’t in current use you can safely drop the column.

The draft referral code (already) does not use these columns.

> It’s also important to confirm that the column is not referenced by
> any dependent objects that could be unsafe to drop.

No indices and views exist that reference these columns.


kth13 (approved):
LGTM

sldblog:
Drat, forgot `reference_number`

PR #1217, opened 2022-10-04
Prepare environments for switching databases

sldblog:

Prepare environments for switching databases.

Keeps all databases connected to the existing one.

Parameterises the secret name so it can be swapped with a single line.


kth13 (approved):
LGTM

PR #1233, opened 2022-10-10
Refactor: Dereference event details in publisher

sldblog:

Dereference event details in publisher (see reasons in the intent section below).

This change uses `EndOfServiceReportEvent` as the example, as it only has one type (submitted).

The related event, publisher, and listener classes have all been pulled into standalone classes with matching test names.

Often enough, our `ApplicationListener` classes blow up with a `NullPointerException` due to some state assumption.

We need to detect assumptions about the state earlier. The likely best place to do so is at publishing because we know _exactly_ what state we expect when we, e.g. fire an event about a submitted end-of-service report.


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/events/EndOfServiceReportEventPublisher.kt
```diff
@@ -1,38 +1,25 @@
...
-      EndOfServiceReportEvent(
-        this, EndOfServiceReportEventType.SUBMITTED,
-        endOfServiceReport, getEndOfServiceReportUrl(endOfServiceReport)
+      EndOfServiceReportSubmittedEvent(
+        this,
+        endOfServiceReport.referral.id,
+        getEndOfServiceReportUrl(endOfServiceReport),
+        endOfServiceReport.id,
+        endOfServiceReport.submittedAt!!,
+        endOfServiceReport.submittedBy!!
```
sldblog, line 22: Main change: switch from passing down `EndOfServiceReport` to the event and pass down simpler data points: IDs, dates, and `AuthUser`. This forces the publisher to ensure we broadcast an event representing what happened.

sldblog:
Note: #1377 ends up with `ReferralConcludedIntegrationListener` which, after these are both merged, should migrate over to `EndOfServiceReportIntegrationListener`

sldblog:
I will leave this one open as I think it's serviceable with minimal work.
@ministryofjustice/hmpps-interventions-dev please review or close as deemed fit 🙇‍♂️

sldblog:
Closing as I left HMPPS. Please reopen the branch if this change is still useful.

PR #1239, opened 2022-10-13
new onboarding contract types

kth13:

- sets up new service category, contract types and complexity level for Dependency and Recovery west midlands

- we need to onboarding new contract types for Dependency and Recovery west midlands


rgforsyth (approved):
Looks good to me. I'm assuming all the id's are new?

PR #1237, opened 2022-10-11
Switch the dev environment to use PostgreSQL14

sldblog:
## What does this pull request do?

Switch the dev environment to use PostgreSQL 14

- Set up by https://github.com/ministryofjustice/cloud-platform-environments/pull/8809
- Migration script in https://github.com/ministryofjustice/hmpps-interventions-ops/pull/2

## What is the intent behind these changes?

We are migrating from PostgreSQL 10 (end of life in November 2022) to PostgreSQL 14 by switching to a new instance.

This step reconfigures the service so that it uses the new database configuration.

PR #1259, opened 2022-10-21
Switch the preprod environment to use PostgreSQL14

sldblog:

Switch the preprod environment to use PostgreSQL14

Also, switch the Sunday database refresh job to copy into the new preprod database, not the old

- Set up by https://github.com/ministryofjustice/cloud-platform-environments/pull/8810
- Migration script in https://github.com/ministryofjustice/hmpps-interventions-ops/commit/73ef83cb4f43949a0819fa2ccad5c7305f14f5cd
- Related: #1237 

We are migrating from PostgreSQL 10 (end of life in November 2022) to PostgreSQL 14 by switching to a new instance.

This step reconfigures the service so that it uses the new database configuration

### DIFF
--- a/helm_deploy/values-preprod.yaml
+++ b/helm_deploy/values-preprod.yaml
@@ -20,7 +20,7 @@ ingress:
 
 env_details:
   contains_live_data: true
-  postgres_secret: postgres
+  postgres_secret: postgres14
 
 env:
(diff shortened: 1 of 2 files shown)


kth13 (approved):
LGTM

PR #1260, opened 2022-10-24
Show all database versions in grafana

sldblog:

Show all database versions in [grafana](https://grafana.live.cloud-platform.service.justice.gov.uk/d/PyQ91ARnk/hmpps-refer-and-monitor-an-intervention?orgId=1).

**Before**: only PostgreSQL10 production database CPU is visible.
image

**After**: show all instances for preprod/prod and v10/v14 PostgreSQL.
image

So we can see and compare the usage of new/old PostgreSQL versions.


kth13 (approved):
LGTM

PR #1270, opened 2022-10-25
Stop testing against old version of PostgreSQL

sldblog:

Stops testing against the old version of PostgreSQL (v10).

We have completed the switchover to PostgreSQL 14. We no longer need to prove we work on v10.


kth13 (approved):
LGTM

PR #1271, opened 2022-10-26
Plot a smoothed 5m average over database CPU

sldblog:

Plot a smoothed 5m average over the database CPU

**Before**
image

**After**
image

To better observe typical usage over longer periods (removes most spikes in load).


kth13 (approved):
LGTM

PR #1272, opened 2022-10-26
Make frequency of requests visible on heatmaps

sldblog:

Changes heatmap from a single-colour scheme with a square root scale (good for showing if anything is happening) to a spectral colour scheme (good for identifying is something happens often or not).

Also, it hides zero values -- which will remove the "background" image and only render colour if there is a value.

| Before | After |
| --- | --- |
| image | image |

To better visualise how often something happens.


kth13 (approved):
LGTM

PR #1286, opened 2022-10-31
backfiling max enforceable days and completion days

kth13:

- creates a temporary postconstruct method which back fills the completion deadline and maximum enforceable days. 
- amends the changelog which has null values in the oldvalue column

- There are old completion deadline and maximum enforceable days in the referral details table which needs to be ported to change log


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/config/ChangelogMigration.kt
```diff
@@ -0,0 +1,130 @@
...
+
+@Configuration
+class ChangelogMigration(
+  private val referralDetailsRepository: ReferralDetailsRepository,
+  private val changelogRepository: ChangelogRepository,
+  private val authUserRepository: AuthUserRepository
+) {
+  companion object : KLogging()
+
+  @PostConstruct
```
sldblog, line 25: I like the `@PostConstruct` idea!

Unfortunately, our runtime environment requires us to make this code idempotent and concurrency-safe.

At startup, we have six instances of the application in production starting **concurrently**:
- 3x hmpps-interventions-service-api
- 2x hmpps-interventions-service-dashboard
- 1x hmpps-interventions-service-performance-report
- The NDMIS CronJob, when it starts, will also execute this.

Additionally, the pods have limited time to boot up, and Kubernetes might kill and restart them if they don't start up quickly enough, so this has to be **performant or async**.

It is probably less of a headache to make this a one-off `Job` instead of reasoning about the effects of such concurrency.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/config/ChangelogMigration.kt
```diff
@@ -0,0 +1,130 @@
...
+
+    // Get all the referral details where the superseded id not null
+
+    val referralDetails = referralDetailsRepository.findAll()
+
+    referralDetails.filter { it.supersededById != null }.forEach {
+      try {
+        val changeLogs = changelogRepository.findByReferralIdOrderByChangedAtDesc(it.referralId)
+        // Do this migration when there is no change log data
+        if (changeLogs.isEmpty()) {
```
sldblog, line 38: The changelog might contain entries from other amendments that happened; the emptiness check would miss those.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/config/ChangelogMigration.kt
```diff
@@ -0,0 +1,130 @@
...
+          migrateReferralToChangelogTable(it)
```
sldblog, line 39: _concurrency_

Possibility of stale data leading to multiple appends:

```kotlin
// first transaction
val changeLogs = changelogRepository.findByReferralIdOrderByChangedAtDesc(it.referralId)

// decision based on in-memory code
if (changeLogs.isEmpty()) {

  // second transaction
  migrateReferralToChangelogTable(it)
}
```

If 
- this code is running concurrently;
- someone amends referrals business as usual,

then the decision of emptiness might be stale already, and we would execute append on an already non-empty state.

Introducing a transaction boundary for `referral_detail` and `changelog` scoped to one specific referral ID would prevent concurrent updates until the check-and-write pattern finishes.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/jobs/oneoff/migration/ChangelogMigrationProcessor.kt
```diff
@@ -0,0 +1,83 @@
...
+        supersededReferralDetails.referralId,
+        UUID.randomUUID(),
+        AmendTopic.COMPLETION_DATETIME,
+        oldValue,
+        newValue,
+        supersededReferralDetails.reasonForChange,
+        it.createdAt,
+        authUserRepository.findById(supersededReferralDetails.createdByUserId).get()
+      )
+      return changelogRepository.save(changelog)
```
sldblog, line 62: Can we use the functions from where we create these changelogs in business-as-usual processes? Seems we can reuse that code almost 1:1

I think those functions should also be moved to `AmendReferralService` from `ReferralService`

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/jobs/oneoff/migration/ChangelogMigrationJobConfiguration.kt
```diff
@@ -0,0 +1,63 @@
...
+  @Bean
+  @JobScope
+  fun referralDetailsReader(
+    sessionFactory: SessionFactory,
+  ): HibernateCursorItemReader<ReferralDetails> {
+    // this reader returns referral entities which need processing for the report.
+    return HibernateCursorItemReaderBuilder<ReferralDetails>()
+      .name("changelogMigrationReader")
+      .sessionFactory(sessionFactory)
+      .queryString("SELECT r FROM ReferralDetails r")
```
sldblog, line 38: _nitpick_ If we ordered this by `createdAt ASC` then the progress of migration will be simpler to understand (if we want to monitor it)

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/jobs/oneoff/migration/ChangelogMigrationJobListener.kt
```diff
@@ -0,0 +1,33 @@
...
+
+  override fun beforeJob(jobExecution: JobExecution) {
+    logger.info("Referral Details to Change log Batch Job Starts")
+    logger.info("Number of change log items before Job is = {}", changelogRepository.count())
+    val changeLogs = changelogRepository.findAll()
+    val changeLogsFilteredElements = changeLogs.filter { it.topic == AmendTopic.COMPLETION_DATETIME || it.topic == AmendTopic.MAXIMUM_ENFORCEABLE_DAYS }
+    val changelogWithNull = changeLogs.filter { changelog -> changelog.oldVal.values.contains("null") }
+    logger.info("Number of change log items with null value before Job is = {}", changelogWithNull.size)
+    logger.info("Deleting all the change log with COMPLETION_DATETIME and MAXIMUM_ENFORCEABLE_DAYS")
+    changeLogsFilteredElements.forEach { changelogRepository.deleteById(it.id) }
```
sldblog, line 24: _not a blocker, user-experience impact_

As far as I know, this deletion will remove the changelog entries in a separate transaction, so users can observe these changelogs disappearing and refilling.

Aborting the jobs for any reason (deployment/relocation of kubernetes resources) will mean the intermediate state will linger around for longer.

We had about ~400 requests to changelogs in production, so this isn't an issue in this data migration.

Later, to support updates at any time in any traffic scenario, a `process` which changes one referral in one database transaction at a time would be more suitable.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/jobs/oneoff/migration/ChangelogMigrationWriter.kt
```diff
@@ -0,0 +1,16 @@
...
+import org.springframework.stereotype.Component
+import uk.gov.justice.digital.hmpps.hmppsinterventionsservice.jpa.entity.Changelog
+import uk.gov.justice.digital.hmpps.hmppsinterventionsservice.jpa.repository.ChangelogRepository
+
+@Component
+class ChangelogMigrationWriter(val changelogRepository: ChangelogRepository) : ItemWriter<Changelog?> {
+  companion object : KLogging()
+  override fun write(items: MutableList<out Changelog?>) {
+    logger().info("writing into the change log table. The total amount of data loaded into changelog is {}", items.size)
+    changelogRepository.saveAll(items.filterNotNull())
```
sldblog, line 14: Do we need the `saveAll`? The process above seems to `save` already

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/jobs/oneoff/migration/ChangelogMigrationProcessor.kt
```diff
@@ -0,0 +1,83 @@
...
+  private val changelogRepository: ChangelogRepository,
+  private val referralDetailsRepository: ReferralDetailsRepository,
+  private val authUserRepository: AuthUserRepository
+) : ItemProcessor<ReferralDetails, Changelog?> {
+  companion object : KLogging()
+
+  override fun process(referralDetails: ReferralDetails): Changelog? {
+    logger.info("processing referral details {} for referral id {}", referralDetails.id, referralDetails.referralId)
+    // check if any changlog is null
+    if (referralDetails.supersededById != null) {
```
sldblog, line 26: We can roll this into the original query to reduce the number of items processed (and could make it more evident with `supersededDetails: ReferralDetails` variable)

`.queryString("SELECT r FROM ReferralDetails r WHERE r.supersededById is not null")`?

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/jobs/oneoff/migration/ChangelogMigrationProcessor.kt
```diff
@@ -0,0 +1,83 @@
...
+      return try {
+        // Do this migration when there is no change log data
+        migrateReferralToChangelogTable(referralDetails)
+      } catch (e: Exception) {
+        // log and continue
+        logger.info("logging the referral details that failed migration {}, {}", referralDetails.id, e.message)
```
sldblog, line 32: Can we let the exception bubble up? I'm not sure `info` would be visible to our monitoring and we lose the original stacktrace this way

helm_deploy/hmpps-interventions-service/templates/cronjob-migrate-referral-details-to-changelog-report.yaml
```diff
@@ -0,0 +1,23 @@
+apiVersion: batch/v1
+kind: CronJob
+metadata:
+  name: migrate-referral-details-to-changelog
+spec:
+  schedule: "30 0 * * *"
```
sldblog, line 6: _not a blocker_
We discussed maybe gating this or making it on-demand -- I'm looking up what we can use for that

sldblog (commented):
Nice -- couple of operational comments

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/jobs/oneoff/migration/ChangelogMigrationJobListener.kt
```diff
@@ -0,0 +1,33 @@
...
+class ChangelogMigrationJobListener(
+  val changelogRepository: ChangelogRepository
+) : JobExecutionListener {
+  companion object : KLogging()
+
+  override fun beforeJob(jobExecution: JobExecution) {
+    logger.info("Referral Details to Change log Batch Job Starts")
+    logger.info("Number of change log items before Job is = {}", changelogRepository.count())
+    val changeLogs = changelogRepository.findAll()
+    val changeLogsFilteredElements = changeLogs.filter { it.topic == AmendTopic.COMPLETION_DATETIME || it.topic == AmendTopic.MAXIMUM_ENFORCEABLE_DAYS }
```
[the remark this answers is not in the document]
kth13: Changed to **`findByTopicIn`**

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/jobs/oneoff/migration/ChangelogMigrationJobListener.kt
kth13: Agreed David, In the future jobs, we will consider process using each referral

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/jobs/oneoff/migration/ChangelogMigrationProcessor.kt
kth13: changed the query to **`select rd from ReferralDetails rd where rd.supersededById is not null order by createdAt ASC`**

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/jobs/oneoff/migration/ChangelogMigrationProcessor.kt
kth13: thrown the exception

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/jobs/oneoff/migration/ChangelogMigrationProcessor.kt
kth13: We will not be able to move the logic from ReferralService to AmendReferralService because ReferralService is already referred by AmendReferralService. so, referring that back in ReferralService is creating cyclic dependency issues. So, I have created a separate class `ChangelogUtil` which has this function and is being reused

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/jobs/oneoff/migration/ChangelogMigrationWriter.kt
kth13: Now writer is just being a place holder

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/jobs/oneoff/migration/ChangelogMigrationProcessor.kt
sldblog: Thanks, @kth13. Could we name it something representing the behaviour, please? "Util" and "Helper" names are considered code smells, as it's undefined what behaviour belongs in them.

kth13: Renamed the Kotlin class `ChangelogMigrationService`

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/jobs/oneoff/migration/ChangelogMigrationProcessor.kt
```diff
@@ -0,0 +1,52 @@
...
+  override fun process(referralDetails: ReferralDetails): Changelog? {
+    logger.info("processing referral details {} for referral id {}", referralDetails.id, referralDetails.referralId)
+    // check if any changlog is null
+    return try {
+      // Do this migration when there is no change log data
+      migrateReferralToChangelogTable(referralDetails)
+    } catch (e: Exception) {
+      // log and continue
+      logger.error("logging the referral details that failed migration {}, {}", referralDetails.id, e.message)
+      throw e
```
sldblog, line 32: _comment is out of date_
The `throw e` seems to contradict "continue"

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/jobs/oneoff/migration/ChangelogMigrationProcessor.kt
```diff
@@ -0,0 +1,52 @@
...
+  private val changeLogMigrationService: ChangeLogMigrationService,
+  private val changelogRepository: ChangelogRepository,
+  private val referralDetailsRepository: ReferralDetailsRepository,
+  private val authUserRepository: AuthUserRepository
+) : ItemProcessor<ReferralDetails, Changelog?> {
+  companion object : KLogging()
+
+  override fun process(referralDetails: ReferralDetails): Changelog? {
+    logger.info("processing referral details {} for referral id {}", referralDetails.id, referralDetails.referralId)
+    // check if any changlog is null
```
sldblog, line 25: _comment is out of date_
We no longer do this check

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/ChangeLogMigrationService.kt
```diff
@@ -0,0 +1,48 @@
...
+
+  fun logChanges(referralDetails: ReferralDetails, amendValue: String, topic: AmendTopic, actor: AuthUser, reasonForChange: String): Changelog {
+    return if (topic == AmendTopic.COMPLETION_DATETIME) {
+      val oldValue = ReferralAmendmentDetails(listOf(referralDetails.completionDeadline.toString()))
+      val newValue = ReferralAmendmentDetails(listOf(amendValue))
+      processChangeLog(topic, oldValue, newValue, referralDetails.referralId, actor, reasonForChange)
+    } else {
+      val oldValue = ReferralAmendmentDetails(listOf(referralDetails.maximumEnforceableDays.toString()))
+      val newValue = ReferralAmendmentDetails(listOf(amendValue))
+      processChangeLog(AmendTopic.MAXIMUM_ENFORCEABLE_DAYS, oldValue, newValue, referralDetails.referralId, actor, reasonForChange)
```
sldblog, line 23: _unsafe branching_

Any `AmendTopic` can be passed into this function; if `topic != AmendTopic.COMPLETION_DATETIME` is true, we'll log it with `AmendTopic.MAXIMUM_ENFORCEABLE_DAYS` even if the topic isn't that.

While there's a `referralDetails` argument, it sounds reasonable to explicitly block invalid values at the point of `logChanges`, not before.

What do you think about either

- ensure that the topic one of `COMPLETION_DATETIME` or `MAXIMUM_ENFORCEABLE_DAYS`
- or detect the topic from `referralDetails`, by passing down the previous value? It would seem that would lead to better cohesion

sldblog (approved):
Looks good. Let's run it on dev to observe its effects. ([Related runbook](https://ministryofjustice.github.io/hmpps-interventions-docs/runbooks/trigger-scheduled-jobs.html))

The comments are not blocking in themselves -- but let's pick them up here or in follow-up PRs.

sldblog:
@kth13 This code doesn't have any flyway migrations -- do we need the label?

kth13:
Ok @sldblog , if that label is meant for that, we can remove the label. The reason why I added that label is to indicate that it will be better off to run this after working hours to have minimal impact on users

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/ChangeLogMigrationService.kt
kth13: Sure @sldblog I got your point. The reason I did this because of both point where this method is been called, its obvious that we will get only `COMPLETION_DATETIME` or `MAXIMUM_ENFORCEABLE_DAYS`. Sure, We will get this done in a different PR

PR #1292, opened 2022-11-02
rename migration version to avoid conflict

rgforsyth:

Updates the name of a migration to avoid a conflict.

Fix pipeline.


kth13 (approved):
LGTM

sldblog:
Bypassing db check, as it is broken due to running the broken main migrations under fix

PR #1308, opened 2022-11-10
Use `increase` instead of multipliers in Grafana

sldblog:

Use `increase` instead of rate times multipliers in Grafana.

`60*rate(metric{}[1m])` == `increase(metric{}[1m])`
(60 is the number of seconds in the time range `1m`)

Simplification.

From the [documentation][doc] (emphasis mine):

> `increase` should only be used with counters and native histograms where the components behave like counters. **It is syntactic sugar for `rate(v)` multiplied by the number of seconds** under the specified time range window, and should be used primarily for human readability. Use `rate` in recording rules so that increases are tracked consistently on a per-second basis.

[doc]: https://prometheus.io/docs/prometheus/latest/querying/functions/#increase

### DIFF
--- a/helm_deploy/hmpps-interventions-service/grafana/hmpps-interventions-dashboard.json
+++ b/helm_deploy/hmpps-interventions-service/grafana/hmpps-interventions-dashboard.json
@@ -288,7 +288,7 @@
       "targets": [
         {
           "exemplar": true,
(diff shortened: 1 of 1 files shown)


kth13 (approved):
LGTM

PR #1309, opened 2022-11-10
Remove old, unused ingress

sldblog:

Remove old, unused ingress

See also:

- https://github.com/ministryofjustice/hmpps-interventions-ui/pull/1794

Remove unused ingresses so we don't block migrations.

> [We are intending to proceed with Kubernetes 1.22 upgrade next week](https://mojdt.slack.com/archives/CH6D099DF/p1667207580078859).
> This is a final reminder for teams that still need to move to new Ingress Controllers and remove deprecated API versions.
> Here is the list of ingresses and namespaces that still need to be updated before Tuesday 15th November.


kth13 (approved):
LGTM

PR #1311, opened 2022-11-14
Restore old ingress (to allow migration of new ingress to different version)

sldblog:

Restores the old ingress.

It also turns off ModSecurity in the development environment.

The new Ingress resource must have `apiVersion: networking.k8s.io/v1`.

Restoring the old ingress ensures there will be one ingress up when the migration to the new version happens.


kth13 (approved):
LGTM

PR #1312, opened 2022-11-14
Migrate new ingress to v1

sldblog:

Migrate new ingress to v1

To be compatible with new ingresses.


kth13 (approved):
LGTM

PR #1342, opened 2022-12-02
Add two main unqualified end reasons

sldblog:

Add two main unqualified end reasons.

Based on user research feedback:

- When referrals achieve outcomes early
- When providers request the end


src/main/resources/db/migration/V1_122__add_uncaptured_end_reasons.sql
```diff
@@ -0,0 +1,3 @@
+INSERT INTO cancellation_reason (code, description)
+VALUES ('RCE', 'Referral completed early'),
+       ('SPR', 'Service provider request');
```
sldblog, line 3: Before this can go live:

- need agreement from the board to avoid it messing up reports (sounds like tight coupling)
- need to see its usage in measurement (dashboard)
- need approval from content
- introduce `hide_after` date so it can be a timed experiment

sldblog:
Closing until it becomes doable work.

PR #1348, opened 2022-12-12
Make changelog migration job callable by CronJob

sldblog:

Fixes #1286's job launching

This hook was missing, causing the
`migrate-referral-details-to-changelog` cronjob (instead of exiting with error!) to spin up the API instead and get stuck.


awhodgson256 (approved):
Looks good to me.

PR #1361, opened 2022-12-19
add flag for running migrate changelog

rgforsyth:

Adds a flag based to decide if to run the change log migration cronjob based on environment. 

Allows us to control when the cron job is run.


sldblog (approved):
Ran `helm template --namespace=hmpps-interventions-prod --set image.tag=FOR_DIFF hmpps-interventions-service hmpps-interventions-service --values=./values-prod.yaml` before and after and it shows this cronjob removed 👍

PR #1365, opened 2022-12-20
Save into auth table

rgforsyth:

Saves into the auth table whenever we are creating a referral detail object. (This is the result of having no foreign key on this table)

Enforced that there is always a row in the auth table for users performing actions regarding refferal details.

### DIFF
--- a/src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/ReferralServiceUnitTest.kt
+++ b/src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/ReferralServiceUnitTest.kt
@@ -209,6 +209,7 @@ class ReferralServiceUnitTest {
       )
 
       whenever(referralDetailsRepository.findLatestByReferralId(referral.id)).thenReturn(existingDetails)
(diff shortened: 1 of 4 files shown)


src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/ReferralServiceUnitTest.kt
```diff
@@ -209,6 +209,7 @@ class ReferralServiceUnitTest {
       )
 
       whenever(referralDetailsRepository.findLatestByReferralId(referral.id)).thenReturn(existingDetails)
+      whenever(authUserRepository.save(referral.createdBy)).thenReturn(referral.createdBy)
```
sldblog, line 212: _suggestion, not a blocker_
We can also use the pass-through-first-argument pattern, so tests aren't that coupled to the implementation:
```kt
whenever(authUserRepository.save(any())).then(AdditionalAnswers.returnsFirstArg<AuthUser>())
```

sldblog (approved):

sldblog:
Should we address the entity's ID fields as well?

PR #1370, opened 2022-12-28
Fix CPU graph scaling

sldblog:

It's showing 10000% where it should be showing 100%.

The change pulled in several other changes as well due to a Grafana upgrade.

The most important change is

```diff
- "unit": "percentunit"
+ "unit": "percent"
```

| Before | After |
| --- | --- |
|  image | image |

To have accurate monitoring.

## Please link to ticket

IPB-110


sldblog:
Given these are for operational monitoring, the worst that it can do is break one graph (and we can revert).

PR #1374, opened 2022-12-29
Deprecate action plan usage in delivery sessions

sldblog:

IPB-116: Our initial data models used a referral → action plan → sessions chain.

If I recall correctly when we implemented action plan revisions, it enabled "extending" the number of sessions, so we had to tie the session counts to the referral instead.

This means all of the `actionPlanId` looks are deprecated and should be removed.

To avoid building on outdated abstractions.


sldblog:

PR #1375, opened 2022-12-29
Move counting delivery sessions closer to its home

sldblog:

IPB-116: These two functions were not even using `ActionPlan` in the `@Query` definitions.

Moved from `ActionPlanRepository` to `DeliverySessionRepository`

Avoid misleading and out-of-date abstractions.


sldblog:
Given these are fully tested, I think this was safe to do.

PR #1376, opened 2022-12-29
Remove warnings from session-related code

sldblog:

Noticed as part of IPB-116: Remove warnings from session-related code.

Please review **commit-by-commit** for specific things fixed.

Reduce the number of warnings in the code.


awhodgson256 (approved):
Mainly looks to be removal of unused values and move from var to val for other fields.  Looks good.

PR #1377, opened 2022-12-30
Terminate NSIs as early as possible

sldblog:

IPB-84: Terminate NSIs on ending, not after end-of-service report submission.

**Please review commit-by-commit**: there are a dozen refactors in this pull request.

### Explicit states

Introduces `DeliveryState`, `ReferralEndState` and `ReferralConcludedState` because these concepts were intertwined in two functions. Having them separately can define relationships between them.

We can now define correlations in domain language, e.g.

```kt
enum class DeliveryState(val requiresEndOfServiceReport: Boolean, val concludedState: ReferralConcludedState) {
  MISSING_FIRST_SUBSTANTIVE_APPOINTMENT(false, ReferralConcludedState.CANCELLED),
  IN_PROGRESS(true, ReferralConcludedState.PREMATURELY_ENDED),
  // ...
```

Read as

- if the referral ended now, MISSING_FIRST_SUBSTANTIVE_APPOINTMENT delivery would not require an end-of-service report and would transition to CANCELLED state
- if the referral ended now, IN_PROGRESS delivery would require an end-of-service report and would transition to PREMATURELY_ENDED state

### New events

- `ReferralConcluder` now signals ending and concluding via events
- Two new events for specific listeners: `ReferralEndingEvent`, `ReferralConcludedEvent`

### Event listeners

- One new listener: `ReferralEndingIntegrationListener`
- Two listeners have been extracted from existing classes: `ReferralConcludedIntegrationListener`, `ReferralConcludedListener`

### Caveats

- URL change: prematurely ended and completed NSI terminations will now link to the referral details URL, as there is _no_ end-of-service report (EoSR) available at this point
- Idempotent: this `signalEnd` will trigger once at ending and once when the EoSR is submitted.
- ❗️ The "concluded" integration now tries to notify the probation practitioner (PP) that an EoSR has been submitted. This will happen **after** we have terminated the NSI. I expect these calls always to fail. If it does, we should remove it.

1. Probation officers are terminating our referrals in Delius, causing integration issues.
2. To disable and delegate that action to our service, we must ensure the termination is **near-instantaneous**.
3. Currently, service providers have five working days to submit the end-of-service report on prematurely ended referrals.
4. This change removes this delay by terminating the NSI on "cancel" and not "EoSR submit".


sldblog:
> Idempotent: this `signalEnd` will trigger once at ending and once when the EoSR is submitted.

> ❗️ The "concluded" integration now tries to notify the probation practitioner (PP) that an EoSR has been submitted. I expect these calls always to fail. If it does, we should remove it.

@mxco86 Do you think these would be a problem?

awhodgson256 (commented):
Splitting into a dedicated object adds clarity and makes sense to move to a dedicated type that can represent a concluded intervention as we will be treating those differently.

awhodgson256 (commented):
DeliveryState appears to have simplified the branching logic and provides more readable code.

awhodgson256 (commented):
Similar to recent front end changes, it makes sense to deploy some of the "heavy lifting" within and enum instead of branching.

awhodgson256 (commented):
Clarifies when a referral can be concluded.

awhodgson256 (commented):
Clarification by removal of Cancellation from EoSR branching clarifies the intent of this event.

awhodgson256 (approved):
Overall code reviewed added messages indicating improved code clarity, happy with the changes.

sldblog:
@awhodgson256 I can't see the code to which these comments refer. Are they supposed to be with code segments?

sldblog:
Thank you!

PR #1385, opened 2023-01-10
Use the reusable helm chart for data-extractor

sldblog:
## What does this pull request do?

Use the reusable helm chart for the data-extractor.

Diff in Kubernetes resources:

```diff
diff --git a/helm_deploy.01before.yaml b/helm_deploy.02after.yaml
index 58e49119..ae7bc1f6 100644
--- a/helm_deploy.01before.yaml
+++ b/helm_deploy.02after.yaml
@@ -2929,23 +2929,34 @@ spec:
         - name: heap-dumps
           emptyDir: {}
 ---
-# Source: hmpps-interventions-service/templates/cronjob-data-extractor-analytics.yaml
+# Source: hmpps-interventions-service/charts/generic-data-analytics-extractor/templates/cronjob.yaml
 apiVersion: batch/v1
 kind: CronJob
 metadata:
-  name: data-extractor-analytics
+  name: hmpps-interventions-service-data-analytics-extractor
+  labels:
+    app: hmpps-interventions-service-generic-data-analytics-extractor
+    release: hmpps-interventions-service
 spec:
-  schedule: "0 1 * * *"
+  schedule: 0 0 * * *
+  successfulJobsHistoryLimit: 3
+  failedJobsHistoryLimit: 3
+  concurrencyPolicy: Forbid
+  # Check if we fail to create jobs within this 2 minute window and count as a missed schedule
+  # The Cron controller will stop creating jobs after 100 consecutive missed schedules
+  # https://kubernetes.io/docs/concepts/workloads/controllers/cron-jobs/#cron-job-limitations
+  startingDeadlineSeconds: 120
   jobTemplate:
     spec:
+      backoffLimit: 3
       template:
         spec:
           restartPolicy: "Never"
           containers:
-          - name: data-extractor-analytics
-            image: ministryofjustice/data-engineering-data-extractor:v1
+            - name: hmpps-interventions-service-generic-data-analytics-extractor
+              image: "ministryofjustice/data-engineering-data-extractor:v1"
               imagePullPolicy: Always
-            args: ["extract_psql_all_tables_to_jsonl.sh && transfer_local_to_s3.sh"]
+              args: ["extract_table_names.py && extract_pg_jsonl_snapshot.py && transfer_local_to_s3.sh"]
               env:
                 - name: PGHOST
                   valueFrom:
@@ -2983,10 +2994,175 @@ spec:
                       name: analytical-platform-reporting-s3-bucket
                       key: secret_access_key
[...]

PR #1386, opened 2023-01-10
pde-2118: use python script instead of bash script 

hemeshpatel-moj:

This change is related to the Data Extractor, I have changed which script is used to extract data from the data and move to the AP. 

The python scrip originally had memory issues this has now been resolved, so we want all consumers of the application to use the python script and it has better error handling - we want to decommission the bash script

The python script has better error handling compared to the bash script.

### DIFF
--- a/helm_deploy/hmpps-interventions-service/templates/cronjob-data-extractor-analytics.yaml
+++ b/helm_deploy/hmpps-interventions-service/templates/cronjob-data-extractor-analytics.yaml
@@ -13,7 +13,7 @@ spec:
           - name: data-extractor-analytics
             image: ministryofjustice/data-engineering-data-extractor:v1
             imagePullPolicy: Always
-            args: ["extract_psql_all_tables_to_jsonl.sh && transfer_local_to_s3.sh"]
+            args: ["extract_table_names.py && extract_pg_jsonl_snapshot.py && transfer_local_to_s3.sh"]
             env:
               - name: PGHOST
                 valueFrom:


helm_deploy/hmpps-interventions-service/templates/cronjob-data-extractor-analytics.yaml
```diff
@@ -13,7 +13,7 @@ spec:
           - name: data-extractor-analytics
             image: ministryofjustice/data-engineering-data-extractor:v1
             imagePullPolicy: Always
-            args: ["extract_psql_all_tables_to_jsonl.sh && transfer_local_to_s3.sh"]
+            args: ["extract_table_names.py && extract_pg_jsonl_snapshot.py && transfer_local_to_s3.sh"]
```
sldblog, line 16: Happy to approve, but could you make these changes in https://github.com/ministryofjustice/hmpps-helm-charts/tree/main/charts/generic-data-analytics-extractor as well, please, as we have #1385 open to replace this file, which would then lose these changes

sldblog: thank you! please let me know the new version and I can fold that into #1385

hemeshpatel-moj: Here is the PR to reflect the above: https://github.com/ministryofjustice/hmpps-helm-charts/pull/85

hemeshpatel-moj: new version is 0.1.4

sldblog: thanks! I will fold that into #1385 and close this PR

PR #1387, opened 2023-01-10
Persist auth user within user mapper

rgforsyth:

Always save the user.

We have had issues where actions have been performed without saving the current user to the auth table. This has caused issues later on with null values being present when other actions have been performed. This change will ensure that one of the first actions that happens in the controller(user mapper) is to save the user into the auth table to prevent this issue happening in the future.

### DIFF
--- a/src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/controller/CaseNoteControllerTest.kt
+++ b/src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/controller/CaseNoteControllerTest.kt
@@ -45,6 +45,7 @@ class CaseNoteControllerTest {
       val referralId = caseNote.referral.id
       val createCaseNoteDTO = CreateCaseNoteDTO(referralId = referralId, subject = "subject", body = "body")
 
+      whenever(authUserRepository.save(any())).thenReturn(user)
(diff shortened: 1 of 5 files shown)


sldblog:
> Stops users from not being saved in the auth table.

@rgforsyth For posterity, could you please add details about why this was a problem?

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/controller/CaseNoteControllerTest.kt
```diff
@@ -109,6 +113,7 @@ class CaseNoteControllerTest {
     fun `can get an individual case note`() {
       val id = UUID.randomUUID()
       val caseNote = caseNoteFactory.create(id = id, subject = "subject", body = "body")
+      whenever(authUserRepository.save(any())).thenReturn(authUserFactory.create())
```
sldblog, line 116: This is probably safer to re-use if we pass through the original argument rather than creating a new object:
```kt
whenever(authUserRepository.save(any())).then(AdditionalAnswers.returnsFirstArg<AuthUser>())
```
This way, we don't mutate the object that is potentially set up by the test -- and we can use this pattern instead of `.thenReturn(user)`.

kth13 (approved):
LGTM

PR #1388, opened 2023-01-10
Use "canStartEoSR" instead of "requiresEoSR"

sldblog:

Use "canStartEoSR" instead of "requiresEoSR" in the `ReferralConcluder`/end-of-service report condition tests.

This wording reflects the natural discussion about the feature.


kth13 (approved):
LGTM

PR #1400, opened 2023-01-13
Remove deprecated fields

sldblog:

**Migrated from #1034 for newer branch name**

Remove deprecated fields from the `referral` table
Remove unused field from the `draft_referral` table

Migration safety:

- `DROP COLUMN` requires an exclusive lock during the DDL change, but does not rewrite the table. So the lock won't be held long. Still, the referral table is used in **everything**, so let's do this migration out of hours (past 9 pm).
- `pg_depends` shows that only the two constraints are referencing these columns and they will be dropped automatically.

`deprecated_assigned_*` fields were a stepping stone in migrating assignments into `referral_assignment`. We left them in case we needed to roll back. They are no longer referenced or needed.

`maximum_enforceable_days`, `completion_deadline` and `further_information` are now in `referral_details` and not needed in either drafts or live referrals.

`reference_number` is not generated for `draft_referral`, so it does not have to exist.


kth13 (approved):
LGTM

sldblog:
deployed to production on 25 Feb 2023 (Saturday) at 20:54:08

PR #1401, opened 2023-01-16
Ipb 7/conclude referrals

marcocolluramoj:

[IPB-7](https://dsdmoj.atlassian.net/browse/IPB-7): Adds a job to conclude referrals which have cancellation requests and no attended sessions.

There is no longer an option to start an end-of-service report from the UI for referrals which have early end (cancellation) requests and have been marked as not attended. The job allows us to conclude these referrals.

[IPB-7]: https://dsdmoj.atlassian.net/browse/IPB-7?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/jobs/oneoff/concludereferrals/ConcludeReferralsJobListener.kt
```diff
@@ -0,0 +1,35 @@
...
+    when (jobExecution.status) {
+      BatchStatus.COMPLETED -> {
+      }
+
+      BatchStatus.FAILED -> {
+      }
+
+      else -> {
+        logger.warn(
+          "Unexpected status encountered for performance report {} {}",
```
sldblog, line 28: This doesn't seem to be related to this job. Do we need this listener at all? Would an exception be caught by Sentry by default?

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/jpa/repository/DeliverySessionRepository.kt
```diff
@@ -36,4 +36,11 @@ interface DeliverySessionRepository : JpaRepository<DeliverySession, UUID> {
       "and appt.appointmentFeedbackSubmittedAt is not null"
   )
   fun countNumberOfAttendedSessions(referralId: UUID): Int
+
+  @Query(
+    "select count(sesh) from DeliverySession sesh join sesh.appointments appt " +
+      "where sesh.referral.id = :referralId and appt.attended in ('NO') " +
+      "and appt.appointmentFeedbackSubmittedAt is not null"
+  )
+  fun countNumberOfNoAttendedSessions(referralId: UUID): Int
```
sldblog, line 45: _nit_
```suggestion
  fun countNumberOfNotAttendedSessions(referralId: UUID): Int
```

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/ReferralConcluder.kt
```diff
@@ -56,7 +58,9 @@ class ReferralConcluder(
   }
 
   private fun deliveryState(referral: Referral): DeliveryState {
-    val approvedActionPlan = referral.approvedActionPlan ?: return DeliveryState.NOT_DELIVERING_YET
+    val approvedActionPlan =
+      actionPlanRepository.findAllByReferralIdAndApprovedAtIsNotNull(referral.id).maxByOrNull { it.approvedAt!! }
```
sldblog, line 62: The derived query method (`findAllByReferralIdAndApprovedAtIsNotNull`) could incorporate the `max` query by using `OrderBy` and `Top`: https://docs.spring.io/spring-data/jpa/docs/current/reference/html/#repository-query-keywords (perhaps `findTopByOrderByApprovedAtDescByReferralId`?)

However, to simplify this and make more domain-readable, could we instead define `actionPlanRepository.findLatestApprovedActionPlan(referral.id)`?

sldblog (commented):
Looks good, the derived query method name is the only blocking comment

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/ReferralConcluder.kt
marcocolluramoj: Sure, I couldn't use top or limit in hql but I could use native query:
`  @Query("select * from action_plan where referral_id = :referralId order by approved_at desc limit 1", nativeQuery = true)`  
would this be fine or should we go with your first suggestion `findTopByOrderByApprovedAtDescByReferralId `

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/jobs/oneoff/concludereferrals/ConcludeReferralsJobConfiguration.kt
```diff
@@ -0,0 +1,60 @@
...
+    val validator = DefaultJobParametersValidator()
+    validator.setRequiredKeys(
+      arrayOf(
+        "timestamp",
+      )
+    )
+
+    return jobBuilderFactory["concludeReferralsJob"]
+      .incrementer(TimestampIncrementer())
+      .validator(validator)
```
kth13, line 41: What is the reason for validating the timestamp?

kth13 (approved):
LGTM

PR #1414, opened 2023-01-20
changing the branch name validator from dependabot to renovate

kth13:

updates the branch name validator to refer to renovate

The dependency library updates framework has been changed to renovate. Hence we need to change the branch name validator

### DIFF
--- a/.github/workflows/branch-name-validator.yml
+++ b/.github/workflows/branch-name-validator.yml
@@ -9,5 +9,5 @@ jobs:
       - uses: actions/checkout@v3
       - uses: deepakputhraya/action-branch-name@master
         with:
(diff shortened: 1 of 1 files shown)


.github/workflows/branch-name-validator.yml
```diff
@@ -9,5 +9,5 @@ jobs:
       - uses: actions/checkout@v3
       - uses: deepakputhraya/action-branch-name@master
         with:
-          regex: '^(no-ticket|dependabot|IPB-[0-9]+)/' # start branch with either 'no-ticket/' or 'dependabot' or 'IPB-NUMBER/'
-          ignore: main
+          regex: '^(no-ticket|renovate|IPB-[0-9]+)/' # start branch with either 'no-ticket/' or 'IPB-NUMBER/', allowing for bots
```
sldblog, line 12: TIL that renovate sometimes opens PRs with `renovate-` prefix, which wouldn't match here

sldblog (approved):
LGTM we can add the `-` here or later

PR #1436, opened 2023-01-26
Reduce frequency for gradle updates

sldblog:

[Update all dependencies once before work on Monday](https://github.com/ministryofjustice/hmpps-interventions-service/commit/82036e3c414723bd25dd2df0afd0c679cfbbe168) 
So that we can handle and merge it Monday first thing.

Reduce PR frequency (predominantly AWS dependencies).

Security updates will be raised independently by dependabot.


kth13 (approved):
LGTM

PR #1439, opened 2023-01-30
Store current location

marcocolluramoj:

Updates Draft Referral with personCurrentLocationType and personCustodyPrisonId fields and creates a record in referral_location table when the Draft Referral is sent.

To streamline the communication of the service user's current location between PPs and SPs


awhodgson256 (commented):
I'm not sure about the feature flags names in helm deploy yaml(s).  Won't it be looking for a FEATURE_FLAGS_....  Instead of "feature-flags.current-location.enabled" as defined in the annotation?

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/DraftReferralService.kt
```diff
@@ -108,6 +114,7 @@ class DraftReferralService(
     updateServiceUserNeeds(referral, update)
     updateDraftRiskInformation(referral, update)
     updateServiceCategoryDetails(referral, update)
+    updatePersonCurrentLocation(referral, update)
```
rgforsyth, line 117: There is a feature flag around the validate functionality but not this, meaning that it will just be stored without any validation at all. Should this also be inside the feature flag?

rgforsyth (approved):

PR #1440, opened 2023-01-31
Define the mapping between Referral and ReferralLocation entities. Up…

marcocolluramoj:
…date SentReferralDTO to include referral location fields

Defines the mapping between the Referral and ReferralLocation entities and updates the SentReferralDTO to include referral location fields.

To expose the person current location fields in the getSentReferral() endpoint so that they can be consumed by the front-end and displayed in the referral details page.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/jpa/entity/Referral.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/jpa/entity/Referral.kt
@@ -91,6 +91,7 @@ class Referral(
 
   @OneToOne(mappedBy = "referral") @Fetch(JOIN) var endOfServiceReport: EndOfServiceReport? = null,
   @OneToOne(mappedBy = "referral") @Fetch(JOIN) var supplierAssessment: SupplierAssessment? = null,
+  @OneToOne(mappedBy = "referral") @Fetch(JOIN) var referralLocation: ReferralLocation? = null,
 
   @OneToMany(fetch = FetchType.LAZY) @JoinColumn(name = "referral_id")
   private val referralDetailsHistory: Set<ReferralDetails>? = null,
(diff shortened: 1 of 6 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/DraftReferralService.kt
```diff
@@ -389,7 +389,7 @@ class DraftReferralService(
     communityAPIReferralService.send(referral)
 
     val sentReferral = referralRepository.save(referral)
-    createReferralLocation(draftReferral)
+    createReferralLocation(draftReferral, sentReferral)
```
rgforsyth, line 392: Not really an issue, but you could also do a single `referralRepository.save(referral)` if you put this line above and don't save inside the method (I think?). Alternatively we could just check that the whole thing is rolled back if the referral table was saved and then the location save failed? I guess both approaches would need to be tested

marcocolluramoj: I also thought it would be good to do a single save on the referralRepository but we need to save the referral first in order to map ReferralLocation (It's being done this way for SupplierAssessment a couple lines below).

rgforsyth: ah okay. Fair enough 👍

rgforsyth (approved):
Looks good!

marcocolluramoj:
Thanks Ryan

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/jpa/entity/Referral.kt
```diff
@@ -91,6 +91,7 @@ class Referral(
 
   @OneToOne(mappedBy = "referral") @Fetch(JOIN) var endOfServiceReport: EndOfServiceReport? = null,
   @OneToOne(mappedBy = "referral") @Fetch(JOIN) var supplierAssessment: SupplierAssessment? = null,
+  @OneToOne(mappedBy = "referral") @Fetch(JOIN) var referralLocation: ReferralLocation? = null,
```
sldblog, line 94: Do we know the expected performance cost of this new mapping? Asking as `Referral` is widely used across the codebase and now every materialisation of it would also make an extra join and allocate more memory

PR #1442, opened 2023-01-31
Update DraftReferralDTO with referral location fields

marcocolluramoj:

Populates DraftReferralDTO with referral location details from Draft Referral

To return the fields for the current location form in the UI


rgforsyth (approved):
Thanks

PR #1448, opened 2023-02-01
introducing multi platform builds

kth13:

configures the circle ci to build and push docker images for different platforms (linux/amd64,linux/arm64)

Building docker images for different architectures((ex: New Macbook M1) will help us use the images seamlessly

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -198,10 +198,8 @@ workflows:
       - validate_helm:
           context:
             - hmpps-interventions-dev-deploy
-      - hmpps/build_docker:
+      - hmpps/build_multiplatform_docker:
           name: build_and_publish_docker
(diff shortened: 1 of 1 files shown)


sldblog (commented):
The other (non-main build that runs on branch builds) was not changed to a multiplatform build. Is that intentional?

kth13:
> The other (non-main build that runs on branch builds) was not changed to a multi platform build. Is that intentional?

The build_multiplatform_docker automatically pushes the image into repo and that is the reason we can't use that in non-main build

sldblog (approved):
Thanks for clarifying! If we are happy with this level of build confidence, I'm happy

Edit: On second reading, this sounds like I'm hinting at something. I am not!

PR #1449, opened 2023-02-01
upgrading the docker file to use eclipse-temurin 19

kth13:

- upgrades the docker file to use eclipse-temurin:19-jre base image for the application
- The new multi architecture docker build is not working with the existing eclipse-temurin:17.0.6_10-jre-alpine base image and hence we have to upgrade it

The latest changes to make the application built multi architecture images  (https://github.com/ministryofjustice/hmpps-interventions-service/pull/1448) requires the base to be eclipse-temurin:19-jre.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM eclipse-temurin:17.0.6_10-jre-alpine AS builder
+FROM --platform=${BUILDPLATFORM:-linux/amd64} eclipse-temurin:17.0.6_10-jre-focal AS builder
 
 WORKDIR /app
 
@@ -23,17 +23,19 @@ RUN ./gradlew assemble
 
 
 # ---
-FROM eclipse-temurin:17.0.6_10-jre-alpine AS final
+FROM eclipse-temurin:17.0.6_10-jre-focal AS final
 LABEL maintainer="HMPPS Digital Studio <info@digital.justice.gov.uk>"
 
 # force a rebuild of `apk upgrade` below by invalidating the BUILD_NUMBER env variable on every commit
 ARG BUILD_NUMBER
 ENV BUILD_NUMBER ${BUILD_NUMBER:-1_0_0}
 
-RUN apk upgrade --no-cache && \
-     apk add --no-cache \
-       curl \
-       tzdata
+RUN apt-get update && \
+    apt-get -y upgrade && \
+    apt-get install -y \
+      curl \
+      tzdata \
+    && rm -rf /var/lib/apt/lists/*
 
 ENV TZ=Europe/London
 RUN ln -snf "/usr/share/zoneinfo/$TZ" /etc/localtime && echo "$TZ" > /etc/timezone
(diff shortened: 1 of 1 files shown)


awhodgson256 (commented):
I assume there has been a change of installer from apk to apt-get, was that a requirement of the new version? @kth13

Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM eclipse-temurin:17.0.6_10-jre-alpine AS builder
+FROM --platform=${BUILDPLATFORM:-linux/amd64} eclipse-temurin:19-jre-jammy AS builder
```
sldblog, line 1: Do we need to upgrade to Java 19?

[17.0.6_10-jre-focal](https://hub.docker.com/layers/library/eclipse-temurin/17.0.6_10-jre-focal/images/sha256-b10df4660e02cf944260b13182e4815fc3e577ba510de7f4abccc797e93d9106?context=explore) looks like it's available

sldblog: image

Dockerfile
```diff
@@ -23,17 +23,17 @@ RUN ./gradlew assemble
...
 ENV BUILD_NUMBER ${BUILD_NUMBER:-1_0_0}
 
-RUN apk upgrade --no-cache && \
-     apk add --no-cache \
-       curl \
-       tzdata
+RUN apt-get update && \
+    apt-get -y upgrade && \
+    apt-get install -y curl && \
+    apt-get install -y tzdata
```
sldblog, line 36:
```suggestion
    apt-get install -y \
      curl \
      tzdata \
    && rm -rf /var/lib/apt/lists/*
```
https://docs.docker.com/develop/develop-images/dockerfile_best-practices/#apt-get

> Below is a well-formed RUN instruction that demonstrates all the apt-get recommendations.
>
>     RUN apt-get update && apt-get install -y \
>         aufs-tools \
>         automake \
>         build-essential \
>         curl \
>         dpkg-sig \
>         libcap-dev \
>         libsqlite3-dev \
>         mercurial \
>         reprepro \
>         ruby1.9.1 \
>         ruby1.9.1-dev \
>         s3cmd=1.1.* \
>      && rm -rf /var/lib/apt/lists/*

Dockerfile
kth13: @sldblog , I will try this version but all the teams that has adopted the multi arch build has upgraded to java19. That was the reason for me to use java 19

sldblog: Sure, but it feels like we should separate those steps, so if something goes wrong, we know if was from the switch to Ubuntu or from the switch to 19

Dockerfile
kth13: Thanks David, I will update it accordingly

Dockerfile
```diff
@@ -23,17 +23,19 @@ RUN ./gradlew assemble
 
 
 # ---
-FROM eclipse-temurin:17.0.6_10-jre-alpine AS final
+FROM eclipse-temurin:17.0.6_10-jre-focal AS final
```
sldblog, line 26: _not a blocker_
Does this need `--platform=${BUILDPLATFORM:-linux/amd64}` too?

sldblog (approved):
LGTM

Dockerfile
kth13: Yes not required, changing it now.

PR #1451, opened 2023-02-03
Ipb 10/add expected release date

awhodgson256:

Adds expected release dates and missing reason fields to draft referral and referral_location.   Also adds validation logic to ensure referral cannot be sent without one of these fields being populated (this is feature flagged).  The fields themselves are nullable, but one of the two should be set in order to send a referral.

Allow PP's and SP's to share expected release date for a person in custody and associated data within the service, instead of having to do this manually.


kth13 (approved):
LGTM

PR #1452, opened 2023-02-03
configuring to use the latest trivy scan and scan only in main build

kth13:

- configure circle ci to use the latest trivy scan in the pipeline
- configure circle ci to run trivy scan only in main

-The new multi arch docker build is not working well with existing trivy scan config. So, we need to switch to the latest trivy scan config which will work only in main. So we need to disable it in non main branches


awhodgson256 (approved):
Looks good Kumar

PR #1456, opened 2023-02-07
add new custody location field to pact test data

rgforsyth:

Adds custody location default data to pact test state

To match the UI's expectation of the default data setup


kth13 (approved):
LGTM

PR #1462, opened 2023-02-13
fixing the integration issue and pact test

kth13:

- Fixing the issues found when integrating with submitting current location
- Integrating expected release date in Pact testing

- submit current location and expected release page needs to be integrated


awhodgson256 (approved):
Looks ok to me.

PR #1463, opened 2023-02-13
Fix NDMIS export job not running overnight

sldblog:

IPB-332: Fix timestamp not incremented in NDMIS job

Today I learned that `.incrementer` cannot be chained; it simply overwrites the previous value.

I also learned that I cannot test this behaviour end-to-end. Calling `jobLauncher.launchJob()` in
`NdmisPerformanceJobLauncherTestUtils` will complain about both arguments missing rather than calling this `incrementer`.

Local testing has resulted in a successful execution, though.


kth13 (approved):
LGTM

PR #1465, opened 2023-02-14
Ipb 201/add nomis release dates

awhodgson256:

Adds additional columns to the interventions database REFERRAL_LOCATION table to store expected release dates as supplied by NOMIS call (to be added)

Provide reference data that can be used to support decision on accuracy and validity of gathering data on release dates or sourcing it from nomis.  Will be used later on to illustrate graphically what data we have and how closely it meets expectations.


marcocolluramoj:
Could we add nomis_prison_id or should I do add it?

marcocolluramoj (approved):

PR #1467, opened 2023-02-14
Show worst performing API calls in a timeline

sldblog:

Show worst-performing API calls in a timeline.

| Before | After |
| --- | --- |
| image |  image

So that we can track which are the consistently worst-performing calls in a better visualisation


kth13 (approved):
LGTM

PR #1469, opened 2023-02-20
fixing the order of the complexity levels are being sent

kth13:

- Sets the order of complexity level to low, medium, high manually

- A medium complexity level description is updated recently and due to that the data that is written from DB was returning the values in low, high and medium complexity instead of returning in low, medium and high complexity

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/jpa/entity/ComplexityLevel.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/jpa/entity/ComplexityLevel.kt
@@ -21,12 +21,12 @@ data class ComplexityLevel(
   @Id val id: UUID,
   val title: String,
   val description: String,
-  @Type(type = "complexities") @Enumerated(EnumType.STRING) @Column(name = "complexity_level_title")
-  val complexityLevelTitle: ComplexityLevelTitle,
+  @Type(type = "complexities") @Enumerated(EnumType.STRING) @Column(name = "complexity")
+  val complexity: Complexity,
 )
 
-enum class ComplexityLevelTitle {
-  LOW_COMPLEXITY,
-  MEDIUM_COMPLEXITY,
-  HIGH_COMPLEXITY;
+enum class Complexity {
+  LOW,
+  MEDIUM,
+  HIGH;
 }
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/jpa/entity/ComplexityLevel.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/jpa/entity/ComplexityLevel.kt
@@ -1,12 +1,32 @@
 package uk.gov.justice.digital.hmpps.hmppsinterventionsservice.jpa.entity
 
+import com.vladmihalcea.hibernate.type.basic.PostgreSQLEnumType
+import org.hibernate.annotations.Type
+import org.hibernate.annotations.TypeDef
+import org.hibernate.annotations.TypeDefs
 import java.util.UUID
+import javax.persistence.Column
 import javax.persistence.Entity
+import javax.persistence.EnumType
+import javax.persistence.Enumerated
 import javax.persistence.Id
 
 @Entity
+@TypeDefs(
+  value = [
+    TypeDef(name = "complexities", typeClass = PostgreSQLEnumType::class)
+  ]
+)
 data class ComplexityLevel(
   @Id val id: UUID,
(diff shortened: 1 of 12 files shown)


sldblog (commented):
We make lots of decisions based on the _category_ of complexity.

We colour it differently on the UI; we have this ordering concern. So far, text matching has solved it, but that can break any time we change titles. Changing titles should not break visual cues and ordering guarantees.

If we had LOW, MEDIUM, and HIGH [enum values](https://www.postgresql.org/docs/14/datatype-enum.html) on the data, it would sort naturally in the enum's declared order (if we add the ordering annotation on the entity). It would also guarantee the content on the API level.

kth13:
@sldblog I was thinking enums would be cleaner way to do but it involves lots of changes in lots of places and that's the reason why I took this quicker approach. But I will change it according to your comments

sldblog:
Would adding a new column, filled from the current titles, and adding that as the default order on the complexity level collections limit the blast radius?

kth13:
> Would adding a new column, filled from the current titles, and adding that as the default order on the complexity level collections limit the blast radius?

@sldblog If I am understanding you currently, let's have a new column with enum datatype and use that for sorting the complexitylevel collection?

sldblog:
@kth13 Yeah. Would that work without too much impact on the rest of the codebase?

src/main/resources/db/migration/R__data_dictionary.sql
```diff
@@ -167,4 +167,6 @@ COMMENT ON TABLE changelog IS 'details about any amendments to an existing refer
 
 COMMENT ON TABLE referral_location IS 'details about the service user''s current location';
 
+COMMENT ON TABLE complexity_level IS 'adding a new column of type Enum to maintain order';
```
sldblog, line 170: Data analysts will read these comments to understand what information these fields convey.

They won't be interested in the use of these in sorting. From their perspective, it's more important that these categorical values represent the complexity of a referred pathway

sldblog (commented):
Looks good! A couple of comments about semantics

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/jpa/entity/ComplexityLevel.kt
```diff
@@ -1,12 +1,32 @@
...
+  value = [
+    TypeDef(name = "complexities", typeClass = PostgreSQLEnumType::class)
+  ]
+)
 data class ComplexityLevel(
   @Id val id: UUID,
   val title: String,
   val description: String,
+  @Type(type = "complexities") @Enumerated(EnumType.STRING) @Column(name = "complexity_level_title")
+  val complexityLevelTitle: ComplexityLevelTitle,
```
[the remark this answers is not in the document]
kth13: Agreed, renamed the name to `complexity`

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/jpa/entity/ComplexityLevel.kt
```diff
@@ -1,12 +1,32 @@
...
 )
+
+enum class ComplexityLevelTitle {
+  LOW_COMPLEXITY,
```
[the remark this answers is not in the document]
kth13: Updated it

src/main/resources/db/migration/R__data_dictionary.sql
kth13: updated to a more understandable statement

sldblog (approved):

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/DraftReferralServiceUnitTest.kt
```diff
@@ -357,7 +358,7 @@ class DraftReferralServiceUnitTest {
     @Test
     fun `updating selected service category with same set doesn't remove complexity level for same service categories`() {
 
-      val complexityLevel1 = ComplexityLevel(UUID.randomUUID(), "title", "description")
+      val complexityLevel1 = ComplexityLevel(UUID.randomUUID(), "title", "description", Complexity.LOW)
```
sldblog, line 361: 💭 After this pull request, we could deprecate the `title` field and assume it from the enum value (it's always the same, after all)

kth13: Yes true @sldblog , I will raise a ticket for that in the backlog

PR #1470, opened 2023-02-20
IPB-282/make-transaction-annotations-consistent

awhodgson256:
Opting for spring's transactional annotation as it is

Specific to spring boot containers
Contains more options surrounding the transaction should we need them

Advice is not to mix javax and spring transactional annotation imports

https://stackoverflow.com/questions/26387399/javax-transaction-transactional-vs-org-springframework-transaction-annotation-tr

Makes all transaction imports across the application consistent (opting to use Spring annotation)

We are having issues with transaction integrity in the production platform.
This is intended to ensure all transactions are Atomic, Consistent, Isolated and Durable.  Production data will be monitored following deployment to ensure issues around data integrity have improved.


sldblog:
> Advice is not to mix javax and spring transactional annotation imports

👍 Could you link to the advice as well, please?

sldblog (approved):
👍 for consistent usage

By default, should they not behave precisely the same, though? As far as I know, Spring implicitly supports the javax ones. Based on this, can we expect any behaviour change?

awhodgson256:
> > Advice is not to mix javax and spring transactional annotation imports
> 
> 👍 Could you link to the advice as well, please?
Link Added

PR #1474, opened 2023-02-22
Remove redundant flush calls

awhodgson256:

saveAndFlush is called twice in the update referral details code, but result isn't read in either record.

Choosing to rely on @Transactional behaviour instead to commit or rollback.

Fixed tests with verify calls expecting saveAndFlush > now expecting save.

Make updates to referral transactional and atomic so we don't get insert of new referral without making the old referral details superseded.   Avoid redundant intermediate flushes and improve performance.


kth13 (approved):
LGTM

PR #1480, opened 2023-02-27
Add database connection acquire time graph

sldblog:

Add database connection acquire time graph.

|Before|After|
|--|--|
| image |  image

This was useful in the past to correlate endpoint slowness. It looks like at random intervals, we have a couple of seconds long delay in acquiring new connections. Consequently, this slows down endpoint responses, which can bubble into timeouts.

## What might be next?

_not included in this PR_

`hikaricp_connections_usage_seconds_sum` might be good for identifying sudden expensive queries:

image

This _might_ correlate to this alert:

image


kth13 (approved):
LGTM

PR #1482, opened 2023-02-28
Update migration version to match main

marcocolluramoj:

Updates flyway migration file from V1_130__add_nomis_prison_id.sql to V1_131__add_nomis_prison_id.sql

To fix conflict with latest migration version on main


kth13 (approved):
LGTM

PR #1484, opened 2023-03-02
Add columns to appointment table to support data cleanup

awhodgson256:
New columns will allow a self referring FK to be added later (not in this commit)

Also allow a column to flag data created by duplicate submits to be "stale" and thus not shown from JPA joins (again to be added later).

Adds new columns to the appoinment table to support data cleanup activities

To allow only relevant data to be selected from the appoinment table, and allow a constraint to be added later to keep data self referring and consistent.


kth13 (approved):
LGTM

PR #1485, opened 2023-03-02
Allow Prisoner release date fields to be null

marcocolluramoj:

Updates Prisoner release dates fields to be nullable

To prevent errors occurring when parsing the Prisoner response with some null release date fields from prisoner-offender-search preventing the non-null fields being recorded.


awhodgson256 (commented):
Looks ok as discussed

awhodgson256 (approved):

PR #1489, opened 2023-03-13
Ipb 262/onboard contracts and providers with task

kth13:

- creates a job to onboard the new contract
- The job reads the jsons present in folders `resources/contracts, resources/desiredoutcomes, resources/providers`
- feeds those jsons to the `upsertContractsJob` 

- This was done to adapt the existing onboarding solution to kotlin. 
- The intent was to avoid the onboarding in two different places with two different solutions.


amolphalnikar (approved):
Looks good to me. Please other devs can check review this as well. Thanks.

PR #1490, opened 2023-03-16
Ipb 399/withdrawn referrals

kth13:

- Fixing Hibernate HHH000104 firstResult maxResults warning.
- Now retrieving the data is a two step process, first we retrieve only the ID's using the Spec query
- Then we retrieve the entity using the Id's which we got as a result of the previous query

When the fix for hiding withdrawn referrals was introduced in production, the performance issue started happening where the entire service was crashing because Hibernate did lots of pagination in memory. This fix was done as part of recommendation to fix the HHH000104 error.

## Feature flag

- As part of our team discussion, we decided to introduce feature flag for this flag. If we tend to have performance problem again after this fix is introduced, we will disable it using the flag


awhodgson256 (commented):
This feels like a sizeable change Kumar, are we intending to cover a full regression in dev?  Other than that I can't see any specific issues.

PR #1497, opened 2023-03-20
Ipb 215/appointment cleanup superseded (create jobs)

awhodgson256:

Introduces two new spring batch jobs, intended to tag a target record as "stale" and back fill all superseded appointment records with a superseding appointment id, thus allowing a foreign key / check constraint to be added which will prevent duplicates.

To improve data integrity and in the case of the stale record, allow it to be filtered from the EoSR counts and thus the report to be generated.


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/jobs/oneoff/linksupersededappointments/LinkSupersededAppointmentsReader.kt
```diff
@@ -0,0 +1,22 @@
...
+@JobScope
+class LinkSupersededAppointmentsReader(
+  sessionFactory: SessionFactory,
+) : HibernateCursorItemReader<Appointment>() {
+
+  init {
+    this.setName("linkSupersededAppointmentsReader")
+    this.setSessionFactory(sessionFactory)
+    this.setQueryString(
+      "SELECT a FROM Appointment a where superseded = true",
```
kth13, line 19: `SELECT a FROM Appointment a where superseded = true`  should  be `SELECT a FROM Appointment a where a.superseded = true` ?

PR #1498, opened 2023-03-20
Ipb 384/resolve vulnerabilities

jhevans:

Resolves some OWASP warnings by upgrading components to non-vulnerable versions.

Fix the nightly build

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -9,6 +9,8 @@
 
 [*.kt]
 indent_size = 2
+ij_kotlin_allow_trailing_comma_on_call_site = true
+ij_kotlin_allow_trailing_comma = true
 
 [*.kts]
 indent_size = 2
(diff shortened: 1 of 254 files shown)


.editorconfig
```diff
@@ -9,6 +9,8 @@
 
 [*.kt]
 indent_size = 2
+ij_kotlin_allow_trailing_comma_on_call_site = true
+ij_kotlin_allow_trailing_comma = true
```
kth13, line 13: what is the significance of this config?

kth13 (changes requested):

.editorconfig
marcocolluramoj: > 

@kth13 These were automatically brought in from the updated dps-gradle-spring-boot plugin (4.8.4) and enforces usage of trailing commas

PR #1500, opened 2023-03-22
PI-916

anthony-britton-moj:
remove synchronous termination of referral via community api as sqs messages for referral ended will now be processed
delete some unused code related to community api


kth13 (approved):
LGTM

PR #1504, opened 2023-03-31
Replace throwing InvalidAssumptionErrors with logging

marcocolluramoj:

Logs the invalid assumption that service users always have a responsible officer, when requesting community-api for a responsible officer, instead of throwing an InvalidAssumptionError.

To prevent Sentry triggering Alerts for InvalidAssumptionErrors.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/CommunityAPIOffenderService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/CommunityAPIOffenderService.kt
@@ -158,9 +158,6 @@ class CommunityAPIOffenderService(
       }
     }
 
-    if (responsibleOfficers.isNullOrEmpty()) {
-      return null
-    }
-    return responsibleOfficers.first()
+    return responsibleOfficers?.firstOrNull()
   }
 }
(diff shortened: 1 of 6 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/CommunityAPIOffenderService.kt
```diff
@@ -158,6 +158,9 @@ class CommunityAPIOffenderService(
       }
     }
 
+    if (responsibleOfficers.isNullOrEmpty()) {
```
amolphalnikar, line 161: Does it make sense to have single return 
```suggestion
    return responsibleOfficers?.firstOrNull()
```

marcocolluramoj: Good spot, thanks Amol!

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/TelemetryService.kt
```diff
@@ -1,13 +1,17 @@
...
 
 @Service
 class TelemetryService(
   private val telemetryClient: TelemetryClient,
 ) {
+
+  companion object {
+    private val logger = KotlinLogging.logger {}
+  }
   fun reportInvalidAssumption(assumption: String, information: Map<String, String> = emptyMap(), recoverable: Boolean = true) {
```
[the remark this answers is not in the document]
marcocolluramoj: Yes, I also don't think we need it as it stands. I'll remove it.

amolphalnikar (approved):
Please someone else as well have a look at this.

PR #1507, opened 2023-04-04
add outcome file to s3 bucket push

rgforsyth:

adds the outcome file to the s3 bucket

as above


kth13 (approved):
LGTM

PR #1508, opened 2023-04-04
Retrieving end of service report if it exists before or else create it

kth13:

retrieves the end of service report if it already exists or else 

Due to the issues in end of service report, users trying to create end of service report again and facing DataIntegrityViolation exception

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/EndOfServiceReportService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/EndOfServiceReportService.kt
@@ -35,7 +35,7 @@ class EndOfServiceReportService(
       createdAt = OffsetDateTime.now(),
     )
     return endOfServiceReportRepository.findByReferralId(referralId)
-      ?: return endOfServiceReportRepository.save(endOfServiceReport)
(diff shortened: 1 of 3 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/EndOfServiceReportService.kt
```diff
@@ -34,8 +34,8 @@ class EndOfServiceReportService(
       createdBy = authUserRepository.save(createdByUser),
       createdAt = OffsetDateTime.now(),
     )
-
-    return endOfServiceReportRepository.save(endOfServiceReport)
+    return endOfServiceReportRepository.findByReferralId(referralId)
+      ?: return endOfServiceReportRepository.save(endOfServiceReport)
```
marcocolluramoj, line 38: can we omit the second return keyword here?

marcocolluramoj (approved):

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/EndOfServiceReportService.kt
kth13: Good spot Marco, will do that. Thanks

PR #1510, opened 2023-04-13
changing the API for retrieving the responsible officer

kth13:

- Change the community API for retrieving the responsible officer
- Minor refactoring of some code

- The existing community API is causing lots of Timeout when called. Community API team has suggested us to use a new slim API which can avoid Timeout

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/CommunityAPIOffenderService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/CommunityAPIOffenderService.kt
@@ -18,13 +18,6 @@ data class ServiceUserAccessResult(
   val messages: List<String>,
 )
 
-data class ResponsibleOfficer(
-  val firstName: String?,
-  val email: String?,
-  val staffId: Long?,
-  val lastName: String?,
-)
-
 private data class UserAccessResponse(
   val exclusionMessage: String?,
   val restrictionMessage: String?,
@@ -42,17 +35,21 @@ data class PrimaryIdentifiersResponse(
   val nomsNumber: String?,
 )
 
-private data class ContactableHumanResponse(
-  val forenames: String?,
-  val surname: String?,
-  val email: String?,
-  val phoneNumber: String?,
+data class OfficerDetails(
+  val communityOfficer: CommunityOfficer,
 )
 
-private data class OffenderManagerResponse(
-  val staff: ContactableHumanResponse?,
-  val staffId: Long?,
-  val isResponsibleOfficer: Boolean?,
+data class CommunityOfficer(
+  val code: String,
+  val name: OfficerName,
+  val username: String,
+  val email: String?,
+  val responsibleOfficer: Boolean,
+)
+
+data class OfficerName(
+  val forename: String,
+  val surname: String,
 )
 
 @Service
@@ -62,6 +59,7 @@ class CommunityAPIOffenderService(
   @Value("\${community-api.locations.staff-details}") private val staffDetailsLocation: String,
   @Value("\${community-api.locations.offender-managers}") private val offenderManagersLocation: String,
   @Value("\${community-api.locations.offender-identifiers}") private val offenderIdentifiersLocation: String,
+  @Value("\${refer-and-monitor-and-delius.locations.responsible-officer}") private val responsibleOfficerLocation: String,
   private val communityApiClient: RestClient,
   private val telemetryService: TelemetryService,
 ) {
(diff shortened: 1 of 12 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/CommunityAPIOffenderService.kt
```diff
@@ -128,36 +126,30 @@ class CommunityAPIOffenderService(
...
-          "service users always have a responsible officer",
-          mapOf("crn" to crn),
-          recoverable = false,
-        )
-        it > 1 -> telemetryService.reportInvalidAssumption(
-          "service users only have one responsible officer",
-          mapOf("crn" to crn, "numberOfResponsibleOfficers" to it.toString()),
-        )
-      }
+    if (responsibleOfficer == null) {
```
marcocolluramoj, line 146: I think we could use the responsibleOfficer flag here `|| !responsibleOfficer.communityOfficer.responsibleOfficer` to check if they are the responsible officer

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/CommunityAPIOffenderService.kt
```diff
@@ -128,36 +126,30 @@ class CommunityAPIOffenderService(
...
 
-  fun getResponsibleOfficer(crn: String): ResponsibleOfficer {
-    val offenderManagersPath = UriComponentsBuilder.fromPath(offenderManagersLocation)
+  fun getResponsibleOfficerDetails(crn: String): OfficerDetails? {
+    val officerDetailsPath = UriComponentsBuilder.fromPath(responsibleOfficerLocation)
       .buildAndExpand(crn)
       .toString()
 
-    val responsibleOfficers = communityApiClient.get(offenderManagersPath)
+    val responsibleOfficer = communityApiClient.get(officerDetailsPath)
```
marcocolluramoj, line 134: Should this be officerDetails instead of responsibleOfficer since we don't know whether they are the responsible officer yet at this point?

kth13: Agree, will update it. thanks

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/CommunityAPIOffenderService.kt
kth13: @marcocolluramoj I think we need to do both. Something like this 

`    if (officerDetails == null || !(officerDetails.communityOfficer.responsibleOfficer)) {
      telemetryService.reportInvalidAssumption(
        "service users always have a responsible officer",
        mapOf("crn" to crn),
        recoverable = false,
      )
    }`

The reason is we need to throw Invalid Assumption error when there is 404 as well

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/ReferralService.kt
```diff
@@ -264,26 +264,25 @@ class ReferralService(
...
-    return getResponsibleProbationPractitioner(referral.serviceUserCRN, sentBy, referral.createdBy)
+    return getResponsibleProbationPractitioner(referral.serviceUserCRN, referral.sentBy, referral.createdBy)
   }
 
   fun getResponsibleProbationPractitioner(crn: String, sentBy: AuthUser?, createdBy: AuthUser): ResponsibleProbationPractitioner {
     try {
-      val responsibleOfficer = communityAPIOffenderService.getResponsibleOfficer(crn)
-      if (responsibleOfficer.email != null) {
+      val responsibleOfficer = communityAPIOffenderService.getResponsibleOfficerDetails(crn)
+      if (responsibleOfficer?.communityOfficer?.email != null) {
```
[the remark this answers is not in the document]
kth13: @marcocolluramoj According to community api team, email is only nullable item and that's why we need to check that

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/CommunityAPIOffenderService.kt
marcocolluramoj: Sure, I agree

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/ReferralService.kt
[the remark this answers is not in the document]
marcocolluramoj: Sure, but don't we also need to check whether communityOfficer is the responsible officer since getResponsibleOfficerDetails() is returning the OfficerDetails. Perhaps we could change that to also return null if the communityOfficer isn't a responsible officer by checking the responsibleOfficer flag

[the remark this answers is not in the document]
kth13: @marcocolluramoj But we are now throwing InvalidAssumption if there is no responsible officer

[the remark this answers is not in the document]
marcocolluramoj: @kth13 But won't it continue to return a communityOfficer which may not be the responsible officer?

marcocolluramoj (approved):
LGTM

PR #1522, opened 2023-04-20
updating the api client for getting the responsible officer

kth13:

updating the api client for the getting the responsible officer

The api client for getting the responsible officer is changed and needs changing


marcocolluramoj (approved):
LGTM

PR #1523, opened 2023-04-20
Increase auth connect timeout to 20s

awhodgson256:

Increases auth connection time out to 20s

Reduce the number of timeout errors associated with user login

### DIFF
--- a/src/main/resources/application.yml
+++ b/src/main/resources/application.yml
@@ -208,7 +208,7 @@ webclient:
   hmpps-auth:
     max-retry-attempts: 2
     connect-timeout-seconds: 3
-    read-timeout-seconds: 3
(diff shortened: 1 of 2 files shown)


src/main/resources/application.yml
```diff
@@ -207,7 +207,7 @@ webclient:
   write-timeout-seconds: 10
   hmpps-auth:
     max-retry-attempts: 2
-    connect-timeout-seconds: 3
+    connect-timeout-seconds: 20
     read-timeout-seconds: 3
```
kth13, line 211: Hi @awhodgson256, is the read time out or the connection time out we need change. I guess the application is timing out while waiting for the response rather than connecting

kth13 (approved):
LGTM

PR #1524, opened 2023-04-20
updating the api model for the new api client

kth13:

updating the api model for the new api client

The api model is different to what it is now, it needs changing


rgforsyth (approved):
LGTM

PR #1529, opened 2023-04-26
PI-1013 update to use new endpoint

anthony-britton-moj:

Moves creation of new appointments to the new refer-and-monitor-and-delius api
Prevents duplicate appointments from being generated in delius
Allows the same endpoint to be used for location changes and rescheduling

Improve resilience for network or other technical errors
Prevent duplicates
Allow Delius to store a reference to the appointment in the Interventions Service rather than the Interventions Service storing a delius database id
Reduce errors when rescheduling due to community API not accepting an outcome


kth13 (approved):
LGTM

PR #1530, opened 2023-04-26
PI-1030 add prison manager details

anthony-britton-moj:
Add prison manager details when community manager is not the responsible officer

Allows retrieving the details of the prison manager and a check if they are the responsible officer when community manager is not the responsible officer

Reduce errors for not finding the responsible officer, in particular when the service user is in prison for example


kth13 (approved):
LGTM

PR #1537, opened 2023-05-05
Only track CustodyLocationLookup events for custody referrals

marcocolluramoj:

Prevents the NOMIS lookup and AppInsights tracking for custody location if the person being referred is not in custody.

To gather more accurate custody location comparison results in AppInsights.


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/listeners/CustodyLocationLookupListener.kt
```diff
@@ -18,7 +19,8 @@ class CustodyLocationLookupListener(
   val telemetryClient: TelemetryClient,
 ) : ApplicationListener<ReferralEvent> {
   override fun onApplicationEvent(event: ReferralEvent) {
-    if (event.type != ReferralEventType.SENT) {
+    if (event.type != ReferralEventType.SENT || event.referral.referralLocation?.type != PersonCurrentLocationType.CUSTODY
```
kth13, line 22: Is this not `if (event.type != ReferralEventType.SENT && event.referral.referralLocation?.type != PersonCurrentLocationType.CUSTODY` ?

kth13 (approved):
LGTM

PR #1540, opened 2023-05-10
PI-1013 added sentence id for fuzzy matching

anthony-britton-moj:

sends the sentence id to delius when creating appointments

allows delius to do a `fuzzy` search for older referrals where the external reference still isn't populated


kth13 (approved):
LGTM

PR #1543, opened 2023-05-11
update session attendance email with new fields

rgforsyth:

Adds the new data fields to the session non-attendance email

This adds the pop full name to the subject of the email


marcocolluramoj (approved):
LGTM

PR #1555, opened 2023-05-18
pass attendance details to community api

rgforsyth:

Passes attendance information to community api.

Currently community api is erroring as there are no attendance details being passed on an saa appointment reschedule


kth13 (approved):
LGTM

PR #1558, opened 2023-05-22
Fix for owasp check, upgrading spring parent

awhodgson256:

Upgrades spring parent project to address OWASP error in build

Prevent errors during OWASP checks and address security issue with referenced library.


kth13 (approved):
LGTM

PR #1559, opened 2023-05-22
Rename files to use lowercase environment names

awhodgson256:

Fix deployment issue encountered with previous PR

Allow deployment to work with correct case of deployment ENV variable


kth13 (approved):
LGTM

PR #1561, opened 2023-05-23
Read resource from classpath instead of resource reader

awhodgson256:
Fix for deploy issue

Changed reader to use a classpath rather than resource loader to pull files in deployed container

Fix for deployment issue seen in DEV


kth13 (approved):
LGTM

PR #1563, opened 2023-05-24
Move to using a yaml based comma separated line for appointment ids 

awhodgson256:
Move to using a yaml based comma separated line for appointment ids marked as stale

Moves from reading resource file based lists to using YAML file with comma separated string

Avoid issues with failing to read resource files in kubernetes environment


kth13 (approved):
LGTM

PR #1564, opened 2023-05-24
Remove mandatory jobParameter

awhodgson256:

Remove mandatory jobParameter and replace with the YAML file string value "appintment: ids"

Fix broken call due to parameter requirement (no longer required due to file based input)


kth13 (approved):
LGTM

PR #1565, opened 2023-05-25
Add branching to handle SAA appointments when linking superseded

awhodgson256:

Adds logic to handle linking of SAA appointments

Ensure all superseded appointments have a linked appointment so constraints can be applied preventing duplicates


kth13 (approved):
LGTM

PR #1566, opened 2023-05-26
Add a guard condition / logging to cope with exceptional appointment

awhodgson256:
Add a guard condition / logging to cope with exceptional appointment records

Adds a guard condition to linkSupersededAppointments job

Specifically where superseded appointment has created timestamp after the appointment that supersedes it...

Cope with / flag data that requires more detailed analysis and rectification, prevent failures in link superseded job runs.


kth13 (approved):
LGTM

PR #1568, opened 2023-05-30
Set parallelism to 1 for batch jobs to avoid locking issues

awhodgson256:

Sets parallelism to 1 for both cleanup jobs

Prevent multiple clients from attempting to update the same record.


kth13 (approved):
LGTM

PR #1569, opened 2023-05-30
Ipb 580/api submit probation practitioner details

kth13:

- captures the probation practitioner details during the draft stage
- saves the probation practitioner details when the referral is changed from draft to live referral
- captures telemetry details of the data to measure and quantify data later

For making a referral for a community with COM, we need save the probation practitioner details


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/DraftReferralService.kt
```diff
@@ -484,6 +508,108 @@ class DraftReferralService(
...
+          "result" to "valid delius details present",
+          "referralId" to draftReferral.id.toString(),
+        ),
+        null,
+      )
+      false -> {
+        telemetryClient.trackEvent(
+          eventName,
+          mapOf(
+            "result" to "valid delius details not present",
```
marcocolluramoj, line 546: Could the result be set to "invalid delius details" here? When I first read this, I thought this meant the delius details are valid but do not exist. And if hasValidDeliusPPDetails is true, we could probably also omit 'present' on line 537?

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/DraftReferralService.kt
```diff
@@ -484,6 +508,108 @@ class DraftReferralService(
...
+            "referralId" to draftReferral.id.toString(),
+          ),
+          null,
+        )
+        if (draftReferral.ppName == draftReferral.nDeliusPPName) {
+          telemetryClient.trackEvent(
+            eventName,
+            mapOf(
+              "result" to "delius name and user name matches",
```
marcocolluramoj, line 555: Could the if-else statements be refactored by doing the checks in the mapping? 

```suggestion
              "result" to if (draftReferral.ppName == draftReferral.nDeliusPPName) "delius name and user name matches" else "delius name and user name does not match",
```

marcocolluramoj (approved):

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/DraftReferralService.kt
kth13: Agreed and done

PR #1572, opened 2023-05-31
Use alternative method to get supplierAssessment object

awhodgson256:
in the cleaner due to nulls from previous call

Changes method of getting supplier assessment from a referral id in the cleaner job

Mark records with a superseded record which were previously null


kth13 (approved):
LGTM

PR #1578, opened 2023-06-02
fixing pact tests which is failing due to community with com

kth13:

- Fixes the pact test which is failing due to community with com

- Pact test were failing after introducing community with com feature


awhodgson256 (approved):
LGTM

PR #1583, opened 2023-06-05
PI-1103 action plan notifications

anthony-britton-moj:

Removes redundant code for synchronous call for action plan notifications

Improve resilience
Reduce errors


awhodgson256 (approved):
LGTM, as it's a clean up activity, fine to merge but will be applied to prod later due to current branch containing test code.

PR #1589, opened 2023-06-14
PI-1166 switch to RAM API for remaining read calls

anthony-britton-moj:

Switches the remaining calls to Community API to read data from the R&M&D API

Move away from deprecated endpoints in Community API
The minimal changes have been made to reduce changes elsewhere in the codebase, ie even though the response may not match exactly that from community api - the service converts the model before returning in order to keep remaining code the same.


kth13 (approved):
LGTM

PR #1593, opened 2023-06-15
adding the ndelius PDU for the location variable

kth13:

- Adding nDeliusPDU for the location

- nDeliusPDU needs to be included to show the location in the dashboard


awhodgson256 (approved):
LGTM

PR #1594, opened 2023-06-15
Add PoPs full name to Concerning behaviour email subject

marcocolluramoj:

Updates the Notify Template ID for Concerning behaviour emails and updates the key-value pairs to match the template variables. 

To add the PoP's full name to the subject for Concerning behaviour emails


rgforsyth (approved):
LGTM

PR #1598, opened 2023-06-23
Remove jobs for cleanup of superseded / stale records

awhodgson256:

Take away job code, YAML configuration for one off jobs that have now run in all environments.

Removes one-off redundant code for jobs which have now completed and are thus no longer needed to be maintained.


marcocolluramoj (approved):
LGTM

PR #1601, opened 2023-07-03
New session feedback fields

marcocolluramoj:

Adds new session feedback fields and updates endpoints.

First iteration of getting better session feedback.


rgforsyth (approved):
LGTM

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/AppointmentService.kt
```diff
@@ -390,21 +408,23 @@ class AppointmentService(
...
   }
 
-  private fun setAttendanceAndBehaviourIfHistoricAppointment(
+  private fun setAttendanceAndSessionFeedbackIfHistoricAppointment(
     appointment: Appointment,
     attended: Attended?,
-    additionalAttendanceInformation: String?,
-    behaviourDescription: String?,
+    attendanceFailureInformation: String?,
+    sessionSummary: String?,
```
kth13, line 415: Why sessionSummary and sessionResponse can be nullable and in later you are checking sessionSummary!! ?

src/main/resources/db/migration/V1_139__add_appointment_session_feedback.sql
```diff
@@ -0,0 +1,16 @@
+alter table appointment
+    add column attendance_failure_information text,
+    add column session_summary text,
+    add column session_response text,
+    add column session_concerns text,
+    add column session_feedback_submitted_by_id text,
+    add column session_feedback_submitted_at timestamp with time zone,
```
kth13, line 7: what about the existing columns which captures the feedback submission time

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/AppointmentService.kt
marcocolluramoj: sessionSummary!! and sessionResponse!! are wrapped in an attended?.let {} block - If the user attended then they would continue to the Session Feedback page where both sessionSummary and sessionResponse fields are required. Otherwise, they would submit the form by only needing to complete the Attendance Feedback page.

kth13: Ok understood, thanks for the clarification

kth13 (approved):
LGTM

PR #1604, opened 2023-07-12
Enable saving the probation practitioner details in prod

kth13:

- Enable saving the probation practitioner details in prod

The flag needs to be enabled to save the probation practitioner details


marcocolluramoj (approved):
LGTM

PR #1613, opened 2023-08-04
Log invalid assumptions instead of throwing InvalidAssumptionErrors

marcocolluramoj:

Logs invalid assumptions instead of throwing InvalidAssumptionErrors

To prevent Sentry alerts for invalid assumptions which can lead to alert fatigue


awhodgson256 (approved):
LGTM

PR #1615, opened 2023-08-08
fixing NPE exception when emailing for concluded events

kth13:

- checks if the current assignee is present before sending an email to them

- conclude events can be called when the user is not assigned as well.
- so the check needs to be done if the user is assigned before the email is sent
- There are around 21K errors due to this issue (https://ministryofjustice.sentry.io/issues/4172206305/events/0fab206c85804c3ca335b846a7df53a2/events/?project=5807819)


awhodgson256 (approved):
LGTM

PR #1631, opened 2023-08-21
Prevent AP submission before SAA complete

marcocolluramoj:

Adds validation to the submitDraftActionPlan endpoint to check that the supplier assessment has been attended and feedback has been submitted.

To prevent service providers from submitting an Action Plan before a Supplier Assessment Appointment has been attended and feedback has been submitted.


src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/component/ActionPlanValidatorTest.kt
```diff
@@ -96,8 +97,11 @@ class ActionPlanValidatorTest {
 
   @Test
   fun `submit action plan fails validation - number of sessions is empty`() {
-    val draftActionPlanId = UUID.randomUUID()
-    val actionPlanUpdate = SampleData.sampleActionPlan(id = draftActionPlanId, numberOfSessions = null)
+    val referral = SampleData.sampleReferral("CRN123", "Service Provider")
```
rgforsyth, line 100: Hey, no worries if you don't want to do this as it looks like the test was already using it, but I think we've stopped using 'sampledata' now and try to use the factories.

rgforsyth (approved):

PR #1640, opened 2023-09-12
add new endpoint for referral information

rgforsyth:

Adds a new endpoint to return various pieces of referral information, given a crn.

To allow the resettlement passport team to have access to our information.


marcocolluramoj (approved):
LGTM

PR #1641, opened 2023-09-12
Fixing the broken pact test

kth13:

- Fixes the pact test failing due to new fields introduced for allocated and unallocated COM work

- The pact test was broken after a couple of new fields was introduced for the allocated and unallocated COM work. This PR will fix those broken test


rgforsyth (approved):
Looks good. Thanks

PR #1651, opened 2023-10-02
Fixing issues in reading the files inside the docker container

kth13:

- Reads the onboarding files using the method that works both in IDE and jar

- We were not able to read the onboarding files from the docker container.


awhodgson256 (approved):
LGTM

PR #1662, opened 2023-10-11
Added service account for generic data analytics extractor

amolphalnikar:

This PR adds service account for generic-data-analytics-extractor

To fix the error while running the generic-data-analytics-extractor


awhodgson256 (approved):
LGTM

PR #1663, opened 2023-10-12
Update auth user username

marcocolluramoj:

Updates the username in the auth_user table.

To resolve an issue where the user is unable to view referrals they created due to the username having been changed in Delius and not matching with our records.


kth13 (approved):
LGTM

PR #1666, opened 2023-10-12
Updated generic-data-analytics-extractor chart version to 1.0.1

amolphalnikar:

This PR sets the chart version for generic-data-analytics-extractor to 1.0.1

To fix the error while running the generic-data-analytics-extractor


kth13 (approved):
LGTM

PR #1671, opened 2023-10-16
Map two auth user records to reflect changes made in auth / delius.

awhodgson256:

Amends user records in auth_user table to reflect changes made in Delius / Auth

Allow access to affected user's referrals which are currently erroring.


kth13 (approved):
LGTM

PR #1675, opened 2023-10-18
adding role INTERVENTIONS_SERVICE to all the end points

kth13:

- protecting all the end points with new role `INTERVENTIONS_SERVICE`

- There is move across MOJ to protect the resources with access based roles


awhodgson256 (approved):
LGTM

PR #1681, opened 2023-10-26
update usernames

rgforsyth:

Update usernames that have changed in delius

The users can no longer access the service due to a change in username


kth13 (approved):
LGTM

src/main/resources/db/migration/V1_148__update_auth_user_username_pt3.sql
```diff
@@ -0,0 +1,4 @@
+update auth_user set user_name = 'NICOLAKEARNSNPS' where id = 'NickyLewisNPS';
+update auth_user set user_name = 'RFL98H' where id = 'wdy26r';
+update auth_user set user_name = 'OXT30Z' where id = 'pda41z';
+update auth_user set user_name = 'ITS48E' where id = 'pbm99r';
```
marcocolluramoj, line 4: I think these are the usernames instead of the delius ids?

marcocolluramoj (approved):
LGTM

PR #1685, opened 2023-10-31
performance tuning for the ndmis report

kth13:

- increased the chunk size from 20 to 100
- Fixes the npe happening in complexity processor
- changes the schedule time to start every night 20:30 so that it can ran and finish up by start of the next days play
- removes the unused page size
- remove the filter sent_at not null in the main query

- the ndmis report started to run slowly and we are taking some corrective measures to improve its performance


amolphalnikar (approved):
Let's test these in preprod -

PR #1703, opened 2023-11-29
trivy ignore for reactor-netty-http

kth13:

- ignore the trivy scan for the latest reactor-netty-http CVE-2023-34062.  This can be suppressed as we are not using directory traversal in our application

- This is done to resolve the trivy scan failure


amolphalnikar (approved):
As discussed, let's get the perf changes and then tackle this.

PR #1716, opened 2023-12-07
Add overrides for trivy issues present to allow dev testing in advanc…

awhodgson256:
…e of Spring upgrade

Overrides trivy check for issues which are to be fixed when Spring upgrade is deployed

Allow the build to work and allow changes to be deployed to DEV environment as a temporary measure.


kth13 (approved):
LGTM

PR #1717, opened 2023-12-11
using native queries to retrieve data for dashboard.The native querie…

kth13:
…s is been projected as views

- configures the dashboard query to a refined and thinner version.
- configures the query to a view which makes it easier for pagination and sorting

- Dashboard query is generally slow and was pulling lots of unwanted data. So we came up with a query on top of a view which pulls in the data that is required


awhodgson256 (approved):
LGTM, pact tests to be monitored on merge

PR #1721, opened 2023-12-12
Use a simple jobLauncher instead of more general interface and

awhodgson256:
Use a simple jobLauncher instead of more general interface and introduce NullJobRepository to onStartupJobLauncherFactory

Passes the nullJobRepository to onStartupJobLauncher, used more generally by reports

Allow report to run against readonly datasource


kth13 (approved):
LGTM

PR #1723, opened 2023-12-12
Ipb 926/read replica env template revert job readonly

awhodgson256:

Reverts changes pushed for read replica that are failing

Restore mainline to working build


marcocolluramoj (approved):
LGTM

PR #1724, opened 2023-12-14
contract offboarding for ETE contracts

kth13:

- contract offboarding for ETE contracts. The contract referral end date is set at 31st December 2023 23:59:00

- ETE contract needs to be off boarded by the end of 2023


marcocolluramoj (approved):
LGTM

PR #1738, opened 2024-01-17
updating service based on changes in UI for updating pp details

kth13:

- saves delius telephone number and delius team telephone number
- changes the way we amend pp name

- The confirm probation practitioner details page design got changed and the service is changed to cater that


awhodgson256 (approved):
LGTM

PR #1746, opened 2024-01-30
add check for old appointents when rescheduling

rgforsyth:

add check for old appointents when rescheduling

add check for old appointents when rescheduling


marcocolluramoj (approved):
LGTM

PR #1748, opened 2024-01-31
NDMIS report configuration to point to read-replica

amolphalnikar:

NDMIS report configuration to point to read-replica

This is to improve performance and reduce load on primary database.


kth13 (approved):
I tested locally but even the PerformanceReport is also failing with the same error. So, it is best to test it directly in dev

PR #1754, opened 2024-02-06
Bump dps spring plugin version to fix vulnerability

awhodgson256:

Increased dps spring version to cope with a trivy vulnerability

Address security concern and move to latest available spring version in the dps plugin


kth13 (approved):
LGTM

PR #1755, opened 2024-02-06
Remove redundant jobRepository bean which is a duplicate

awhodgson256:

Removes a redundant jobRepository bean as it's already defined.

Fix deploy issue observed releasing into DEV


kth13 (approved):
LGTM

PR #1756, opened 2024-02-06
revising PP emails

kth13:

- revising PP email subject in the notification

- The pp email subject required modification including the crn's of the person on probation


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/notifications/NotifyService.kt
```diff
@@ -46,13 +46,18 @@ class NotifyActionPlanService(
       ActionPlanEventType.SUBMITTED -> {
         val recipient = referralService.getResponsibleProbationPractitioner(event.actionPlan.referral)
         val location = generateResourceUrl(interventionsUIBaseURL, ppActionPlanLocation, event.actionPlan.referral.id)
+        val popFirstName = event.actionPlan.referral.serviceUserData!!.firstName?.lowercase()?.replaceFirstChar { it.uppercase() }
```
amolphalnikar, line 49: With Irish / Scottish names like O'Conner / MacDonnell etc, this will be little odd. Can we not used the actual first name without lowercasing it ?

kth13: Hi @amolphalnikar I reused this logic which was used before and live now. I am ok to change it if you want to change that too.

amolphalnikar: I will anyway approve this.

amolphalnikar (approved):

PR #1758, opened 2024-02-08
Revert to code using deprecated hibernate cursors

awhodgson256:

Revert to code using deprecated hibernate cursors to prevent entityManagerFactory closure

Will require a technical debt ticket to investigate and resolve later

Fix bug due to autoclose of EntityManagerFactory observed in dev when multiple reports run


kth13 (approved):
LGTM

PR #1760, opened 2024-02-09
Point batch jobs to use primary JobRepository

awhodgson256:
instead of batchJobRepository due to issue with TM on H2

Point batch jobs to primary spring defined job repository

Resolve issue with batch jobs seen in Dev


kth13 (approved):
LGTM

PR #1762, opened 2024-02-09
Move batch jobs to run with a JpaTransactionManager

awhodgson256:

Moves tranaction manager from platform to JPA

Fix issue observed in dev with closed EntotyManagerFactory, preventing reports running


kth13 (approved):
LGTM

PR #1768, opened 2024-02-13
Fallback to single connection for reporting

awhodgson256:

Reverts earlier changes and falls back to using a single connection for transactional / reporting use

Fix issues observed in dev for Spring upgrade due to EntityManagerFactory being closed.


kth13 (approved):
LGTM

PR #1776, opened 2024-02-16
Revert to legacy job and status builders

awhodgson256:

Reverts job and step builders to the legacy factories

Provide accurate status indication for transfer / concude referrals jobs


kth13 (approved):
LGTM

PR #1783, opened 2024-02-21
Revert "Pointed the perf report to read-replica"

amolphalnikar:
This reverts commit 7b95fb6d812d9381816e88a7e5f651bb4dbaea62.

Reverted the previous commit which was pointing to read-replica.

The change did not work as expected hence reverting so other changes can go.


kth13 (approved):
LGTM

PR #1787, opened 2024-02-26
Change isolation level to a value H2 supports

awhodgson256:
(to prevent fallback to SERIALIZED)

Sets isolation level to h2 supported value

Allow use of H2 database for reporting batch config, allowing read only replica to serve the report


kth13 (approved):
LGTM

PR #1788, opened 2024-02-26
Set job launcher for NDMIS job to be a read only job launcher

awhodgson256:
Set job launcher for NDMIS job to be a read only job launcher referencing the read only job repository in the batch job's configuration

Change how NDMIS report is launched, to use the batch job launcher instead of read / write launcher

Fix problem with NDMIS report to allow it to run against read replica, thereby imoroving performance,

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/jobs/oneoff/OnStartupJobLauncherFactory.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/jobs/oneoff/OnStartupJobLauncherFactory.kt
@@ -6,7 +6,6 @@ import org.springframework.batch.core.Job
 import org.springframework.batch.core.converter.DefaultJobParametersConverter
 import org.springframework.batch.core.launch.JobLauncher
 import org.springframework.batch.core.launch.support.SimpleJvmExitCodeMapper
-import org.springframework.batch.core.launch.support.TaskExecutorJobLauncher
 import org.springframework.boot.ApplicationArguments
 import org.springframework.boot.ApplicationRunner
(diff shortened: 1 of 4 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/jobs/oneoff/OnStartupJobLauncherFactory.kt
```diff
@@ -44,6 +45,22 @@ class OnStartupJobLauncherFactory(
       return exitCodeMapper.intValue(execution.exitStatus.exitCode)
     }
 
+    return makeLauncher(job.name, entryPoint)
+  }
+  fun makeReadOnlyBatchLauncher(job: Job, readOnlyJobLauncher: TaskExecutorJobLauncher): ApplicationRunner {
```
kth13, line 50: makeBatchLauncher and makeReadOnlyBatchLauncher almost the same code with the only the launcher is difference. could you please create the readonlylauncher as another bean and as that as a parameter to a private method inside. That private method will be used both by `makeBatchLauncher` and makeReadOnlyBatchLauncher`

awhodgson256: No worries Kumar, will work that up this morning.

awhodgson256: Refactored as discussed @kth13 please review

kth13 (approved):
LGTM

PR #1791, opened 2024-02-28
Ipb 979/move ndmis report to read replica

awhodgson256:

Passes additional arguments to EnableBatchProcessing and removes unwanted beans

Allow NDMIS report to run against a read only source


kth13 (approved):
LGTM

PR #1792, opened 2024-02-29
Ipb 979/move ndmis report to read replica

awhodgson256:

Moves to a single global H2 config for job info

Fix deployment failures and prevent spring from randomly assigning job configuration based on multiple annotations


kth13 (approved):
LGTM

PR #1793, opened 2024-03-01
fixing pact test

kth13:

- Fix pact test after reason for referral

- Fix pact test after reason for referral


marcocolluramoj (approved):
LGTM

PR #1795, opened 2024-03-05
Parallelize the NDMIS report into separate flows

awhodgson256:

Runs ndmis query / file creation in parallel flows

Speed up the overall throughput of the report


amolphalnikar:
As per my understanding of the NDMIS reporting, I believe the first step / flow of reading referral would be on its own as ndmisReader, is input to all other steps/flows. That way it won't keep reading every time from database as it will bring all associated entities. It will be followed by the parallel flows of remaining 3; appointments / outcome / complexities and then the final step to push to S3 which seems to be right.
The approach which has been followed now, I believe will result into running 4 set of queries against referral table at the same time. Or did I get it wrong altogether ?

PR #1797, opened 2024-03-05
update id for new sessions concerns email

rgforsyth:

Used the new session concerns email.

The new id was not being used, so this change replaces the p2b id with the p2c id. The new crn field was already updated previously


kth13 (approved):
LGTM

PR #1798, opened 2024-03-05
Temporarily check notifyProbationPractitioner  in feedback requests

marcocolluramoj:

Temporarily check notifyProbationPractitioner  in feedback requests

To prevent issues with in-progress feedback in between deployments.


kth13 (approved):
LGTM

PR #1799, opened 2024-03-06
Add logging statements to the processors to allow info reveal in dev

awhodgson256:

Add exception handling for investigation in DEV

Reveal why report is failing in DEV but not locally.


kth13 (approved):
LGTM

PR #1802, opened 2024-03-08
Case note and notify pp hotfix

marcocolluramoj:

Reverts the revert of case note email opt out commit and adds the notify pp of poor behaviour changes backwards compatibility commit

Reverts the revert of case note email opt out commit and adds the notify pp of poor behaviour changes backwards compatibility commit


kth13 (approved):
LGTM

PR #1805, opened 2024-03-11
Make main db transaction manager a prototyoe bean

awhodgson256:
Make main db transaction manager a prototyoe bean & pass to individual flows instead of at global job level

(cherry picked from commit ea50da77ff48b5e11ad5596c5ed47fd1f082d894)

Passes a prototype bean for TX manager down to individual flows

Make parallel processing of NDMIS report work and speed up the process


kth13 (approved):
LGTM

PR #1807, opened 2024-03-12
NDMIS reader singleton > prototype

awhodgson256:

Makes job reader a prototype bean to prevent sharing in the NDMIS report config

Overcome issues observed when parallel processor is deployed to DEV


kth13 (approved):
LGTM

PR #1808, opened 2024-03-12
Define dedicated txManagers

awhodgson256:

Defines dedicated txManagers for each of the flows in NDMIS report

Investigate failing report on DEV


kth13 (approved):
LGTM

PR #1809, opened 2024-03-12
add date first saa was completed & update sessions count to only include completed appointments

rgforsyth:

Adds a new field added to show the data the `date_saa_first_completed`, as well as the old `date_saa_first_attended` field as these can be different appointments now.

Update to the  `count_of_sessions_attended` field to only include appointments that are completed.

The current data is incorrect as of it2v4 going live.


kth13 (approved):
LGTM

PR #1815, opened 2024-03-14
Move to using a JPAItemCursor / builder

awhodgson256:

Move to using a JPAItemCursor / builder - Instead of deprecated hibernate classes.

Investigate issue with EntityManagerFactory being closed in DEV


kth13 (approved):
LGTM

PR #1816, opened 2024-03-14
update ndmis report to have relevant data after it2v4 changes

rgforsyth:

adds a new field added to the appointment csv: `did-session-happen`

Update to the count_of_sessions_attended on referral csv field to only include appointments that are completed.

This can be either:

`did_session_happen` = null and `attended` = YES/LATE

`did_session_happen` = true and `attended` = YES

After it2v4 went live, some data in the service provider report is no longer correct.


kth13 (approved):
LGTM

PR #1822, opened 2024-03-18
fixing the pact test

kth13:

- fixing the pact test after introducing new field `alreadyKnowPrisonName`

- fixing the pact test after introducing new field `alreadyKnowPrisonName`


awhodgson256 (approved):
LGTM

PR #1826, opened 2024-03-19
add missing ifEmpty check to appointment processor

rgforsyth:

Adds back in the empty check to the appointment processor

Currently the ndmis is failing in dev.


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/reporting/ndmis/performance/AppointmentProcessor.kt
```diff
@@ -24,13 +24,13 @@ class AppointmentProcessor(
         appointmentTime = NdmisDateTime(it.appointmentTime),
         durationInMinutes = it.durationInMinutes,
         bookedAt = NdmisDateTime(it.createdAt),
-        didSessionHappen = it.didSessionHappen,
         attended = it.attended,
+        didSessionHappen = it.didSessionHappen,
```
rgforsyth, line 28: Just moved where this is in the list to be consistent with the AppointmentData file.

PR #1827, opened 2024-03-19
changing the contract to only cornwall

kth13:

- changing the D&R south west contract only to cornwall

- changing the D&R south west contract only to cornwall


awhodgson256 (approved):
LGTM

PR #1830, opened 2024-03-20
Make ndmis 4 jobs and launch from a staggered start 

awhodgson256:
- initial implementation

Remove code to call service / controller for ndmis as it's not used

Splits NDMIS report into 4 jobs / processes

Fix issues observed in DEV when testing NDMIS report performance improvements


amolphalnikar:
Look good to me Andrew although I would prefer to keep the existing job (with 5 sequencial steps) as it is and just not to invoke it. Would it be possible to keep the code that way so we have an opportunity to work on the parallel steps when we get a chance.

amolphalnikar (commented):
Look good to me Andrew although I would prefer to keep the existing job (with 5 sequencial steps) as it is and just not to invoke it. Would it be possible to keep the code that way so we have an opportunity to work on the parallel steps when we get a chance.

amolphalnikar (approved):

PR #1838, opened 2024-03-26
remove did-session-happen field from appointmen csv

rgforsyth:

Removes did-session-happen from the appointment.csv which is no longer needed.

Removes did-session-happen from the appointment.csv which is no longer needed.


kth13 (approved):
LGTM

PR #1843, opened 2024-03-27
remove validation checks that aren't necessary for this flow

rgforsyth:

Removes validation checks that aren't needed for the flow where did-session-happen is false.

Current you are prevented from submitting feedback for a historic appointment if you select did-session-happen as false, as the validation is expecting fields to be there that aren't.


awhodgson256 (approved):
LGTM

PR #1847, opened 2024-03-28
Update FBD in North East contract

marcocolluramoj:

Updates the title, description, service category and adds a medium complexity level for Finance, Benefit & Debt in the North East contract as it is now delivered in both custody and community.

Updates the title, description, service category and adds a medium complexity level for Finance, Benefit & Debt in the North East contract as it is now delivered in both custody and community.


kth13 (approved):
LGTM

PR #1850, opened 2024-04-02
Update FBD NE contract type

marcocolluramoj:

Update FBD-NE contract type name from 'Finance, Benefit and Debt in Custody (North East)' to 'Finance, Benefit and Debt in the North East'

Update FBD-NE contract type name from 'Finance, Benefit and Debt in Custody (North East)' to 'Finance, Benefit and Debt in the North East'


kth13 (approved):
LGTM

PR #1859, opened 2024-04-18
point dashboard queries to read replica

awhodgson256:

point dashboard queries to read replica and allow instance size for primary database to be reduced in the future.

Move load for read-only dashboard queries to read replica, to improve application performance


amolphalnikar (approved):

kth13 (approved):
LGTM

PR #1861, opened 2024-04-19
Remove temporary trivy ignores following dps plugin / spring upgrade

awhodgson256:

Removes trivy ignores

Allow further security checking having upgraded the affected library


kth13 (approved):
LGTM

PR #1870, opened 2024-04-26
Cherry picked changes with SRF and view for profiling

awhodgson256:

Adds a view and set returning function to profile changes in preprod

Investigate change options to improve service provider report response times


kth13 (approved):
LGTM

PR #1873, opened 2024-05-07
Ipb 1121/withdraw referral

kth13:

- new api for withdraw controller
- notify users based on withdrawal state

- the app is moving from cancel referral to withdrawal referral


rgforsyth (commented):
Looks good. 
I noticed the AC mentions: 
"Record in App Insights as withdrawn action
Update a count of ‘withdrawn post ICA’ in DB for MI "

Not sure where this needs to happen exactly but I don't think I seen it.

rgforsyth (approved):

PR #1876, opened 2024-05-09
Amend contract code to avoid primary key duplication, minor presentation changes

awhodgson256:

Amend contract code to avoid primary key duplication, minor presentation changes

Avoid primary key violations; match expected presentation


kth13 (approved):
LGTM

PR #1891, opened 2024-05-22
Fix issue with duplicate contract / intervention id's for Bolton PWB

awhodgson256:

Amends UUIDs for Bolton PWB

Allow both FBD and PWB contracts to be visible for Bolton Big Life Group


kth13 (approved):
LGTM

PR #1909, opened 2024-06-06
No ticket/pi 2220 add id reference and completed date to probation case

pmcphee77:

This adds the referral id, referral reference and completed date to the ProbationCaseReferralDTO so that this is returned in the /probation-case/{crn}/referral endpoint

To enable Manage a Supervision to render a list of referrals and provide links into Refer and Monitor


amolphalnikar:
Hi Paul, this looks good and got confirmation that adding these fields won't have impact on resettlement passport. 

I believe you would need to add role ROLE_INTERVENTIONS_API_READ_ALL (which integration layer is using when calling R&M APIs) for method getReferralByCrn under ProbationCaseController. The existing role is for resettlement passport.

pmcphee77:
> ROLE_INTERVENTIONS_API_READ_ALL

Hi Amol - yes we have added that role to our client in dev so we can access the endpoint. Thanks

amolphalnikar (approved):

PR #1912, opened 2024-06-07
Update supplier code which was previously too long for auth mapping

awhodgson256:

Update supplier code which was previously too long for auth mapping for rochdale connections trust

Allow auth mappings to work correctly for this supplier and users to gain access


kth13 (approved):
LGTM

PR #1918, opened 2024-06-17
Fix bugs in SP report output

awhodgson256:

Fix bugs with nulls being substituted with current time imestamps

Fix bug with attendance time for first session post SAA (was SAA value)

Ensure consistency in output for revised SP performance report

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/jpa/repository/ReferralPerformanceReportRepositoryImpl.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/jpa/repository/ReferralPerformanceReportRepositoryImpl.kt
@@ -127,8 +127,7 @@ class ReferralPerformanceReportRepositoryImpl : ReferralPerformanceReportReposit
   }
 
   fun timestampToOffset(timestamp: Timestamp?): OffsetDateTime? {
-    val resolved = timestamp
-    return OffsetDateTime.ofInstant(resolved?.toInstant(), ZoneId.of("UTC"))
+    return OffsetDateTime.ofInstant(timestamp?.toInstant(), ZoneId.of("UTC"))
   }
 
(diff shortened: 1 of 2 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/jpa/repository/ReferralPerformanceReportRepositoryImpl.kt
```diff
@@ -127,9 +127,8 @@ class ReferralPerformanceReportRepositoryImpl : ReferralPerformanceReportReposit
   }
 
   fun timestampToOffset(timestamp: Timestamp?): OffsetDateTime? {
-    val currentTimeMillis = System.currentTimeMillis()
-    val resolved = timestamp ?: Timestamp(currentTimeMillis)
-    return OffsetDateTime.ofInstant(resolved.toInstant(), ZoneId.of("UTC"))
+    val resolved = timestamp
```
kth13, line 130: I believe you don't need resolved variable. You can directly do `OffsetDateTime.ofInstant(timestamp?.toInstant(), ZoneId.of("UTC"))`.

src/main/resources/db/migration/V1_176__update_referral_performance_report_first_attendance_srf.sql
```diff
@@ -0,0 +1,19 @@
+create or replace function performance_report_first_attendance(referral_id UUID)
```
kth13, line 1: Just out of curiosity, how this function gets used in the main function?

awhodgson256: It's not called by the main report function @kth13 that's just used by the reader - it gets called at the later stage to find the appointment time for the processor (and added to the output)

kth13 (approved):
LGTM

PR #1920, opened 2024-06-17
Removed draft referrals from API EP response

amolphalnikar:

Removes the draft referrals from the API EP response

There are many hanging draft referrals which is not needed in the API EP response. There is no contact created in nDelius for draft referrals.


kth13 (approved):
LGTM

PR #1922, opened 2024-06-18
Fix final column affected by not null protection issue

awhodgson256:

Allows date of first attended session to be nullable instead of defaulting to current data

Fix issue with nulls not being shown in the SP performance report


amolphalnikar:
Hi @awhodgson256 , This changes looks good to me. 
I believe we don't need to use timestampToOffsetNotNull method. It is used date_referral_received and we ideally don't expect that to be null. And even if that is null, I feel we should not set that to current date time.

awhodgson256:
> Hi @awhodgson256 , This changes looks good to me. I believe we don't need to use timestampToOffsetNotNull method. It is used date_referral_received and we ideally don't expect that to be null. And even if that is null, I feel we should not set that to current date time.

@amolphalnikar the not null version is required not for runtime issue, but to satisfy the compiler.  The column is not nullable so in the event of receiving a null it needs to map to a value.  This will never happen at runtime as the value always has a legitimate value for sent referrals (it's the date sent).   Hope that makes sense?

amolphalnikar (approved):

PR #1923, opened 2024-06-18
update description for some withdrawal reasons

kth13:

- update withdrawal reason description for couple of withdrawal reasons

- update withdrawal reason description for couple of withdrawal reasons


awhodgson256 (approved):
LGTM

PR #1932, opened 2024-06-25
Apply fix for not null failures when number of outcomes is null 

awhodgson256:

Apply fix for not nulll failures when number of outcomes is null in ihe SP Perf Report

Prevent errors some SP's have reported with report generation


kth13 (approved):
LGTM

PR #1938, opened 2024-06-28
sp perf report bugfixes - type mapping fix

awhodgson256:

Fixes job failure cause due to casting

Allow the report to run with final column


kth13 (approved):
LGTM

PR #1944, opened 2024-07-02
Rework report queries to provide fixes

awhodgson256:
SAA appointment appearing in the attended session count Latest expected session count is used where action plan is revised

Fixes issues reported by St Giles in relation to report session attendance counts

Provide consistent report with required output as was before performance rework was applied.


kth13 (approved):
LGTM

PR #1947, opened 2024-07-03
Tweak SQL to remove duplicates due to superseded details / appointments

awhodgson256:

Tweak SQL for providers report to remove duplicates due to superseded details referral details records
also don't count superseded appointments for attendance counts.

Address duplication issues and count discrepencies reported in csv output

### DIFF
--- a/src/main/resources/db/migration/V1_186__update_referral_performance_report_number_of_attendances_srf.sql
+++ b/src/main/resources/db/migration/V1_186__update_referral_performance_report_number_of_attendances_srf.sql
@@ -1,0 +1,19 @@
+create or replace function performance_report_attendances_count(referral_id UUID)
+  returns integer
+as
+$body$
+SELECT count(*)
+FROM appointment a
+WHERE a.referral_id = $1
+	AND a.stale = false
+	AND (a.superseded is null OR a.superseded = false)
+	AND ((a.did_session_happen is null AND (a.late = true OR a.attended = 'YES'))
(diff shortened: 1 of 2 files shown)


src/main/resources/db/migration/V1_186__update_referral_performance_report_number_of_attendances_srf.sql
```diff
@@ -0,0 +1,19 @@
...
+SELECT count(*)
+FROM appointment a
+WHERE a.referral_id = $1
+	AND a.stale = false
+	AND (a.superseded is null OR a.superseded = false)
+	AND ((a.did_session_happen is null AND (a.late = true OR a.attended = 'YES'))
+	OR (a.did_session_happen = true AND a.attended = 'YES'))
+	AND NOT EXISTS
+    (
+        SELECT 1
```
kth13, line 14: are we selecting only delivery session appointment here. If that is the case we could have did a join in delivery session appointments table instead of checking not exists

awhodgson256: From what I can tell not exists is more efficient than a join (for a negative case), but we can benchmark time taken once we're in preprod. @kth13

kth13: Ok sure Andrew.

kth13 (approved):
LGTM

PR #1953, opened 2024-07-09
Increase dashboard replica count to 3

awhodgson256:

Increase dashboard replica count to 3

Spread load to additional instance with a view of reducing overall load level on DB


rgforsyth (approved):
LGTM

PR #1954, opened 2024-07-09
Filter out SAA from first attendance column

awhodgson256:

Filter out SAA from first attendance column

Avoid saa appointment being returned as a first attendance


kth13 (approved):
LGTM

PR #1959, opened 2024-07-11
use correct field for withdrawal comments

rgforsyth:

use correct field for withdrawal comments. Previously it was setting the withdrawal reason code description as the comment.

Fix failing pact tests.


kth13 (approved):
LGTM

PR #1964, opened 2024-07-15
notify amending expected release date unknown reason

kth13:

- fixing the issue when email is not send for expected release not known reason

- fixing the issue when email is not send for expected release not known reason


awhodgson256 (approved):
LGTM

PR #1967, opened 2024-07-16
upgrade owasp dependency check jdk version to 21

kth13:

- upgrade owasp dependency check jdk version to 21

- The new version of dps gradle spring boot will only support jdk 21 and above


awhodgson256 (approved):
LGTM

PR #1968, opened 2024-07-17
Fix logic in first attendance and session attendance counts

awhodgson256:

Fix logic in first attendance and attendance counts to use a.attended = 'LATE' instead of a.late boolean flag.

Fix incorrect reporting of attendance counts and date time of session


kth13 (approved):
LGTM

PR #1971, opened 2024-07-23
Fix null checking on saa_completed in SP report

awhodgson256:

Fix null checking on saa_completed in SP report

Replace (null =) with (is null), to cater for legacy records prior to having a did_session_happen value (approx 01/01/2024)


kth13 (approved):
LGTM

PR #1975, opened 2024-07-25
Add preventEmailNotification flags to the amend probation office flows

marcocolluramoj:

Adds preventEmailNotification flags to the amend probation office flows.

To be able to prevent duplicate emails being sent for pre-release referrals with no COM when both expected and PP probation offices are updated at the same time.


kth13 (approved):
LGTM

PR #1978, opened 2024-07-29
Ipb 1163/add end sentence with batch load

awhodgson256:

Adds a batch job to load data to support adding end of sentence date, call out to record the date on sending a referral.

Allow end of sentence date to be recorded, with data backfilling.


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/jpa/entity/RelevantEndOfSentenceDataload.kt
```diff
@@ -0,0 +1,15 @@
...
+
+import jakarta.persistence.Column
+import jakarta.persistence.Entity
+import jakarta.persistence.Id
+import jakarta.persistence.Table
+import java.time.LocalDate
+
+@Entity
+@Table(name = "relevant_sentence_end_dataload")
+class RelevantEndOfSentenceDataload(
```
kth13, line 11: Hi @awhodgson256 , did you contemplated on creating a csv for this rather than a table. The reason why I am thinking this is a one off job and reading from csv would have been easier

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/RamDeliusAPIConvictionService.kt
```diff
@@ -0,0 +1,73 @@
...
+  val mainOffence: Offence,
+  val active: Boolean,
+)
+
+data class ConvictionDetails(
+  val conviction: Conviction,
+)
+
+@Service
+class RamDeliusAPIConvictionService(
```
kth13, line 34: We already have a `uk.gov.justice.digital.hmpps.hmppsinterventionsservice.component.RamDeliusClient` class which I believe we can use that.

src/main/resources/db/migration/V1_195__add_sentence_end_date_loading_table.sql
```diff
@@ -0,0 +1,9 @@
+CREATE TABLE relevant_sentence_end_dataload (
```
kth13, line 1: Same question as above. can this be done via csv?

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/RamDeliusAPIConvictionService.kt
awhodgson256: Pushed changes to use RamDeliusClient as requested @kth13

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/jpa/entity/RelevantEndOfSentenceDataload.kt
kth13: Ok @awhodgson256 , no worries.

kth13 (approved):
LGTM

PR #1981, opened 2024-08-01
Remove call to retrieve token 

awhodgson256:

Remove call to retrieve token - addressing issue with nDelius integration

Fix integration with nDelius on referral send

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/component/RamDeliusClient.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/component/RamDeliusClient.kt
@@ -14,5 +14,5 @@ class RamDeliusClient(private val ramDeliusApiClient: RestClient) {
     ramDeliusApiClient.put(uri, requestBody).retrieve().bodyToMono(AppointmentResponseDTO::class.java).block()
 
   fun makeGetConvictionRequest(uri: String): ConvictionDetails =
-    ramDeliusApiClient.get(uri, null, null).retrieve().bodyToMono(ConvictionDetails::class.java).block()
+    ramDeliusApiClient.get(uri).retrieve().bodyToMono(ConvictionDetails::class.java).block()
 }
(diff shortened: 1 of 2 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/RamDeliusAPIConvictionService.kt
```diff
@@ -41,13 +39,7 @@ class RamDeliusAPIConvictionService(
...
       .toString()
 
-    val auth = SecurityContextHolder.getContext()?.getAuthentication()
-    var token: JwtAuthenticationToken? = null
-    if (auth !== null) {
-      token = auth as JwtAuthenticationToken
-    }
-
-    val convictionDetails = ramDeliusApiClient.makeGetConvictionRequest(convictionPath, token)
+    val convictionDetails = ramDeliusApiClient.makeGetConvictionRequest(convictionPath, null)
```
kth13, line 42: @awhodgson256 could you please default the null value in `ramDeliusApiClient.makeGetConvictionRequest(convictionPath, null)`

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/component/RamDeliusClient.kt
```diff
@@ -14,6 +13,6 @@ class RamDeliusClient(private val ramDeliusApiClient: RestClient) {
   fun makePutAppointmentRequest(uri: String, requestBody: Any): AppointmentResponseDTO? =
     ramDeliusApiClient.put(uri, requestBody).retrieve().bodyToMono(AppointmentResponseDTO::class.java).block()
 
-  fun makeGetConvictionRequest(uri: String, token: JwtAuthenticationToken?): ConvictionDetails =
-    ramDeliusApiClient.get(uri, null, token).retrieve().bodyToMono(ConvictionDetails::class.java).block()
+  fun makeGetConvictionRequest(uri: String): ConvictionDetails =
+    ramDeliusApiClient.get(uri, null, null).retrieve().bodyToMono(ConvictionDetails::class.java).block()
```
kth13, line 17: @awhodgson256 The last two parameters are already defaulted to null. So we don't need to pass that as well `ramDeliusApiClient.get(uri, null, null)`

kth13 (changes requested):

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/service/RamDeliusAPIConvictionService.kt
awhodgson256: Rectified that @kth13

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/component/RamDeliusClient.kt
awhodgson256: @kth13 removed to simplify as suggested

kth13 (approved):
Approved

PR #1985, opened 2024-08-05
Ignore errors on end of sentence lookup

awhodgson256:

Ignores result of call to retrieve end of sentence date from nDelius

Allow referral to still be sent in the event we don't get a value or an error occurs with the end of sentence lookup


kth13 (approved):
LGTm

PR #1987, opened 2024-08-06
Increase chunk size to resolve long job timings

awhodgson256:

Increase chunk size to resolve long job timings

Reduce job run time


kth13 (approved):
LGTM

PR #1993, opened 2024-08-08
Hotfix/fix sp viewing issue

kth13:

- fixes the issue where sp's can see referrals out of their region

- fixes the issue where sp's can see referrals out of their region


awhodgson256 (approved):
LGTM

PR #1994, opened 2024-08-09
Reduce batch sizing for load end of sentence job

awhodgson256:

Reduced load job transaction size to 5000 records

Prevent out of memory errors when running the end of sentence loading job


kth13 (approved):
LGTM

PR #1995, opened 2024-08-09
Add referral status

marcocolluramoj:

Adds a status column to the referral table in the database and updates this to POST_ICA when feedback is given for an attended delivery session.

To correctly filter referrals in the dashboard tabs


src/main/resources/db/migration/V1_199__add_referral_status.sql
```diff
@@ -0,0 +1,6 @@
+CREATE TYPE status AS ENUM ('PRE_ICA','POST_ICA');
```
kth13, line 1: You may have to create a index for this column

kth13 (approved):
LGTM

PR #1996, opened 2024-08-09
Delete old draft referral batch job

kth13:

- deleting the draft referral older than 90 days

- deleting the draft referral older than 90 days


awhodgson256 (approved):
LGTM

PR #1997, opened 2024-08-09
Modify sp report queries to ignore appointments without feedback

awhodgson256:

Filters SAA and delivery session first appointment records where the record has no feedback from the report

Prevent rescheduled appointments without feedback showing in the report


kth13 (approved):
LGTM

kth13 (approved):
LGTM

PR #2016, opened 2024-09-13
Fix for SAR data controller so it accepts UK dates (not ISO)

awhodgson256:

Fix for SAR data controller so it accepts UK dates (not ISO) and empty string date submissions

Fix integration issues with SAR data gathering process


amolphalnikar (approved):
LGTM

PR #2020, opened 2024-09-17
Move concluded / withdrawal event to post db update

awhodgson256:

Move notification event(s) to after the updates are applied in R&M

Avoid out of sync notifications when database change in unsuccessful causing issues with NDelius


amolphalnikar:
Looks good @awhodgson256 . Just wondering if we need to add exception handling block for conclude method call ? If something goes wrong with that, we don't want to publish conclude event.

awhodgson256:
> Looks good @awhodgson256 . Just wondering if we need to add exception handling block for conclude method call ? If something goes wrong with that, we don't want to publish conclude event.

Hi @amolphalnikar In the event of exception (or a throwable), processing stops and we jump back to exception handling further up the call stack - so in that event we don't get as far as the "signalEnding" method which would send the event.  I think we're ok?

amolphalnikar:
If that's the case then great. Thanks @awhodgson256.  I have approved the PR.

amolphalnikar (approved):

PR #2023, opened 2024-09-23
Reset referral status

marcocolluramoj:

Amend LoadStatusReader to only look for post-ICA referrals

To re-run the batch job and reset the status for referrals with no attended delivery sessions back to pre-ICA


rgforsyth (approved):
LGTM

PR #2026, opened 2024-10-01
Ipb 1062/add notify email for pp name

awhodgson256:

Passes required user parameter into the pp name email template

Fix bug with initial implementation preventing email send


rgforsyth (approved):
LGTM

PR #2031, opened 2024-10-07
Ipb 1243/amend reason for referral

rgforsyth:

Allows reason_for_referral & reason_for_referral_further_information to be amended.
Adds change log entries for amendments to the above.
Sends an email if any of the above were changed.

Allows reason_for_referral & reason_for_referral_further_information to be amended.


awhodgson256 (approved):
LGTM

PR #2059, opened 2025-02-06
Ipb 1306/spring major version update

awhodgson256:

- Upgrades spring dps plugin version to 7.0.0
- Change java version
- Fix deprecated cursor code within reporting and other regular jobs due to lack of support
- Peg to supported version of hibernate to avoid errors in tests due to ordering of deletes

- Move to latest version with a view of incorporating current and future vulnerabilty fixes
- Tidy up naming on routine jobs and remove redundant ones


thomaswilsonxyz:
I've done a review of the first 25% and left a couple of comments - most of this appears to be "just" formatting of various bits, as well as some deletion. 

One thing that stood out is the use of both `scheduled` and `routine` - is there a meaningful difference between those two things or are they synonyms ?

build.gradle.kts
```diff
@@ -2,11 +2,13 @@ import org.jetbrains.kotlin.gradle.tasks.KotlinCompile
 plugins {
   kotlin("plugin.spring") version "2.0.10"
   id("org.jetbrains.kotlin.plugin.jpa") version "2.0.10"
-  id("uk.gov.justice.hmpps.gradle-spring-boot") version "6.0.9"
+  id("uk.gov.justice.hmpps.gradle-spring-boot") version "7.0.0"
   id("jacoco")
   id("project-report")
 }
 
+ext["hibernate.version"] = "6.5.3.Final"
```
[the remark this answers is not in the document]
awhodgson256: It was added to limit the version upgraded to in order to maintain compatibility - it's a temporary fix which can be removed if additional work is done to fix deletion order both within main source and the integration tests.  Without it there were a bunch of test failures due to persisting entities in the wrong order.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/config/BatchConfiguration.kt
```diff
@@ -46,25 +44,11 @@ class BatchConfiguration(
     return launcher
   }
 
-  @Bean("batchJobBuilderFactory")
```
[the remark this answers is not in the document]
awhodgson256: JobBuilderFactories have been deprecated for some time and now don't work.  I've reworked to use JobBuilder instead.   Same applies to StepBuilder.

awhodgson256:
> I've done a review of the first 25% and left a couple of comments - most of this appears to be "just" formatting of various bits, as well as some deletion.
> 
> One thing that stood out is the use of both `scheduled` and `routine` - is there a meaningful difference between those two things or are they synonyms ?

I've corrected the docs - we appear to have only scheduled jobs (they are all realistically cronjobs) - routine showed up as i did a multistep rename / copy - should all be in the scheduled folder in the final state.  I've left the guidance on oneoff jobs in case someone wants to create one later.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/reporting/ndmis/performance/NdmisAppointmentPerformanceReportJobConfiguration.kt
```diff
@@ -89,7 +87,7 @@ class NdmisAppointmentPerformanceReportJobConfiguration(
     val validator = DefaultJobParametersValidator()
     validator.setRequiredKeys(arrayOf("timestamp", "outputPath"))
 
-    return jobBuilderFactory["ndmisAppointmentPerformanceReportJob"]
+    return JobBuilder("ndmisAppointmentJob", jobRepository)
```
thomaswilsonxyz, line 90: can I just check that the re-naming of this job won't cause any weird side-effects ?

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/reporting/ndmis/performance/NdmisComplexityPerformanceReportJobConfiguration.kt
```diff
@@ -89,7 +87,7 @@ class NdmisComplexityPerformanceReportJobConfiguration(
     val validator = DefaultJobParametersValidator()
     validator.setRequiredKeys(arrayOf("timestamp", "outputPath"))
 
-    return jobBuilderFactory["ndmisComplexityPerformanceReportJob"]
+    return JobBuilder("ndmisComplexityJob", jobRepository)
```
thomaswilsonxyz, line 90: same here re: job name change

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/reporting/ndmis/performance/NdmisReferralPerformanceReportJobConfiguration.kt
thomaswilsonxyz, line 1: to save leaving a bunch of similar comments: there are several jobs that have changed names, and which now seem to mismatch the names specified in the annotations

I just want to double-check that's okay, won't cause any annoyance in debugging/finding the right code for future devs, or if that's an unreasonable concern.

thomaswilsonxyz (approved):
okay, I've reviewed this to the best of my ability, given how new I am and the size of the PR.

I've left a handful of comments, mostly double-checking things being moved/renamed/removed.

I'm putting a lot of faith in the automated tests to catch and particular issues here, but everything seems sensible.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/reporting/ndmis/performance/NdmisReferralPerformanceReportJobConfiguration.kt
awhodgson256: As above comment - reverted to old names

PR #2067, opened 2025-03-24
Contact onboarding setup for remaining SE contracts

awhodgson256:
Change hide-before dates

Contact onboarding setup for remaining SE contracts
Change hide-before dates

Allow referrals to contracts within SE England to be made

### DIFF
--- a/src/main/resources/providers/providers.json
+++ b/src/main/resources/providers/providers.json
@@ -460,12 +460,24 @@
     "name": "Touchstone - Leeds"
   },
   {
+    "code": "TURNING_POINT",
+    "name": "Turning Point"
+  },
+  {
     "code": "UP_BEAT_LIFE",
     "name": "Up Beat Life"
   },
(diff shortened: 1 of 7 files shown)


src/main/resources/providers/providers.json
```diff
@@ -467,6 +471,14 @@
     "code": "URBAN_OUTREACH",
     "name": "Urban Outreach Bolton"
   },
+  {
+    "code": "V_I_A",
+    "name": "v-i-a"
```
thomaswilsonxyz, line 476: why the case change here ? the other `name` fields seem to have capitals

thomaswilsonxyz (approved):

src/main/resources/providers/providers.json
awhodgson256: It was lower case as defined in the ticket... IPB-1319

PR #2068, opened 2025-03-25
Upgrade to latest version of spring dps for cve-2025-22228

awhodgson256:

Upgrades spring dps version to move spring security update (crypto) to 6.4.4

Resolve a security related issue which was causing owasp scan/pipeline to fail


Jack-Dempster (approved):
👍

PR #2072, opened 2025-04-07
IPB-1327: Add new desired outcomes for all non-accommodation service categories

Jack-Dempster:

Add local migrations for the non-accommodation desired outcomes changes

Add the remaining sql inserts for the updates to the desired outcomes.


src/main/resources/db/local/V99_4_0__deprecate_old_desired_outcomes.sql
```diff
@@ -36,13 +39,16 @@ where id IN (
 -- -- New outcomes
 -- UPDATE desired_outcome
```
Davansah, line 40: Barring the last sql statement, are the others meant to be commented out?

Jack-Dempster: yeah they are commented out as they are just the revert back to original scripts if we need them. Have just left them there when we are wanting to do some testing in dev

Davansah (approved):

PR #2079, opened 2025-04-16
IPB-1345: Align our local migrations/data to higher env data

Jack-Dempster:

This pr aligns the data we have for service categories and their desired outcomes in local to the higher environment data.

This data has been loaded into the higher environments via a job. However as we want to update the desired outcomes via a migration we need to have this data locally before the job runs or else locally the service will never start.


niaevans9 (approved):

niaevans9 (approved):
lgtm

PR #2081, opened 2025-04-22
updating the description for FBD in North east

kth13:

- updates the description for FBD in north east

- The description of the FBD is changed by the providers. So, it requires an update


thomaswilsonxyz (approved):
lgtm

PR #2111, opened 2025-06-13
fixing vulnerability issue. upgrading org.postgresql:postgresql library

kth13:

- trivy and owasp dependency fails with current version postgres library.  This PR fixes by upgrading to the version of postgres library which fixes it.

- trivy and owasp dependency fails with current version postgres library. The latest version 42.7.7 is fixing it.


Jack-Dempster (approved):
👍

PR #2120, opened 2025-07-09
IPB-1407: Added contract for North West Accommodation Services

Davansah:

Creates the required json file and updates Provider information for the Upsert Contracts job to consume and process contract details ready to be used in the service.

Onboards contract for North West Accommodation Services in the Community


kth13 (approved):
LGTM

PR #2220, opened 2026-04-15
fix the issue where expired contracts is shown

kth13:

- fixes the issue where the api throws error when the intervention is expired

- The user were able to access the expired contracts from the interventions page


moj-keith-au (approved):
LGTM

PR #2221, opened 2026-04-15
added a query param to control the referral end date check

kth13:

- added a query param to control the referral end date check

- This is to allow only find interventions page to check the referral end date


moj-keith-au (approved):
LGTM

PR #2225, opened 2026-06-02
Adding SAR mustache template endpoint

kth13:

- adds mustache template for sar data
- adds a new end point `/subject-access-request/template` to provide the template to the caller service
- since intervention service is still in older version, we need to create the api ourselves

- This is to provide mustache endpoint for the sar data


Davansah (approved):
LGTM

PR #2226, opened 2026-06-02
Fixing slow queries in SAS referral endpoint

kth13:

- Instead of using lazy fetch referral, use join eager fetch to avoid n+1 queries
- Doing this for both referral and draft referral
- adding a new index for service_usercrn in the draft referral

- This is to fix slow queries for the sas referral endpoint


Davansah (approved):
LGTM

PR #2227, opened 2026-06-02
Fixing slow queries in SAS referral endpoint

kth13:

- refactoring to move away from JPA queries to native queries

- This is to fix the performance issues in the application


Davansah (approved):
LGTM

PR #2230, opened 2026-06-25
not including draft referrals as part of SAS api response

kth13:

- not including draft referrals as part of SAS api response

- This is done to resolve timeout issues and exposing Draft referrals is not of much value to the client


Giruvagen (approved):

moj-keith-au (approved):
LGTM

PR #2235, opened 2026-07-24
Add draft-end-of-service-report url in ingress (modsec) exclude list

moj-keith-au:

Update the Ingress / ModSecurity configuration to exclude the following URL from ModSecurity rules

/draft-end-of-service-report/{reportId}

When user submitting a update for a EOSR with additional task comment to the application, the request is being blocked by ModSecurity at the Ingress layer and returns a 423 Locked status. The request never reaches the application. This is currently preventing users from completing the form submission.
After investigation, the block is caused by ModSecurity input validation rules incorrectly flagging the request payload for this particular endpoint.


kth13 (approved):
LGTM

PR #2236, opened 2026-07-30
revert the junit platform launcher upgrade as it is not working

kth13:

- revert the junit platform launcher upgrade as it is not working

-revert the junit platform launcher upgrade as it is not working


moj-keith-au (approved):
LGTM

